PR #11, opened 2016-02-12
There is no need to create another client factory since TestHazelcast…

eminn:
…Factory can create both clients and members. It was causing tests to fail when ran from maven.


cangencer:
:+1:

PR #303, opened 2017-02-06
Do not fail job when new members join

mdogan:
Do not fail job when new members join after execution is planned and even when those new members are removed.

PS: *According to our latest meeting on Friday, this change should target 0.3.1.*

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetService.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetService.java
@@ -18,8 +18,6 @@ package com.hazelcast.jet.impl;
 
 import com.hazelcast.client.impl.ClientEngineImpl;
 import com.hazelcast.core.Member;
-import com.hazelcast.core.MigrationEvent;
-import com.hazelcast.core.MigrationListener;
 import com.hazelcast.instance.BuildInfoProvider;
 import com.hazelcast.instance.HazelcastInstanceImpl;
 import com.hazelcast.instance.JetBuildInfo;
@@ -41,6 +39,9 @@ import com.hazelcast.spi.ConfigurableService;
 import com.hazelcast.spi.LiveOperations;
 import com.hazelcast.spi.LiveOperationsTracker;
 import com.hazelcast.spi.ManagedService;
+import com.hazelcast.spi.MemberAttributeServiceEvent;
+import com.hazelcast.spi.MembershipAwareService;
+import com.hazelcast.spi.MembershipServiceEvent;
 import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.spi.impl.NodeEngineImpl;
 import com.hazelcast.spi.impl.PacketHandler;
@@ -48,17 +49,14 @@ import com.hazelcast.spi.impl.PacketHandler;
 import java.io.IOException;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.Collection;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
-import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 
 public class JetService
         implements ManagedService, ConfigurableService<JetConfig>, PacketHandler, LiveOperationsTracker,
-        CanCancelOperations {
+        CanCancelOperations, MembershipAwareService {
 
     public static final String SERVICE_NAME = "hz:impl:jetService";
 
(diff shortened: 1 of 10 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetService.java
```diff
@@ -208,38 +205,30 @@ public void handle(Packet packet) throws IOException {
...
-
-        @Override
-        public void migrationCompleted(MigrationEvent migrationEvent) {
-        }
-
-        @Override
-        public void migrationFailed(MigrationEvent migrationEvent) {
-        }
+    @Override
+    public void memberRemoved(MembershipServiceEvent event) {
```
cangencer: how reliable is this event? What if the member was removed just as the job was completing?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
```diff
@@ -134,6 +147,30 @@
         return plans;
     }
 
+    private static Address[] initPartitionOwnersAndMembers(NodeEngine nodeEngine, Collection<Member> members) {
```
cangencer: getPartitionOwners would be a better name.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetService.java
mtopolnik: Won't we either get a signal from the member that it's done, or wait indefinitely until this event occurs and breaks the job?

mdogan: This event is a local event and is guaranteed to be fired after a member is removed.

mdogan: Also, target invocations (the ones those are sent to the specific target) are also notified with `MemberLeftException` when target leaves.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
mdogan: This method modifies the given member collection too, not just returns the partition owners.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
```diff
@@ -134,6 +147,30 @@
...
+        for (int partitionId = 0; partitionId < partitionService.getPartitionCount(); partitionId++) {
+            Address address = partitionService.getPartitionOwnerOrWait(partitionId);
+            partitionOwners[partitionId] = address;
+
+            if (!addresses.contains(address)) {
+                // Address in partition table doesn't exist in initial member list.
+                // Check whether it's a newly joined member?
+                Member member = clusterService.getMember(address);
+                if (member == null) {
+                    throw new TopologyChangedException("Topology changed! " + address + " is not member anymore!");
```
cangencer: This logic feels racy to me. What if the member leaves the cluster just after we have made this check? The next operation will fail anyway since it will try to invoke an operation on a member which is not part of the cluster.

If the member wasn't in the initial member list but there is a partition assigned to it - then it must mean it's a new member. This case will only be hit if the new member already leaves the cluster between the call where we got the initial member list and the call where we ask for the owner of the partition. Is there a way we can get a copy of the partition table atomically, and use it as the list of members as well and not use two separate lists which need reconciling?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
cangencer: Fair enough.. but see my comments below regarding having two separate lists.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
mdogan: If member leaves after this point, it's not different than member leaving while task is being executed. It will be handled by invocation system, invoking on a non-existent member will fail.

Logic here guarantees that, if an address is included in partition-owners list then it's going to exist in member-list too. But member list have members those are not in partition table, a member may not have any partition assignment at that moment or even that member may be a lite one.

> If the member wasn't in the initial member list but there is a partition assigned to it - then it must mean it's a new member.

It's either a new member or an old member that has just left but partition table is not updated yet.

PR #307, opened 2017-02-08
Fix behavior and test of ReadFileStreamP

### LINKED ISSUE #298: ReadFileStreamPTest fails intermittently
The tests are currently racy as they rely on file system event timing. 

```Error Message

expected:<3> but was:<0>
Stacktrace

java.lang.AssertionError: expected:<3> but was:<0>
	at com.hazelcast.jet.FileStreamReaderTest.lambda$when_watchType_appendedOnly$5(FileStreamReaderTest.java:107)
	at com.hazelcast.jet.FileStreamReaderTest.when_watchType_appendedOnly(FileStreamReaderTest.java:107)```

### PR DESCRIPTION
**Pull request by viliam-durina**
The processor is still barely useful, its weaknesses are at least better documented.

Fixes #298

### DIFF
--- a/hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/ReadFileStreamPTest.java
+++ b/hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/ReadFileStreamPTest.java
@@ -36,6 +36,9 @@ import java.io.IOException;
 import java.io.PrintWriter;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 
 import static com.hazelcast.jet.Edge.between;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/ReadFileStreamPTest.java [hunk 1]
```diff
@@ -60,65 +62,149 @@ public void setup() throws IOException {
...
+        File file = new File(directory, randomName());
+        appendToFile(file, "hello");
+
+        assertTrueEventually(() -> assertEquals(1, list.size()), 2);
+
+        finishDirectory(jobFuture, file);
+    }
+
+    @Test
+    @Ignore //this does not work reliably due to ENTRY_CREATE might be reported before contents is written
```
cangencer: then it seems WatchType.NEW isn't very useful, if you can't rely that the contents are there. We should get rid of it. I'm also questioning the usefulness of WatchType.REPROCESS. What do you think?

viliam-durina: NEW really isn't. However, maybe, Linux only reports NEW after the file is closed by the writing process, I don't know. Maybe you can test...

REPROCESS is the most correct, if you are ready to remove all data generated by the previous run. However, when line is emitted, you don't know, which file it is from (such as the `docId` needed in `TfIdf`).

APPEND is quite useful, with single limitation, that it can break words.

Anyway, the class is just for quick and dirty tests. Including the fact, that during heavy load the events might OVERFLOW.

PR #308, opened 2017-02-08
Add mandatory map/list name parameter to collector

### LINKED ISSUE #174: [j.u.s] Provide a map name for DistributedCollectors.toIMap
`DistributedCollectors.toIMap` dumps result to map with autogenerated name.
 which not very easy to reference. Of course one can `map.getName()` but in management center it's not very convenient.
Provide a method allows to specify map name.

viliam-durina:
For real use cases, the name is required. At least, the user is aware, that a new map/list
is created.

Fixes #174

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/stream/impl/pipeline/DoublePipeline.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/stream/impl/pipeline/DoublePipeline.java
@@ -44,6 +44,7 @@ import java.util.stream.DoubleStream;
 import java.util.stream.Stream;
 
 import static com.hazelcast.jet.stream.impl.StreamUtil.checkSerializable;
+import static com.hazelcast.jet.stream.impl.StreamUtil.uniqueListName;
 
 @SuppressWarnings("checkstyle:methodcount")
 class DoublePipeline implements DistributedDoubleStream {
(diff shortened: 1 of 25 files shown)


cangencer:
Since this is a breaking change, code samples, user manual and the website will also need to be updated.

viliam-durina:
Samples didn't need any update.

cangencer:
It will: https://github.com/hazelcast/hazelcast-jet-code-samples/blob/master/java.util.stream/wordcount-j.u.s/src/main/java/WordCount.java#L63

hazelcast-jet-core/src/main/java/com/hazelcast/jet/stream/impl/pipeline/DoublePipeline.java
```diff
@@ -136,15 +137,20 @@ public void forEachOrdered(DoubleConsumer action) {
 
     @Override
     public double[] toArray() {
-        IList<Double> list = inner.collect(DistributedCollectors.toIList());
-        double[] array = new double[list.size()];
+        IList<Double> list = inner.collect(DistributedCollectors.toIList(uniqueListName()));
+        try {
+            double[] array = new double[list.size()];
+
+            Iterator<Double> iterator = list.iterator();
```
mtopolnik: Why not simply
```java
int index = 0;
for (Double d : list) {
    array[index++] = d;
}
```
?

viliam-durina: You got it.

PR #310, opened 2017-02-10
Better DAG.toString()

mtopolnik:
Before PR:
```
Vertices {sink=Vertex sink, accumulator=Vertex accumulator, source=Vertex source, combiner=Vertex combiner, doc-lines=Vertex doc-lines, tokenizer=Vertex tokenizer}
Edges [(source, 0) -> (doc-lines, 0), (tokenizer, 0) -> (accumulator, 0), (combiner, 0) -> (sink, 0), (doc-lines, 0) -> (tokenizer, 0), (accumulator, 0) -> (combiner, 0)]
```
After PR:
```
dag
    .vertex("source").localParallelism(1)
    .vertex("doc-lines").localParallelism(1)
    .vertex("tokenize")
    .vertex("reduce")
    .vertex("combine")
    .vertex("sink")
    .edge(between("source", "doc-lines"))
    .edge(between("doc-lines", "tokenize"))
    .edge(between("tokenize", "reduce").partitioned(?))
    .edge(between("reduce", "combine").partitioned(?).distributed())
    .edge(between("combine", "sink"))
```

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/DAG.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/DAG.java
@@ -33,6 +33,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -71,7 +72,7 @@ import static java.util.stream.Collectors.toMap;
  */
 public class DAG implements IdentifiedDataSerializable, Iterable<Vertex> {
 
-    private Set<Edge> edges = new HashSet<>();
+    private Set<Edge> edges = new LinkedHashSet<>();
     private Map<String, Vertex> verticesByName = new HashMap<>();
     private Set<Vertex> verticesByIdentity = newSetFromMap(new IdentityHashMap<Vertex, Boolean>());
 
@@ -209,7 +210,7 @@ public class DAG implements IdentifiedDataSerializable, Iterable<Vertex> {
     }
 
     /**
-     * Returns an iterator over the DAG's vertices in topological order.
+     * Returns an iterator over the DAG's vertices in reverse topological order.
      */
     public Iterator<Vertex> reverseIterator() {
         validate();
@@ -217,7 +218,7 @@ public class DAG implements IdentifiedDataSerializable, Iterable<Vertex> {
     }
 
     /**
-     * Returns an iterator over the DAG's vertices in reverse topological order.
+     * Returns an iterator over the DAG's vertices in topological order.
      */
     @Override
     public Iterator<Vertex> iterator() {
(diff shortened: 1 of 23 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Edge.java
```diff
@@ -317,7 +317,35 @@ public Edge setConfig(EdgeConfig config) {
...
+            if (destOrdinal != 0) {
+                b.append(", ").append(destOrdinal);
+            }
+            b.append(')');
+        }
+        switch (getForwardingPattern()) {
+            case VARIABLE_UNICAST:
+                break;
+            case PARTITIONED:
+                b.append(getPartitioner() instanceof Single ? ".allToOne()" : ".partitioned(?)");
```
viliam-durina: I would output `.partitioned(...)` here.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/DAG.java
```diff
@@ -229,7 +230,19 @@ public Vertex getVertex(String vertexName) {
...
     @Override
     public String toString() {
-        return "Vertices " + verticesByName + "\nEdges " + edges;
+        final StringBuilder b = new StringBuilder("dag\n");
+        for (Iterator<Vertex> it = iterator(); it.hasNext();) {
+            final Vertex v = it.next();
+            b.append("    .vertex(").append(v.getName());
+            if (v.getLocalParallelism() != -1) {
+                b.append(".localParallelism(").append(v.getLocalParallelism()).append(')');
+            }
```
viliam-durina: This outputs:

    .vertex(vertexName.localParallelism(1))

Better would be:

    .vertex(vertexName)
        .localParallelism(1)

Next, I would prefer the vertexName to be in quotes.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Edge.java
mtopolnik: That was my first version, but i didn't like:
1. the extra characters that give no information
2. that it's a dot, which is used in the syntax elsewhere. It's not easily distinguishable.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/DAG.java
mtopolnik: Would you also then have quotes in `between("a", "b")`? Seems like noise to me, what's the advantage? This isn't well-formed Java anyway.

viliam-durina: Yes. Less characters don't always mean it's easier to read. Advantage: you clearly know, that it is user-provided text (even though generated in case of j.u.s).

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Edge.java
viliam-durina: 1. Ellipsis (...) means, that there is something to continue. We know, what it is, but just don't write it here.
2. I can easily distinguish between "." and "..." :). Notice the usage of quotes :).

viliam-durina: The "..." better says "more code here", than "?". Anyway, I told to myself not to argue about the style. Both will do the thing, use whatever. My vote is for ... and ".

hazelcast-jet-core/src/main/java/com/hazelcast/jet/DAG.java
mtopolnik:
```java
.vertex("stopword-source".localParallelism(1))
```
This stings my eyes more than
```java
.vertex(stopword-source.localParallelism(1))
```
I can squint at the fact that `stopword-source` is not a legal Java identifier, but seeing a `localParallelism()` method call on a `String` looks just plain wrong.

mtopolnik: About the `.localParallelism` on its own line, i tried it, but it makes the log entry take more vertical space. I'm worried about vertical space even without this change.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Edge.java
mtopolnik: `.edge(between(combine, xcombine).partitioned(...).distributed())`

Just... so many dots in a row. But OK.

Issue #311, opened 2017-02-13
Replace `ReadFileStreamP` with batch file processor


viliam-durina:
The current API tries to use `WatchService` for watching changes to files in a directory and stream them. However, it's very unreliable, and thus unusable beyond some may-be-working example. Stream-processing of log files cannot be reliably done using files.

We need to replace it with a processor, that will process all files in a directory and then finish. This will handle the use case for batch processing of log files, for example.

viliam-durina:
`APPENDED_ONLY` option could be made correct, if we only read lines, that are terminated with a newline. This way, we well avoid reading incomplete lines.

viliam-durina:
Batch processor added in #373, while also keeping the streaming processor

PR #312, opened 2017-02-13
Deprecate ReadFileStreamP

viliam-durina:
Also unrelated error message improvement

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/DAG.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/DAG.java
@@ -156,12 +156,16 @@ public class DAG implements IdentifiedDataSerializable, Iterable<Vertex> {
         if (getInboundEdges(edge.getDestName())
                 .stream().anyMatch(e -> e.getDestOrdinal() == edge.getDestOrdinal())) {
             throw new IllegalArgumentException("Vertex '" + edge.getDestName()
-                    + "' already has an inbound edge at ordinal " + edge.getDestOrdinal());
+                    + "' already has an inbound edge at ordinal " + edge.getDestOrdinal()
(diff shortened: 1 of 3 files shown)


cangencer:
I believe there isn't any reason to deprecate this class, since it's in an impl package and was never part of public API. It sounds to me like it should be deleted outright if it can't be fixed to work in a reliable way.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/DAG.java
```diff
@@ -156,12 +156,14 @@ public DAG edge(Edge edge) {
         if (getInboundEdges(edge.getDestName())
                 .stream().anyMatch(e -> e.getDestOrdinal() == edge.getDestOrdinal())) {
             throw new IllegalArgumentException("Vertex '" + edge.getDestName()
-                    + "' already has an inbound edge at ordinal " + edge.getDestOrdinal());
+                    + "' already has an inbound edge at ordinal " + edge.getDestOrdinal()
+                    + ", use Edge.from().to() to specify an ordinal");
```
mtopolnik: This doesn't seem extra-helpful to me. The same message will be used regardless of the syntax used to create the edge. The user may have written `from(a, 1).to(b, 2)` and still get the message. Maybe it can be added only when we see that the edge is configured as 0-to-0?

viliam-durina: It's good for beginners, as I myself first didn't know, what to do. I'll try to elaborate.

viliam-durina:
If it's meant to be internal, then delete it right away, it's not used anywhere. I wanted to deprecate it for now, until it's replaced with a Processor that will read all files in a directory, as per #311.

cangencer:
I'm fine with deleting it for now if it's unusable. Can you bring this up with the PM ?

PR #314, opened 2017-02-14
TransformP in terms of Traverser

mtopolnik:
`TransformP` implemented the same logic that was later encapsulated into `Traverser` and `AbstractProcessor`. This PR redoes the logic in terms of existing API.


mtopolnik:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/stream/impl/pipeline/TransformPipeline.java
```diff
@@ -32,24 +32,22 @@
 
     private final List<TransformOperation> operations = new ArrayList<>();
 
-    public TransformPipeline(StreamContext context,
```
viliam-durina: Didn't you forget to remove `public` here?

PR #328, opened 2017-02-27
Added key/value mappers for WriteHdfsP

gurbuzali:


mtopolnik:
Not sure if this is the direction we want to take. It's a single-purpose solution to the problem we intend to solve in general, by providing optimized one-to-one edges which would merge the processors into a single tasklet.

cangencer:
Agree with Marko - but I think the aim is to make it immediately useful. In general, there is a need for mapping when dealing with sources and sinks and having to create yet another vertex for this purpose seems cumbersome.

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/connector/hadoop/WriteHdfsP.java
```diff
@@ -89,18 +94,38 @@ public boolean isCooperative() {
...
 
+    /**
+     * Returns a meta-supplier of processors that write HDFS files.
+     *
+     * @param jobConf     JobConf for writing files with the appropriate output format and path
+     * @param keyMapper   mapper which can be used to map the key to another value
+     * @param valueMapper mapper which can be used to map the value to another value
+     * @return {@link ProcessorMetaSupplier} supplier
+     */
+    public static <T, R, K, V> ProcessorMetaSupplier writeHdfs(JobConf jobConf,
```
cangencer: it would be better if type parameters were named "K, V, RK, RV" or something similar.

Javadoc for params is also missing

cangencer: Javadoc should also include comments about `Writable` requirements in certain cases.

gurbuzali: 👍

PR #337, opened 2017-03-06
ReadFileStreamP should not emit incomplete lines

viliam-durina:
Previously, lines from appended files could be emitted in two chunks, if we
read the file with concurrent writing. Also, only appended lines from
pre-existing files will be read.


cangencer:
Does this fix the intermittently failing tests?

viliam-durina:
No. I fact, I don't really understand, why they fail.

viliam-durina:
verify

viliam-durina:
Don't merge, more stuff will come here.

cangencer:
verify

PR #342, opened 2017-03-09
In case of highwater, seek to the latest offset

gurbuzali:
added random groupId for the test


hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/connector/kafka/ReadKafkaP.java
```diff
@@ -76,6 +75,7 @@ private ReadKafkaP(String[] topicIds, Properties properties) {
      *                   broker address and key/value deserializers
      */
     public static <K, V> ProcessorMetaSupplier readKafka(Properties properties, String... topicIds) {
+        assert properties.containsKey("group.id") : "Properties should contain `group.id`";
```
cangencer: should throw an exception here, rather than assert.

gurbuzali: 👍

Issue #353, opened 2017-03-16
Support adding `JobConfig` to java.util.stream API


cangencer:
Currently there is no way to pass a `JobConfig` object to a distributed stream.

gurbuzali:
how about this https://github.com/gurbuzali/hazelcast-jet/commit/f617778a01d13a03552649872082c6107cc0b9d6 ?

PR #359, opened 2017-03-21
Added configuration option to DistributedStream

gurbuzali:
fixes https://github.com/hazelcast/hazelcast-jet/issues/353


gAmUssA:
Quick Q @gurbuzali 
why `map.stream.configure` but not `map.stream(config)` or `map.stream().withConfig(config)` ?

cangencer:
looks good, but missing tests

@gAmUssA we don't want to overload .stream() since there will come more overloads when we introduce `Predicate` and `Projection` support on `IMap`. `withConfig` vs `configure` is just a matter of style, `configure` is keeping with more with general j.u.s. naming I believe (which are all verbs)

PR #361, opened 2017-03-23
Fix ResourceIterator and add test

viliam-durina:
It had issues:
- single use ill-defined Util.read() method
- improper use of InputStream.available() to detect EOF
- didn't handle empty files (not a requirement, but anyway)

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/ResourceIterator.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/ResourceIterator.java
@@ -20,23 +20,27 @@ import com.hazelcast.jet.config.ResourceConfig;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 import java.util.Set;
 
-import static com.hazelcast.jet.impl.util.Util.read;
 import static com.hazelcast.jet.impl.util.ExceptionUtil.rethrow;
 
 public final class ResourceIterator implements Iterator<ResourcePart> {
-    private final int partSize;
+
     private final Iterator<ResourceConfig> configIterator;
     private InputStream inputStream;
     private ResourceConfig resourceConfig;
     private int offset;
+    private byte[] buffer;
+    private ResourcePart nextPart;
 
     public ResourceIterator(Set<ResourceConfig> resourceConfigs, int partSize) {
         this.configIterator = resourceConfigs.iterator();
-        this.partSize = partSize;
+
+        buffer = new byte[partSize];
+        readNext();
     }
 
     private void switchFile() throws IOException {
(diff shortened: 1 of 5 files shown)


hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/deployment/ResourceIteratorTest.java
```diff
@@ -0,0 +1,116 @@
...
+            }
+            assertEquals("offsets not in sequence", lastOffset, rp.getOffset());
+            lastOffset += rp.getBytes().length;
+            assertEquals("contents don't match", contents[fileIndex].substring(rp.getOffset(), lastOffset), new String(rp.getBytes()));
+        }
+        assertEquals("not all config files read", lastIndex, numFiles - 1);
+        assertEquals("config not fully read", contents[lastIndex].length(), lastOffset);
+
+        // delete the files
+        for (Path file : files) {
```
cangencer: is this necessary? you already have deleteOnExit()

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/ResourceIterator.java
```diff
@@ -64,33 +64,41 @@ public void remove() {
...
-                ResourcePart part = new ResourcePart(resourceConfig.getDescriptor(), bytes, offset);
-                offset += bytes.length;
-                return part;
-            } else {
-                throw new NoSuchElementException();
+                int bytesRead = inputStream.read(buffer);
+
+                if (bytesRead < 0) {
+                    bytesRead = 0;
+                    inputStream.close();
```
cangencer: is inputStream closed when there is some exception?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/ResourceIterator.java
```diff
@@ -20,23 +20,27 @@
...
 
     public ResourceIterator(Set<ResourceConfig> resourceConfigs, int partSize) {
         this.configIterator = resourceConfigs.iterator();
-        this.partSize = partSize;
+
+        buffer = new byte[partSize];
+        readNext();
     }
 
     private void switchFile() throws IOException {
```
cangencer: is the configIterator.hasNext() check still necessary here?

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/deployment/ResourceIteratorTest.java
viliam-durina: it's like early delete. If you run all the tests, they will be deleted earlier. I moved it to finally and remove deleteOnExit

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/ResourceIterator.java
viliam-durina: It's not, it's also not, when we don't finish the iteration, if the exception is in our caller.

viliam-durina:
verify

viliam-durina:
verify

PR #367, opened 2017-03-29
Shade Hazelcast JARs into Jet

cangencer:


cangencer:
This should be fixed from Hazelcast side: https://github.com/hazelcast/hazelcast/issues/10177

PR #375, opened 2017-04-04
Outbox.offer instead of add

mtopolnik:
Replace `isHighWater()` and `add()` with just `offer()` which refuses to accept an item when full, returning `false`.


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Outbox.java
```diff
@@ -58,20 +58,20 @@ default void add(@Nonnull Object item) {
 
     /**
      * Returns {@code true} if any of this outbox's buckets is above its
```
viliam-durina: Should be "is full" instead of "is above its high water mark"

viliam-durina: We should also document the exception thrown from `add` methods in javadoc.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processor.java
```diff
@@ -38,32 +38,29 @@
...
- * cooperative multithreading scheme. The processing methods should also
+ * cooperative multithreading scheme. The processing methods must also
  * limit the amount of data they output per invocation because the outbox
  * will not be emptied until the processor yields control back to its
- * caller. Specifically, {@code Outbox} has a method {@link
- * Outbox#isHighWater isHighWater()} that can be tested to see whether it's
- * time to stop pushing more data into it.  There is also a finer-grained
- * method {@link Outbox#isHighWater(int) isHighWater(ordinal)}, which tells
- * the state of an individual output bucket.
+ * caller. Outbox's {@code add(...)} methods return {@code true} when it is
```
viliam-durina: They throw exception, they don't return `true`.

hazelcast-jet-core/src/main/resources/hazelcast-jet-default.xml
```diff
@@ -26,8 +26,8 @@
         <!-- capacity of the concurrent SPSC queue between each two processors -->
        <queue-size>1024</queue-size>
 
-        <!-- number of slots before hitting the high water mark in the outbox -->
-       <high-water-mark>2048</high-water-mark>
+        <!-- sumber of slots in each outbox's bucket -->
```
viliam-durina: Typo here.

hazelcast-jet-core/src/main/resources/hazelcast-jet-full-example.xml
```diff
@@ -33,8 +33,8 @@
         <!-- number of available slots for each concurrent queue between two vertices -->
        <queue-size>1024</queue-size>
 
-        <!-- number of slots before high water is triggered for the outbox -->
-       <high-water-mark>2048</high-water-mark>
+        <!-- sumber of slots in each outbox's bucket -->
```
viliam-durina: Typo here.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Outbox.java
mtopolnik: Exception is not thrown from `add`, threre's just an `assert`.

viliam-durina: Yes, but we should document, that the call fails in case of full outbox. Assertions are disabled by default and some might not notice, until after they enable them.

I would just throw `JetException`. Assertions should be used to check for our programming errors, not for user errors. This assertion is easily violated by the user.

mtopolnik:
All tests were failing due to a bug in a single class (`SkipP`).

mtopolnik:
verify

PR #376, opened 2017-04-05
Add WriteFileP

viliam-durina:
A processor to write to single file, one item.toString() per line.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
@@ -29,11 +29,11 @@ import com.hazelcast.jet.impl.connector.WriteFileP;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.BufferedWriter;
+import java.io.File;
 import java.io.OutputStreamWriter;
 import java.net.Socket;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -336,7 +336,7 @@ public final class Processors {
         String fileNameSuffix = "";
 
         // if the file name has an extension, use it as a suffix
-        String fileNameWithoutPath = Paths.get(file).getFileName().toString();
+        String fileNameWithoutPath = new File(file).getName();
         int lastDot = fileNameWithoutPath.lastIndexOf('.');
         if (lastDot > 0) {
             // non-zero is intentional, to not consider files starting with '.' as extension
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
@@ -33,6 +33,7 @@ import java.io.OutputStreamWriter;
 import java.net.Socket;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -307,12 +308,12 @@ public final class Processors {
      * UTF-8 charset, overwriting the target file and no early file buffer flush.
      */
     @Nonnull
-    public static ProcessorSupplier writeFile(@Nonnull String file) {
+    public static ProcessorMetaSupplier writeFile(@Nonnull String file) {
         return writeFile(file, null, false, false);
     }
 
     /**
-     * Returns a supplier of processor, that writes all items to a local file on
+     * Returns a meta-supplier of processor, that writes all items to a local file on
      * each member. {@code item.toString()} is written to the file, followed by
      * {@code '\n'} (both on linux and Windows).
      * <p>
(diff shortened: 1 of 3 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
```diff
@@ -260,6 +263,38 @@ public static ProcessorSupplier readSocket(@Nonnull String host, int port) {
...
+
+    /**
+     * Returns a supplier of processor, that writes all items to a local file on
+     * each member. {@code item.toString()} is written to the file, followed by
+     * {@code '\n'} (both on linux and Windows).
+     * <p>
+     * The same file must be available for writing on all nodes. The file on
+     * each node will contain part of the data processed on that member.
+     * <p>
+     *  The vertex should have {@link Vertex#localParallelism(int) local parallelism} of 1.
```
gurbuzali: if you start 2 nodes in a single machine, each node will try to write to same file

viliam-durina: is there a way to avoid that? And shall we? One only does it for test.

gurbuzali:
I think we can make use of `WriteBufferedP` for this

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
gurbuzali: imagine you have a big machine which have a lot of memory, you can't have a single JVM which uses all the memory but you start multiple JVM instances so that you don't suffer from big GC pauses. so there are uses-cases which you start 2 nodes in a single machine

gurbuzali:
with `WriteBuffered`
```
public class WriteFileP {

    public static ProcessorMetaSupplier writeFile(String fileName) {
        return addresses -> address -> count -> range(0, count)
                .mapToObj(i -> new WriteBufferedP<>(
                        () -> createBufferedWriter(fileName, address, i),
                        (bufferedWriter, o) -> uncheckRun(() -> bufferedWriter.write(o.toString())),
                        bufferedWriter -> uncheckRun(bufferedWriter::flush),
                        bufferedWriter -> uncheckRun(bufferedWriter::close)
                )).collect(Collectors.toList());
    }

    private static BufferedWriter createBufferedWriter(String fileName, Address address, int processorId) {
        String name = fileName + "-" + address + "-" + processorId;
        return uncheckCall(() -> new BufferedWriter(new OutputStreamWriter(new FileOutputStream(name), "UTF-8")));
    }

}
```

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
cangencer: Have a look at the approach here: https://github.com/hazelcast/hazelcast-jet-code-samples/tree/master/core/map-dump

It writes to a folder, with every processor writing to own file - this is also how it's typically done with HDFS.

viliam-durina:
@gurbuzali done

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
viliam-durina: What's the use case for that? You'll have randomly split files.

cangencer: If it's text files, you can easily `cat` them together. If you don't want split files, you can set parallelism to 1.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteFileP.java
```diff
@@ -0,0 +1,64 @@
...
+        return count -> {
+            if (count > 1) {
+                throw new JetException(WriteFileP.class.getSimpleName() + " must have localParallelism=1");
+            }
+            // create the processors
+            return Collections.singletonList(new WriteBufferedP<>(
+                    () -> createBufferedWriter(fileName, charset, append),
+                    (writer, item) -> uncheckRun(() -> {
+                        writer.write(item.toString());
+                        writer.write('\n');
```
cangencer: should be `writer.newLine()` to use system new line.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteFileP.java
```diff
@@ -0,0 +1,64 @@
...
+ * @see com.hazelcast.jet.Processors#writeFile(String, Charset, boolean, boolean)
+ */
+public final class WriteFileP {
+
+    private WriteFileP() { }
+
+    public static ProcessorSupplier supplier(String fileName, String charset, boolean append, boolean flushEarly) {
+        return count -> {
+            if (count > 1) {
+                throw new JetException(WriteFileP.class.getSimpleName() + " must have localParallelism=1");
```
cangencer: I feel this is a rather arbitrary restriction - furthermore it will not work across multiple nodes running on same JVM.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
viliam-durina: Feature added.

viliam-durina:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteFileP.java
```diff
@@ -0,0 +1,82 @@
...
+            return count -> IntStream.range(0, count)
+                    .mapToObj(index -> new WriteBufferedP<>(
+                            () -> createBufferedWriter(createFileName(fileNamePrefix, fileNameSuffix, hasMultipleAddresses,
+                                    sAddress, count, index), charset, append),
+                            (writer, item) -> uncheckRun(() -> {
+                                writer.write(item.toString());
+                                writer.newLine();
+                            }),
+                            flushEarly ? writer -> uncheckRun(writer::flush) : writer -> {
+                            },
```
viliam-durina: Was looking for that. Only the `noopConsumer` is unnecessarily serializable, it has some overhead, but not much i believe...

PR #379, opened 2017-04-07
Exploit new blocking outbox in StreamKafkaP

viliam-durina:
Also add cancellation check to test and rename to new convention


cangencer:
verify

hasancelik:
verify

PR #381, opened 2017-04-10
JobFuture in Processor.Context

viliam-durina:
The need is to allow for non-cooperative processor to check for job cancellation/failure, without returning from `complete()` method.

Solution moves the call to `Processor.init` from `Tasklet` constructor, which is called from `ExecutionPlan.initialize()` to `Tasklet.init()` method, because only then the `jobFuture` is available.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processor.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processor.java
@@ -154,12 +154,12 @@ public interface Processor {
          * after each blocking call (note, that blocking calls are allowed only
          * in {@link #isCooperative() non-cooperative} processors). In this case,
          * the methods should regularly check the {@code jobFuture}'s
-         * {@link CompletableFuture#isCompletedExceptionally()} and return, when
-         * it returns {@code true}:
+         * {@link CompletableFuture#isDone() isDone()} and return, when it returns
+         * {@code true}:
          *
          * <pre>
          * public boolean complete() {
-         *     while (!jobFuture.isCompletedExceptionally()) {
+         *     while (!jobFuture.isDone()) {
          *         // we should not block indefinitely, but rather with a timeout
          *         Collection data = blockingRead(timeout);
          *         for (Object item : data) {
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processor.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processor.java
@@ -19,6 +19,7 @@ package com.hazelcast.jet;
 import com.hazelcast.logging.ILogger;
 
 import javax.annotation.Nonnull;
+import java.util.concurrent.CompletableFuture;
 
 /**
  * Does the computation needed to transform zero or more input data streams
@@ -144,5 +145,29 @@ public interface Processor {
          */
         @Nonnull
         String vertexName();
+
+        /**
+         * Returns the future to check for cancellation status.
+         * <p>
+         * This is only necessary, if the {@link #complete()} method never returns and blocks
+         * indefinitely, which is only legal for {@link #isCooperative() non-cooperative} processors.
+         * In this case, the {@link #complete()} should check regularly the {@code jobFuture}'s
+         * {@link CompletableFuture#isCompletedExceptionally()} and return, when it returns
+         * {@code true}:
+         *
+         * <pre>
+         * public boolean complete() {
+         *     while (!jobFuture.isCompletedExceptionally()) {
+         *         // we should not block indefinitely, but rather with a timeout
+         *         Collection data = blockingRead(timeout);
+         *         for (Object item : data) {
+         *             emit(item);
+         *         }
+         *     }
+         * }
+         * </pre>
+         */
+        @Nonnull
+        CompletableFuture<Void> jobFuture();
     }
 }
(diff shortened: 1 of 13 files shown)


cangencer:
verify

viliam-durina:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTaskletBase.java
```diff
@@ -44,15 +47,18 @@
...
 
-    private final Context context;
     private final ArrayDequeInbox inbox = new ArrayDequeInbox(progTracker);
     private final Queue<ArrayList<InboundEdgeStream>> instreamGroupQueue;
     private final String vertexName;
     private CircularListCursor<InboundEdgeStream> instreamCursor;
 
-    ProcessorTaskletBase(String vertexName, Processor.Context context, Processor processor,
-                         List<InboundEdgeStream> instreams, List<OutboundEdgeStream> outstreams) {
+    ProcessorTaskletBase(String vertexName, JetInstance instance, ILogger logger, Processor processor,
```
mtopolnik: I couldn't find any usages of the `JetInstance`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processor.java
```diff
@@ -144,5 +145,29 @@ default boolean isCooperative() {
...
         @Nonnull
         String vertexName();
+
+        /**
+         * Returns the future to check for cancellation status.
+         * <p>
+         * This is only necessary, if the {@link #complete()} method never returns and blocks
+         * indefinitely, which is only legal for {@link #isCooperative() non-cooperative} processors.
+         * In this case, the {@link #complete()} should check regularly the {@code jobFuture}'s
+         * {@link CompletableFuture#isCompletedExceptionally()} and return, when it returns
```
mtopolnik: It can also check `future.isDone()` which is a more general check.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processor.java
```diff
@@ -144,5 +145,29 @@ default boolean isCooperative() {
          */
         @Nonnull
         String vertexName();
+
+        /**
+         * Returns the future to check for cancellation status.
+         * <p>
+         * This is only necessary, if the {@link #complete()} method never returns and blocks
```
mtopolnik: This is not the only case, it is enough that the processor does more than one blocking call per processor call.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processor.java
```diff
@@ -144,5 +145,29 @@ default boolean isCooperative() {
...
+         * indefinitely, which is only legal for {@link #isCooperative() non-cooperative} processors.
```
mtopolnik: Could you please format this text so the length of the text line is up to 72 chars? This doesn't include any chars preceding the first char of actual text (it's not about the right margin, but actual text line length).

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTaskletBase.java
viliam-durina: It goes to the `Processor.Context`:

    void initProcessor(Outbox outbox, CompletableFuture<Void> jobFuture) {
        ProcCtx context = new ProcCtx(instance, logger, vertexName, processorIdx, jobFuture);
        processor.init(outbox, context);
    }

mtopolnik: I tried to study the code to see where it's actually used, but I couldn't find it.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processor.java
viliam-durina: updated

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processor.java
viliam-durina: ok

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTaskletBase.java
cangencer: it's used in creating Context I believe. I think Logger creation here should also be moved to `init` method.

mtopolnik: I see; no usage of JetInstance occurs in this PR because it was there in the Context before, but since the point in the lifecycle where init is done changed, now it has to be propagated to that new place.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/Contexts.java
```diff
@@ -65,6 +67,22 @@ public int index() {
...
+         * which means it's <i>practically non-null</i>.
+         */
+        @Nonnull
+        @Override
+        public CompletableFuture<Void> jobFuture() {
+            return jobFuture;
+        }
+
+        public void initJobFuture(CompletableFuture<Void> jobFuture) {
+            assert this.jobFuture == null;
```
mtopolnik: We should supply a message here.

viliam-durina:
verify

viliam-durina:
verify

viliam-durina:
verify

PR #383, opened 2017-04-11
Fix bug in processor tasklets

mtopolnik:
Ensure DONE_ITEM submitted only once


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/ArrayDequeOutbox.java
```diff
@@ -90,4 +90,10 @@ private boolean isBucketFull(int ordinal) {
     public Queue<Object> queueWithOrdinal(int ordinal) {
         return buckets[ordinal];
     }
+
+    public void addIgnoringCapacity(Object item) {
```
viliam-durina: I would add a comment here: why do we need this:
"Adds an item to all buckets while ignoring the capacity limit. It's added to simplify some code and should be used with care and for limited number of items".

PR #385, opened 2017-04-13
Fix right margin

mtopolnik:
Non-semantic change that sets CheckStyle's right margin to 123 and fixes the resulting errors. This is a backport of the same change on the `windowing` branch.


mtopolnik:
Since this is formatting-only PR and the tests have passed, I'll merge it now.

PR #390, opened 2017-04-20
Infinite stream processing

mtopolnik:
Merges the work done on a parallel `windowing` branch.


viliam-durina:
verify

.gitignore
```diff
@@ -21,5 +21,7 @@ performance*.log
 dependency-reduced-pom.xml
 *.jfr
 
+# flight recorder
```
cangencer: duplicated above

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ExecutionService.java
```diff
@@ -136,6 +138,15 @@ private synchronized void ensureThreadsStarted() {
         Arrays.stream(threads).forEach(Thread::start);
     }
 
+    private String trackersToString() {
```
viliam-durina: Unused method

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Edge.java
```diff
@@ -335,11 +349,14 @@ public String toString() {
             case VARIABLE_UNICAST:
                 break;
             case PARTITIONED:
-                b.append(getPartitioner() instanceof Single ? ".allToOne()" : ".partitioned(...)");
+                b.append(getPartitioner() instanceof Single ? ".allToOne()" : ".partitioned(?)");
```
mtopolnik: I've given it a chance and `?` is just better, especially because the dot is also used with the usual Java meaning.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ExecutionService.java
mtopolnik: I'm keeping this method to be used when needed while debugging.

PR #391, opened 2017-04-20
Make passing of punctuation default

viliam-durina:
Also some minor tweaks

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteBufferedP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteBufferedP.java
@@ -53,11 +53,11 @@ public final class WriteBufferedP<B, T> implements Processor {
 
     @Override
     public void process(int ordinal, @Nonnull Inbox inbox) {
-        for (Object o; (o = inbox.poll()) != null; ) {
-            if (!(o instanceof Punctuation)) {
-                addToBuffer.accept(buffer, (T) o);
+        inbox.drain(item -> {
+            if (!(item instanceof Punctuation)) {
+                addToBuffer.accept(buffer, (T) item);
             }
-        }
+        });
         flushBuffer.accept(buffer);
     }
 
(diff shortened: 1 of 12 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteBufferedP.java
```diff
@@ -52,7 +53,11 @@
 
     @Override
     public void process(int ordinal, @Nonnull Inbox inbox) {
-        inbox.drain((T t) -> addToBuffer.accept(buffer, t));
+        for (Object o; (o = inbox.poll()) != null; ) {
+            if (!(o instanceof Punctuation)) {
+                addToBuffer.accept(buffer, (T) o);
+            }
+        }
```
mtopolnik: The code would be simpler if it kept the `drain(Consumer)` call.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/WindowOperation.java
```diff
@@ -83,13 +88,22 @@
      * #combineAccumulatorsF() combine} so that {@code deduct(combine(acc, x),
      * x)} returns an accumulator in the same state as {@code acc} was before
      * the operation.
+     * <p>
+     * <b>Note:</b> this operation is optional and can return {@code null}. If
```
mtopolnik: It's not the operation that can return `null`. Also the text should consistently use either "operation" or "function" to refer to the object. Since we use "operation" for the entire type (a tuple of functions), "function" would be better.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteBufferedP.java
viliam-durina: Not so much simpler: 

        inbox.drain(item -> {
            if (!(item instanceof Punctuation)) {
                addToBuffer.accept(buffer, (T) item);
            }
        });

And with plain old java it's easier to read.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/WindowOperation.java
viliam-durina: I know, but couldn't come up with better wording. I improved it a bit.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteBufferedP.java
cangencer: I also prefer the `drain(Consumer)` call. You're repeating the same for loop which is inside that call already.

viliam-durina:
breaking change: it does, very slightly. I removed the type parameter from `countDistinct` :). Anyway, I removed the label.

PR #392, opened 2017-04-21
Increase test coverage for SlidingWP and GBFP

viliam-durina:
Improve tests:

- always check for memory leaks
- test SlidingWindowP with both with and without `deductF`
- compare expected and actual outbox as long string: easier to see differences

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Traverser.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Traverser.java
@@ -125,6 +125,27 @@ public interface Traverser<T> {
     }
 
     /**
+     * Returns a traverser that will emit the same items as this traverser and
+     * will additionally run the supplied action first time this traverser
+     * returns {@code null}.
+     */
+    @Nonnull
+    default Traverser<T> onFirstNull(@Nonnull Runnable action) {
+        return new Traverser<T>() {
+            private Runnable action2 = action;
+            @Override
+            public T next() {
+                T t = Traverser.this.next();
+                if (t == null && action2 != null) {
+                    action2.run();
+                    action2 = null;
+                }
+                return t;
+            }
+        };
+    }
+
+    /**
      * Returns a traverser that will emit the same items as this traverser,
      * additionally passing each item to the supplied consumer. A {@code null}
      * return value is not passed to the action.
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Traverser.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Traverser.java
@@ -125,27 +125,6 @@ public interface Traverser<T> {
     }
 
     /**
-     * Returns a traverser that will emit the same items as this traverser and
-     * will additionally run the supplied action first time this traverser
-     * returns {@code null}.
-     */
-    @Nonnull
-    default Traverser<T> onFirstNull(@Nonnull Runnable action) {
-        return new Traverser<T>() {
-            private Runnable action2 = action;
-            @Override
-            public T next() {
-                T t = Traverser.this.next();
-                if (t == null && action2 != null) {
-                    action2.run();
-                    action2 = null;
-                }
-                return t;
-            }
-        };
-    }
-
-    /**
      * Returns a traverser that will emit the same items as this traverser,
      * additionally passing each item to the supplied consumer. A {@code null}
      * return value is not passed to the action.
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/SlidingWindowP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/SlidingWindowP.java
@@ -33,6 +33,7 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.LongStream;
 
+import static com.hazelcast.jet.Distributed.Function.identity;
 import static java.lang.Math.min;
 
 /**
@@ -102,38 +103,34 @@ class SlidingWindowP<K, F, R> extends AbstractProcessor {
     }
 
     private Traverser<Object> slidingWindowTraverser(Punctuation punc) {
-        long rangeStart = nextFrameSeqToEmit;
-        nextFrameSeqToEmit = wDef.higherFrameSeq(punc.seq());
-        return Traversers.traverseStream(range(rangeStart, nextFrameSeqToEmit, wDef.frameLength()).boxed())
-                .flatMap(frameSeq -> Traversers.traverseIterable(computeWindow(frameSeq).entrySet())
-                        .map(e -> (Object) new Frame<>(frameSeq, e.getKey(), finishF.apply(e.getValue())))
-                        .onFirstNull(() -> completeWindow(frameSeq)))
-                .append(punc);
+        Traverser<Object> traverser = Traversers.traverseStream(
+                range(nextFrameSeqToEmit, nextFrameSeqToEmit = wDef.higherFrameSeq(punc.seq()), wDef.frameLength())
+                        .mapToObj(frameSeq ->
+                                computeWindow(frameSeq, seqToKeyToFrame.remove(frameSeq - wDef.windowLength()))
+                                        .entrySet().stream()
+                                        .map(e -> new Frame<>(frameSeq, e.getKey(), finishF.apply(e.getValue()))))
+                        .flatMap(identity()));
+
+        traverser = traverser.append(punc);
+        return traverser;
     }
 
-    private void completeWindow(long frameSeq) {
-        Map<K, F> evictedFrame = seqToKeyToFrame.remove(frameSeq - wDef.windowLength() + wDef.frameLength());
-        if (deductF != null) {
-            // deduct trailing-edge frame
-            patchSlidingWindow(deductF, evictedFrame);
-        }
-    }
-
-    private Map<K, F> computeWindow(long frameSeq) {
+    private Map<K, F> computeWindow(long frameSeq, Map<K, F> evictedFrame) {
         if (deductF != null) {
             // add leading-edge frame
             patchSlidingWindow(combineF, seqToKeyToFrame.get(frameSeq));
+            // deduct trailing-edge frame
+            patchSlidingWindow(deductF, evictedFrame);
             return slidingWindow;
-        } else {
-            // without deductF we have to recompute the window from scratch
-            Map<K, F> window = new HashMap<>();
-            Map<Long, Map<K, F>> frames = seqToKeyToFrame.subMap(frameSeq - wDef.windowLength(), false, frameSeq, true);
-            for (Map<K, F> keyToFrame : frames.values()) {
-                keyToFrame.forEach((key, currAcc) ->
-                        window.compute(key, (x, acc) -> combineF.apply(acc != null ? acc : createF.get(), currAcc)));
-            }
-            return window;
         }
+        // without deductF we have to recompute the window from scratch
+        Map<K, F> window = new HashMap<>();
+        Map<Long, Map<K, F>> frames = seqToKeyToFrame.subMap(frameSeq - wDef.windowLength(), false, frameSeq, true);
+        for (Map<K, F> keyToFrame : frames.values()) {
+            keyToFrame.forEach((key, currAcc) ->
+                    window.compute(key, (x, acc) -> combineF.apply(acc != null ? acc : createF.get(), currAcc)));
+        }
+        return window;
     }
 
     private void patchSlidingWindow(BinaryOperator<F> patchOp, Map<K, F> patchingFrame) {
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/SlidingWindowP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/SlidingWindowP.java
@@ -32,9 +32,9 @@ import java.util.function.BinaryOperator;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.LongStream;
+import java.util.stream.Stream;
 
 import static java.lang.Math.min;
-import static java.util.function.Function.identity;
 
 /**
  * Sliding window processor. See {@link
@@ -47,6 +47,12 @@ import static java.util.function.Function.identity;
  * @param <R> type of the finished result
  */
 class SlidingWindowP<K, F, R> extends AbstractProcessor {
+
+    // package-visible for test
+    final NavigableMap<Long, Map<K, F>> seqToKeyToFrame = new TreeMap<>();
+    final Map<K, F> slidingWindow = new HashMap<>();
+
+    private final Traverser<Object> nullTraverser = Traversers.newNullTraverser();
     private final WindowDefinition wDef;
     private final Supplier<F> createF;
     private final BinaryOperator<F> combineF;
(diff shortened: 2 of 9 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/SlidingWindowP.java
```diff
@@ -100,34 +102,38 @@ protected boolean tryProcessPunc0(@Nonnull Punctuation punc) {
...
+                        .map(e -> (Object) new Frame<>(frameSeq, e.getKey(), finishF.apply(e.getValue())))
+                        .onFirstNull(() -> completeWindow(frameSeq)))
+                .append(punc);
     }
 
-    private Map<K, F> computeWindow(long frameSeq, Map<K, F> evictedFrame) {
+    private void completeWindow(long frameSeq) {
+        Map<K, F> evictedFrame = seqToKeyToFrame.remove(frameSeq - wDef.windowLength() + wDef.frameLength());
         if (deductF != null) {
-            // add leading-edge frame
```
mtopolnik: I disagree with these changes and think they should be reverted. The code looked much more natural and obvious before, and the supposed benefit is one frame out of a typical hundred or thousand evicted slightly earlier.

viliam-durina: Typical hundred or thousand could also be one or two with larger window, in which case you double the memory usage for no reason.

And could you please show, which part of code is unnatural or not obvious now, compared to the previous version?

mtopolnik: While trying to validate your claims on memory savings, i ran this code:
```java
public static void main(String[] args) throws Exception {
    SlidingWindowPTest t = new SlidingWindowPTest();
    t.hasDeduct = true;
    t.isMutableFrame = true;
    t.before();
    SlidingWindowP<Object, ?, Long> proc = t.processor;
    for (long frameSeq = 0; frameSeq < 40; frameSeq++) {
        Punctuation punc = new Punctuation(frameSeq);
        while (!proc.tryProcessPunc0(punc)) {
        }
        for (long key = 0; key < 200_000; key++) {
            Frame<String, MutableLong> frame = new Frame<>(frameSeq, "" + key, MutableLong.valueOf(1));
            while (!proc.tryProcess0(frame)) {
            }
        }
    }
    System.out.println("seqToKeyToFrame " + proc.seqToKeyToFrame.size());
}

private static class MockOutbox implements Outbox {

    @Override
    public int bucketCount() {
        return 1;
    }

    @Override
    public boolean offer(int ordinal, @Nonnull Object item) {
        System.out.println("offer " + ordinal + ' ' + item);
        return true;
    }

    @Override
    public boolean offer(int[] ordinals, @Nonnull Object item) {
        return true;
    }
}

@Before
public void before() {
    WindowDefinition windowDef = new WindowDefinition(1, 0, 1);
    WindowOperation<Object, ?, Long> operation;

    if (isMutableFrame) {
        operation = WindowOperation.<Object, MutableLong, Long>of(
                MutableLong::new,
                (acc, val) -> {
                    throw new UnsupportedOperationException();
                },
                (acc1, acc2) -> {
                    acc1.value += acc2.value;
                    return acc1;
                },
                hasDeduct ? (acc1, acc2) -> {
                    acc1.value -= acc2.value;
                    return acc1;
                } : null,
                acc -> acc.value);
    } else {
        operation = WindowOperation.of(
                () -> 0L,
                (acc, val) -> {
                    throw new UnsupportedOperationException();
                },
                (acc1, acc2) -> acc1 + acc2,
                hasDeduct ? (acc1, acc2) -> (Long) acc1 - acc2 : null,
                identity());
    }

    processor = slidingWindow(windowDef, operation).get();
    processor.init(new MockOutbox(), mock(Context.class));
}
```

It turns out that there's a memory leak and no frames are evicted. I ran the same code on master and frames were being evicted.

viliam-durina: Finally a constructive comment, not just calling names. I'll have a look at it. 

But I'll still be glad, if you show, where the code is unnatural or not obvious, compared to the old version.

mtopolnik: The label "name-calling" you put on my words is an act of name-calling itself; this casts a shadow on your ostensibly righteous remark.

On master, the logic is this:

1. Compute the current window by deducting the trailing frame and adding the leading frame.
2. Emit the window.

In this PR the logic becomes:

1. Perform the second step in computing the current window.
2. Emit the window.
3. Perform the first step in computing the next window.

This fact alone would be enough to characterize the master code as "more natural", but the PR also introduces the more complex expression to calculate the frame seq to evict: `frameSeq - wDef.windowLength() + wDef.frameLength()`. Finally, to support this there is a method added to `Traverser`, one we already had and were happy to get rid of.

mtopolnik: Frames were being retained in my code due to the ordering between punctuation and frames. Punctuation N should come after all frames with seq N have been received.

mtopolnik: I find your claim of "doubling the memory usage" unjustified. The smallest number of frames per sliding window is 2. In this case, master code peaks at 3 frames per key, after this PR it's 2 frames. That's a 33% saving, but it assumes no disorder in events. With disorder, there will be more like 4-5 frames with master code, 3-4 frames in case of this PR. So the difference is now 25% or less, and we are still discussing the worst case which is also quite unrealistic (the window is much more likely to slide by a smaller fraction of its length).

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Traverser.java
```diff
@@ -125,6 +125,27 @@ public T next() {
...
 
     /**
+     * Returns a traverser that will emit the same items as this traverser and
+     * will additionally run the supplied action first time this traverser
+     * returns {@code null}.
+     */
+    @Nonnull
+    default Traverser<T> onFirstNull(@Nonnull Runnable action) {
+        return new Traverser<T>() {
+            private Runnable action2 = action;
```
mtopolnik: Using a `boolean actionRan` is better because the anonymous class already captures the `action` parameter as a final field; `action2` just duplicates it and nulling it out has no effect.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/SlidingWindowP.java
```diff
@@ -100,34 +102,38 @@ protected boolean tryProcessPunc0(@Nonnull Punctuation punc) {
...
-                                computeWindow(frameSeq, seqToKeyToFrame.remove(frameSeq - wDef.windowLength()))
-                                        .entrySet().stream()
-                                        .map(e -> new Frame<>(frameSeq, e.getKey(), finishF.apply(e.getValue()))))
-                        .flatMap(identity()));
-
-        traverser = traverser.append(punc);
-        return traverser;
+        long rangeStart = nextFrameSeqToEmit;
+        nextFrameSeqToEmit = wDef.higherFrameSeq(punc.seq());
+        return Traversers.traverseStream(range(rangeStart, nextFrameSeqToEmit, wDef.frameLength()).boxed())
```
mtopolnik: I see no value in leaving the `Traversers` qualifier here. It should be statically imported.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/SlidingWindowP.java
```diff
@@ -100,34 +102,38 @@ protected boolean tryProcessPunc0(@Nonnull Punctuation punc) {
...
+                .flatMap(frameSeq -> Traversers.traverseIterable(computeWindow(frameSeq).entrySet())
+                        .map(e -> (Object) new Frame<>(frameSeq, e.getKey(), finishF.apply(e.getValue())))
+                        .onFirstNull(() -> completeWindow(frameSeq)))
+                .append(punc);
     }
 
-    private Map<K, F> computeWindow(long frameSeq, Map<K, F> evictedFrame) {
+    private void completeWindow(long frameSeq) {
```
mtopolnik: It would be better if the method `completeWindow` came below `computeWindow` since that's their "program order"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/SlidingWindowP.java
```diff
@@ -100,34 +102,38 @@ protected boolean tryProcessPunc0(@Nonnull Punctuation punc) {
...
-            keyToFrame.forEach((key, currAcc) ->
-                    window.compute(key, (x, acc) -> combineF.apply(acc != null ? acc : createF.get(), currAcc)));
+    }
+
+    private Map<K, F> computeWindow(long frameSeq) {
+        if (deductF != null) {
+            // add leading-edge frame
+            patchSlidingWindow(combineF, seqToKeyToFrame.get(frameSeq));
+            return slidingWindow;
+        } else {
```
mtopolnik: This is a redundant else.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/SlidingWindowP.java
mtopolnik: I do realize that this same logic is used for tumbling windows as well, in which case the master code causes a peak retention of two windows' worth per key.

viliam-durina: Marko, do you believe, that this is so much more complex as to justify to afford 50% more memory in special, but real case and a bit more memory in all cases? This is just inconsistent with what you did in `WindowDefinition.higherFrameSeq()` for a case, that will _never_ happen. It could be just one-liner instead of 4 lines, and you argued "it's good to be correct even in edge cases". Why not this time?

Btw, we should special-case tumbling windows: we'll deduct all keys to zero or unnecessarily copy the frame, if there is no `deductF`. But will this be too much complexity? I don't dare to answer.

And we could also avoid the `onFirstNull` if you don't like it, in `tryProcessPunc()` create a traverser just for single frameSeq. We'll then have a chance to clean up after `emitFromTraverser` returns `true`. This is how I originally written it, before you converted it to use `flatMap`. `Stream` or `Traverser` doesn't make everything less complex, just because it's hip.

And the punctuation N before or after frameN: this is another issue I intended to raise, after this is merged. It should be before, and after frameN, there should be punct(N+1) saying, that frameN is complete.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/SlidingWindowP.java
viliam-durina: It was intentional: at first glance you see, that there are two paths that can be taken, each with it's own `return`. However, I consider this a matter of style, which I don't consider that important. But I would add [`InnerAssignmentCheck`](http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/InnerAssignmentCheck.html) to checkstyle...

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/SlidingWindowP.java
mtopolnik: You appear to conflate correctness with efficiency; a pure function that gives an incorrect answer in some cases is not on the same footing as a piece of code what uses a low-constant-factor more memory in an edge case. You also appear to continue your name-calling, this time going as far as inventing my opinions out of thin air. Traversers aren't "hip", they are a practical means of eliminating complex boilerplate of cooperative emission.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/SlidingWindowP.java
mtopolnik: If, as you say, you can see at the first glance that each code path has its own `return`, that means the `else` is redundant.

viliam-durina:
The test will fail, until #394 is merged, I'll rebase then.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/WindowOperations.java
```diff
@@ -62,7 +62,7 @@ private WindowOperations() {
     /**
      * Returns an operation that counts the items in the window.
      */
-    public static <T> WindowOperation<T, ?, Long> summingToLong(@Nonnull Distributed.ToLongFunction<T> mapper) {
+    public static <T> WindowOperation<T, ?, Long> summingToLong(@Nonnull Distributed.ToLongFunction<? super T> mapper) {
```
mtopolnik: Was there an actual problem with just `T` here? I don't see how `T` could be inferred as anything else but the `mapper`'s type.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/WindowingProcessors.java
```diff
@@ -88,14 +89,11 @@ private WindowingProcessors() {
...
-     * @param <F> type of the frame
+     * @param <F> type of accumulator
      * @param <R> type of the result derived from a frame
-     * @param emitPunctuation If {@code true}, punctuation will be emitted when window
-     *                        is closed. Enable, if downstream vertex needs to know,
-     *                        that it has received all keys for a particular window.
      */
     public static <K, F, R> Distributed.Supplier<SlidingWindowP<K, F, R>> slidingWindow(
-            WindowDefinition windowDef, WindowOperation<K, F, R> windowOperation, boolean emitPunctuation) {
+            WindowDefinition windowDef, WindowOperation<?, F, R> windowOperation) {
```
mtopolnik: Seems like we have no basis to infer `K` here, it can resolve to anything. This is a part of the wider concern on what to declare as the return type. We could have just `Distributed.Supplier<Processor>` if we don't want to expose the implementation type, but then we lose all the type information; OTOH the type information is arbitrary and could be wrong.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/WindowOperations.java
viliam-durina: No. I just did it, because I always do it with mappers. Yes, in this case, T isn't used anywhere else.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/WindowingProcessors.java
viliam-durina: I'd prefer `Supplier<Processor>`, not to expose our internal API

viliam-durina:
verify

viliam-durina:
There was a VM crash: https://hazelcast-l337.ci.cloudbees.com/job/Jet-pr-builder/1428/

viliam-durina:
Another VM crash https://hazelcast-l337.ci.cloudbees.com/job/Jet-pr-builder/1429/

viliam-durina:
verify

hazelcast-jet-core/src/test/java/com/hazelcast/jet/windowing/SlidingWindowPTest.java
```diff
@@ -17,45 +17,97 @@
...
 public class SlidingWindowPTest extends StreamingTestSupport {
 
-    private SlidingWindowP<Object, Long, Long> processor;
+    @Parameter
+    public boolean hasDeduct;
+
+    @Parameter(1)
+    public  boolean isMutableFrame;
+
+    @Parameters
```
cangencer: parameter should have names

viliam-durina:
verify

PR #394, opened 2017-04-21
Add mutable accumulators and fix equals for WindowOps.reducing

viliam-durina:
SlidingWindowP, if it has deductF, compares current accumulated value with
empty accumulator, and if it's equal, then removes the corresponding key.
`Object[]` doesn't implement equals, so I replaced it with MutableObject

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Accumulators.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Accumulators.java
@@ -31,7 +31,10 @@ public final class Accumulators {
      */
     public static class MutableInteger {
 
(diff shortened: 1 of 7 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Accumulators.java
```diff
@@ -0,0 +1,260 @@
...
+        @Override
+        public String toString() {
+            return "MutableDouble(" + value + ')';
+        }
+    }
+
+    /**
+     * Mutable object container.
+     */
+    public static class MutableObject<T> {
```
mtopolnik: This should be called `MutableReference`.

PR #401, opened 2017-04-25
Remove unused parameter from SlidingWindowP

cangencer:


cangencer:
realized this is already fixed by #392

PR #406, opened 2017-04-28
Fix termination of blocking workers on shutdown

viliam-durina:
On shutdown, cooperative a and blocking workers are attempted to quit. For
blocking workers, we called `blockingTaskletExecutor.shutdown()`, which,
however, only prevents further tasks from being submitted. The existing tasks
are untouched, so the blocking workers remain running.

Solution is to cancel the jobFutures on shutdown.


mtopolnik:
Superseded by https://github.com/hazelcast/hazelcast-jet/pull/409.

PR #407, opened 2017-04-28
Refactor windowing processors into single class

viliam-durina:
Refactor windowing processors into single class

GroupByFrameP, SlidingWindowP and recently also TumblingWindowP were very 
similar. They had the same internal state, they processed either events or 
Frames on input, and their output Frames contained either accumulator or 
finished result. Here, they are refactored into single class.

Advantages of this:
- less code
- uses HashMap instead of TreeMap (a bit faster)
- as a bonus, it allowed us to do one stage sliding window aggregation, which 
might be needed, if there are too many keys to fit into single machine.


hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/WindowingProcessors.java
```diff
@@ -17,13 +17,58 @@
...
 /**
  * Contains factory methods for processors dealing with windowing operations.
+ *
+ * <h1>Two-step aggregation</h1>
+ *
+ * This way of processing aggregates events first on local member and
+ * only pre-accumulated results are sent over the network. This greatly
+ * reduces the amount of data sent over the network. The DAG is:
+ *
+ * <pre>{@code
```
mtopolnik: Since you don't use `<` in the diagram, you don't need `{@code}`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/WindowingProcessors.java
```diff
@@ -17,13 +17,58 @@
...
 import com.hazelcast.jet.stream.DistributedCollector;
 
 import javax.annotation.Nonnull;
 
+import static com.hazelcast.jet.Distributed.Function.identity;
+
 /**
  * Contains factory methods for processors dealing with windowing operations.
+ *
+ * <h1>Two-step aggregation</h1>
```
mtopolnik: We should use the term "stage" consistently.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/windowing/SlidingWindowIntegrationTest.java
```diff
@@ -44,24 +49,33 @@
...
 
-@Category(QuickTest.class)
-@RunWith(HazelcastParallelClassRunner.class)
+@RunWith(Parameterized.class)
+@Category({QuickTest.class, ParallelTest.class})
+@Parameterized.UseParametersRunnerFactory(HazelcastParametersRunnerFactory.class)
 public class SlidingWindowIntegrationTest extends JetTestSupport {
 
+    @Parameter
+    public boolean oneStageProcessor;
```
mtopolnik: Should be renamed to "singleStageProcessor"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/WindowingProcessors.java
```diff
@@ -47,59 +93,155 @@ private WindowingProcessors() {
...
-            WindowDefinition windowDef,
-            WindowOperation<? super T, F, ?> collector
+    @Nonnull
+    public static <T, A> Distributed.Supplier<Processor> slidingWindowStage1(
+            @Nonnull Distributed.ToLongFunction<? super T> extractEventSeqF,
+            @Nonnull WindowDefinition windowDef,
+            @Nonnull WindowOperation<? super T, A, ?> windowOperation
     ) {
-        return groupByFrame(t -> "global", extractEventSeqF, windowDef, collector);
+        return slidingWindowStage1(t -> "global", extractEventSeqF, windowDef, windowOperation);
```
cangencer: I would prefer `ALL` instead of "global". Also it should be a static variable.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/WindowingProcessors.java
```diff
@@ -47,59 +93,155 @@ private WindowingProcessors() {
...
+        // use a single-frame window in this stage; the downstream processor
+        // combines the frames into a window with the user-requested size
+        WindowDefinition tumblingWinDef = new WindowDefinition(
+                windowDef.frameLength(), windowDef.frameOffset(), 1);
+
+        return () -> new WindowingProcessor<T, A, A>(
+                tumblingWinDef,
+                item -> tumblingWinDef.higherFrameSeq(extractEventSeqF.applyAsLong(item)),
+                extractKeyF,
+                WindowOperation.of(
```
mtopolnik: Another thing we can have is a windowOperation that takes a `BiConsumer accumulateItemF` and wraps it into a `BiFunction`. There are already many places where this is done by hand.

viliam-durina: I wouldn't bother with an overload, they are pain to maintain and writing `a -> a` is trivial. Normally, mutable accumulator is the way to go, due to much less GC load. I would even consider using `BiConsumer` for all `accumulateItem`, `combineAccumulators` and `deductAccumulators`, this will remove a lot of code. And if one wants an immutable container, he will use `WindowingOperations.reducing`, which even doesn't have a finisher. In next PR, if you wish.

viliam-durina:
I addressed your comments. Additionally, I renamed tests. I removed the `TumblingWindowPTest`, which was essentially a copy of `GroupByFramePTest`, but with finished accumulator value.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/windowing/WindowingProcessors.java
mtopolnik: It's not `a -> a`; it's `(acc, item) -> { biConsumer.accept(acc, item); return biConsumer; }` instead of simply `biConsumer`. You may be right on the point that the separation of the two cases (mutable/immutable) could go even deeper.

PR #412, opened 2017-04-30
Linear regression window operation

mtopolnik:
Also improve mutable `Accumulators`.


hazelcast-jet-core/src/main/java/com/hazelcast/jet/accumulator/LinRegAccumulator.java
```diff
@@ -0,0 +1,150 @@
...
+import java.io.IOException;
+import java.math.BigInteger;
+
+/**
+ * Maintains the components needed to compute the linear regression on a
+ * set of {@code (long, long)} pairs. The intermediate results are held in
+ * the form of {@link BigInteger} and the finished value is a {@code
+ * double}-valued linear coefficient.
+ */
+public final class LinRegAccumulator {
```
viliam-durina: There are two components to linear regression: slope and y-position: `y = ax + b`, `a` being the slope. We should call this `LinearRegressionSlope` or calculate both.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/accumulator/LinRegAccumulator.java
```diff
@@ -0,0 +1,150 @@
...
+        sumXY = sumXY.subtract(that.sumXY);
+        sumX2 = sumX2.subtract(that.sumX2);
+        return this;
+    }
+
+    /**
+     * Computes the linear coefficient of the linear regression of the
+     * accumulated samples.
+     */
+    public double finish() {
```
viliam-durina: We will have division by zero, if N<=1.
If all pairs had the same `x`-value (this is when the denominator is zero), the line is vertical and has infinite slope.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/accumulator/LinRegAccumulator.java
```diff
@@ -0,0 +1,150 @@
...
+ * Maintains the components needed to compute the linear regression on a
+ * set of {@code (long, long)} pairs. The intermediate results are held in
+ * the form of {@link BigInteger} and the finished value is a {@code
+ * double}-valued linear coefficient.
+ */
+public final class LinRegAccumulator {
+    private static final int MAX_BIGINT_LEN = 255;
+
+    private long n;
+    private BigInteger sumX;
```
viliam-durina: We could use int128, it's much faster than BigInteger. Normally, X will be time. We add `X^2`, n-times. If `x` is `long`, then we need int192. I experimented with this one: https://github.com/apache/hive/blob/master/common/src/java/org/apache/hadoop/hive/common/type/SignedInt128.java, when I was doing my original streaming sample, it was considerably faster.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/accumulator/LinRegAccumulator.java
viliam-durina: Actually,  this is correct. We'll never have n=0, as we don't emit windows, when there were no events.
With n=1, we'll have `0 / 0`, which is `NaN`, which is correct. With `n>1` and all x-values the same, we'll have `a / 0`, which is `Inf` or `-Inf`, which is also correct to emit. Anyway, I'll add this to the test.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/accumulator/LinRegAccumulator.java
mtopolnik: I'm fine with the current name... if we ever need to differentiate, we'll rename to more specific names.

cangencer:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/accumulator/LinRegAccumulator.java
mtopolnik: let's defer this level of optimization for some other time

hazelcast-jet-core/src/main/java/com/hazelcast/jet/accumulator/LinRegAccumulator.java
```diff
@@ -0,0 +1,150 @@
...
+    private BigInteger sumY;
```
mtopolnik: We shouldn't since the sum is always an integer.

PR #418, opened 2017-05-06
Improve tests, fix bug in linearTrend

mtopolnik:
`n++` was missing in `LinTrendAccumulator#accumulate`, but tests were passing because the term that involves `n` is apparently not needed when the y-intercept of linear regression is zero.

The test now uses data with a non-zero y-intercept.


mtopolnik:
VM crash: https://hazelcast-l337.ci.cloudbees.com/job/Jet-pr-builder/1497/

mtopolnik:
verify

PR #426, opened 2017-05-09
Add support for debugging

### LINKED ISSUE #414: Add support for debugging
Add general-purpose peek processor. It will use `System.out.println`. Ideas
- print output or input
- sampling (output every n-th item)
- implement it by wrapping `ProcessorSupplier`
- have a formatter

viliam-durina:
Resolves #414

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
@@ -17,13 +17,12 @@
 package com.hazelcast.jet;
 
 import com.hazelcast.client.config.ClientConfig;
-import com.hazelcast.jet.function.DistributedConsumer;
+import com.hazelcast.jet.Traversers.ResettableSingletonTraverser;
 import com.hazelcast.jet.function.DistributedBiConsumer;
 import com.hazelcast.jet.function.DistributedBiFunction;
+import com.hazelcast.jet.function.DistributedConsumer;
 import com.hazelcast.jet.function.DistributedFunction;
 import com.hazelcast.jet.function.DistributedIntFunction;
-import com.hazelcast.jet.Distributed.Predicate;
-import com.hazelcast.jet.Traversers.ResettableSingletonTraverser;
 import com.hazelcast.jet.function.DistributedPredicate;
 import com.hazelcast.jet.function.DistributedSupplier;
 import com.hazelcast.jet.impl.connector.HazelcastWriters;
@@ -56,11 +55,10 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
-import static com.hazelcast.jet.DistributedFunctions.alwaysTrue;
-import static com.hazelcast.jet.DistributedFunctions.noopConsumer;
-import static com.hazelcast.jet.function.DistributedFunctions.noopConsumer;
 import static com.hazelcast.jet.Traversers.lazy;
 import static com.hazelcast.jet.Traversers.traverseStream;
+import static com.hazelcast.jet.function.DistributedFunctions.alwaysTrue;
+import static com.hazelcast.jet.function.DistributedFunctions.noopConsumer;
 import static com.hazelcast.jet.impl.util.Util.uncheckCall;
 import static com.hazelcast.jet.impl.util.Util.uncheckRun;
 
@@ -390,7 +388,7 @@ public final class Processors {
     }
 
     /**
-     * Convenience for {@link #writeFile(String, Distributed.Function, Charset,
+     * Convenience for {@link #writeFile(String, DistributedFunction, Charset,
      * boolean)} with UTF-8 charset and with overwriting of existing files.
      *
      * @param directoryName directory to create the files in. Will be created,
@@ -402,7 +400,7 @@ public final class Processors {
     }
 
     /**
-     * Convenience for {@link #writeFile(String, Distributed.Function, Charset,
+     * Convenience for {@link #writeFile(String, DistributedFunction, Charset,
      * boolean)} with UTF-8 charset and with overwriting of existing files.
      *
      * @param directoryName directory to create the files in. Will be created,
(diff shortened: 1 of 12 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
```diff
@@ -803,6 +810,322 @@ public void process(int ordinal, @Nonnull Inbox inbox) {
...
+     * INFO level.
+     * <p>
+     * Note, that the event will be logged on the nodes, not on the client.
+     * Useful for testing.
+     *
+     * @param toStringF Function to convert item to String, if {@code null},
+     *                  {@link Object#toString()} is used.
+     */
+    public static Distributed.Supplier<Processor> writeSystemOut(
+            @Nullable Distributed.Function<Object, String> toStringF
```
cangencer: should be @NotNull instead of @Nullable

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/WrappingProcessorMetaSupplier.java
```diff
@@ -0,0 +1,54 @@
...
+import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.function.Function;
+
+/**
+ * A {@link ProcessorMetaSupplier}, which wraps another {@code
+ * ProcessorMetaSupplier} with one, that will wrap its processors using
+ * {@code wrapperSupplier}.
+ */
+public final class WrappingProcessorMetaSupplier implements ProcessorMetaSupplier {
```
gurbuzali: can we move these supplier classes inside processor class?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
```diff
@@ -823,6 +828,154 @@ public void process(int ordinal, @Nonnull Inbox inbox) {
...
+    public static Distributed.Supplier<Processor> peekInput(Distributed.Supplier<Processor> wrapped) {
+        return peekInput(null, null, wrapped);
+    }
+
+    /**
+     * See {@link #peekInput(Distributed.Function, Predicate, ProcessorMetaSupplier)}
+     */
+    public static Distributed.Supplier<Processor> peekInput(
+            Distributed.Function<Object, String> toStringF, Predicate<Object> shouldLogF,
+            Distributed.Supplier<Processor> wrapped) {
```
mtopolnik: Here and elsewhere, instead of accepting `null` we should annotate with `@Nonnull` and have our library provide convenient default function factories, like `DistributedFunctions.alwaysTrue`. Where applicable we can add more overloads.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/WrappingProcessorMetaSupplier.java
mtopolnik: This is a wrapper which doesn't belong to any processor class.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/WrappingProcessorSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+        this.wrapped = wrapped;
+        this.wrapperSupplier = wrapperSupplier;
+    }
+
+    @Nonnull
+    @Override
+    public Collection<? extends Processor> get(int count) {
+        Collection<? extends Processor> processors = wrapped.get(count);
+        return processors.stream()
+                .map(wrapperSupplier)
```
mtopolnik: This should be indented so it's aligned on the dot with the previous line. There is an IDEA formatter setting for this: `Wrapping and Braces -> Chined method calls -> Align when multiline`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/WrappingProcessorMetaSupplier.java
```diff
@@ -0,0 +1,54 @@
...
+ * A {@link ProcessorMetaSupplier} which wraps another {@code
+ * ProcessorMetaSupplier} with one that will wrap its processors using
+ * {@code wrapperSupplier}.
+ */
+public final class WrappingProcessorMetaSupplier implements ProcessorMetaSupplier {
+    private ProcessorMetaSupplier wrapped;
+    private Function<Processor, Processor> wrapperSupplier;
+
+    public WrappingProcessorMetaSupplier(ProcessorMetaSupplier wrapped, Function<Processor,
+            Processor> wrapperSupplier) {
```
mtopolnik: We shouldn't have a line break between two type parameters. The whole second argument declaration should be on the next line.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/Processors_peekTest.java
```diff
@@ -0,0 +1,184 @@
...
+        logger = mock(ILogger.class);
+        context = new ProcCtx(null, logger, null, 0);
+    }
+
+    @Test
+    public void when_peekInput_SupplierProcessor() {
+        // Given
+        DistributedSupplier<Processor> passThroughPSupplier = Processors.filter(alwaysTrue());
+        Processor wrappedP =
+                (toStringF == null ? peekInput(passThroughPSupplier) : peekInput(toStringF, shouldLogF, passThroughPSupplier))
```
mtopolnik: Line length exceeds 120 here and in several more places below. This rule should be present in CheckStyle, though. I'll add it.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/WrappingProcessorMetaSupplier.java
viliam-durina: that was a mistake

PR #430, opened 2017-05-11
Add glob to StreamFilesP

viliam-durina:
This makes StreamFilesP in-line with ReadFilesP which has that option.

It also allows for use case where users create `.tmp` files and then rename them to real file name and only then want them to be processed, as mentioned in the PRD.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
@@ -308,12 +308,7 @@ public final class Processors {
     }
 
     /**
-     * Returns a supplier of processors, that read all files in a directory and emit them
-     * line by line. Files should be in UTF-8 encoding.
-     *
-     * @param directory Parent directory of the files.
-     *
-     * @see #readFiles(String, Charset, String)
+     * See {@link #readFiles(String, Charset, String)}.
      */
     @Nonnull
     public static ProcessorSupplier readFiles(@Nonnull String directory) {
@@ -344,10 +339,10 @@ public final class Processors {
     }
 
     /**
-     * @see #streamFiles(String, Charset)
+     * See {@link #streamFiles(String, Charset, String)}
      */
     public static ProcessorSupplier streamFiles(@Nonnull String watchedDirectory) {
-        return streamFiles(watchedDirectory, null);
+        return streamFiles(watchedDirectory, null, null);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
```diff
@@ -339,10 +334,10 @@ public static ProcessorSupplier readFiles(
     }
 
     /**
-     * @see #streamFiles(String, Charset)
+     * See {@link #streamFiles(String, Charset, String)}
```
mtopolnik: Javadoc here and above is missing information on the default values for missing arguments.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
```diff
@@ -375,11 +377,13 @@ public static ProcessorSupplier streamFiles(@Nonnull String watchedDirectory) {
...
      * @param charset charset to use to decode the files, or {@code null} to use UTF-8
+     * @param glob the file filter, see {@link
+     *             java.nio.file.FileSystem#getPathMatcher(String) getPathMatcher()}
      */
     public static ProcessorSupplier streamFiles(
-            @Nonnull String watchedDirectory, @Nullable Charset charset
+            @Nonnull String watchedDirectory, @Nullable Charset charset, @Nullable String glob
     ) {
-        return StreamFilesP.supplier(watchedDirectory, charset == null ? null : charset.toString());
+        return StreamFilesP.supplier(watchedDirectory, charset == null ? null : charset.toString(), glob);
```
mtopolnik: Here we have another place where we accept `null` and substitute defaults. Maybe not in this PR, but we should change this.

viliam-durina: For me, "null" meaning "use default" is perfectly acceptable. Especially here, when you might want to use just one of the parameters (either glob or charset) and want the other to be the default. You just don't care. I didn't want to add another 2 overloads... Yesterday, it also made the test more complicated because I cannot say "use default" using parameters; I had to use different overload to test the default.

mtopolnik: We need a consistent story for the whole API. We either do or don't substitute defaults for `nulls`.  Calling with `StandardCharsets.UTF_8` instead of `null` is not much more difficult and makes the choice clear. I'm especially concerned wit calls like `(path, null, null)`.

mtopolnik: Providing a constant instead of `null` is practically always better: `(path, DEFAULT_CHARSET, DEFAULT_GLOB)`.

viliam-durina:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
viliam-durina: pushed

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadFilesP.java
```diff
@@ -57,9 +56,9 @@
 
     @Override
     public boolean complete() {
-        try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(directory, glob == null ? "*" : glob)) {
+        try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(directory, glob)) {
             StreamSupport.stream(directoryStream.spliterator(), false)
-                    .filter(this::shouldProcessEvent)
+                    .filter(this::shouldProcessEventAndNotDir)
```
mtopolnik: I don't understand this name. After studying its implementation for a while, looks like `filesAssignedToMe` or something similar.

viliam-durina:
verify

PR #433, opened 2017-05-12
Move specific newNullTraverser out from public API

viliam-durina:
The `newNullTraverser()` was required to always create an unique
instance. This requirement is specific to `FlatMappingTraverser`,
however, a `nullTraverser() ` method returning a singleton is useful in
public API.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Traversers.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Traversers.java
@@ -37,7 +37,7 @@ public final class Traversers {
      * Returns a traverser that always returns null.
      */
     @Nonnull
-    public static <T> Traverser<T> nullTraverser() {
+    public static <T> Traverser<T> empty() {
         return () -> null;
     }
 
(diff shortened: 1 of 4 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Traversers.java
```diff
@@ -34,14 +34,11 @@ private Traversers() {
...
-     * Returns an empty traverser equivalent to {@code () -> null}, but with
-     * guaranteed unique identity. Such a traverser is useful in a <em>null
-     * object</em> pattern, which relies on identity checks with {@code ==}
-     * against a private object.
+     * Returns a traverser that always returns null.
      */
     @Nonnull
-    public static <T> Traverser<T> newNullTraverser() {
-        return new NullTraverser<>();
+    public static <T> Traverser<T> nullTraverser() {
```
mtopolnik: This should be called `empty()`, to be called as `Traversers.empty()`. The name `newNullTraverser()` was used in reference to the "NULL object" pattern, which this doesn't serve.

viliam-durina:
verify

PR #434, opened 2017-05-16
Add test for CountDistinctP

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
```diff
@@ -801,11 +802,18 @@ public static ProcessorMetaSupplier writeFile(@Nonnull String directoryName) {
     }
 
     /**
-     * Returns a supplier of {@link CountDistinctP} processors.
+     * Returns a supplier of processors that will calculate the number of
+     * distinct keys they have seen on input. To work properly the vertex has
+     * to be preceded by an {@link Edge#allToOne() all-to-one} and a {@link
+     * Edge#distributed() distributed} edge. It will emit single {@code
```
mtopolnik: I don't understand this sentence. Does it need two edges, a single edge with both properties, or any of the above? Does "preceded" mean any edge preceding it in topological order?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
```diff
@@ -1128,30 +1136,10 @@ protected boolean tryProcess(int ordinal, @Nonnull Object item) throws Exception
...
-     * </li><li>
-     *     Maintains a set of all seen keys.
-     * </li><li>
-     *     Emits the size of the set (the number of seen distinct keys) as a
-     *     {@code Long} value.
-     * </li></ul>
-     *
-     * @param <T> type of received item
-     * @param <K> type of grouping key
-     */
```
mtopolnik: This Javadoc should not be removed, it documents clearly what the processor does. Instead it should be moved to the factory method and adapted. See `groupAndAccumulate` for an example.

viliam-durina: I just considered it to be too detailed for such simple processor.

PR #442, opened 2017-05-22
[WIP] Converted StreamKafkaP to cooperative

gurbuzali:
Added pause/resume logic for KafkaConsumer session timeout

### DIFF
--- a/hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/connector/kafka/StreamKafkaP.java
+++ b/hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/connector/kafka/StreamKafkaP.java
@@ -17,6 +17,7 @@
 package com.hazelcast.jet.connector.kafka;
 
 import com.hazelcast.jet.AbstractProcessor;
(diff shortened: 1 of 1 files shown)


hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/connector/kafka/StreamKafkaP.java
```diff
@@ -71,28 +75,58 @@ private StreamKafkaP(String[] topicIds, Properties properties) {
...
     }
 
-    @Override
-    public boolean isCooperative() {
-        return false;
+    private void resume() {
+        consumer.resume(consumer.assignment());
+    }
+
+    private void emptyPoll() {
```
mtopolnik: I find this method nome confusing, it makes the same `poll` call as the "nonempty" poll. `pollWhilePaused` would be more self-explanatory.

Also, it uses `poll(0)` while the main path uses `poll(ZERO_POLL_TIMEOUT)`.

gurbuzali:
verify

PR #444, opened 2017-05-23
Add static method for NoopP processor

viliam-durina:
Make it same style as other processors

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
@@ -843,6 +843,15 @@ public final class Processors {
     }
 
     /**
+     * Returns a supplier of processor that do nothing. They consume all input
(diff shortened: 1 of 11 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Processors.java
```diff
@@ -843,6 +843,15 @@ public static ProcessorMetaSupplier writeFile(@Nonnull String directoryName) {
     }
 
     /**
+     * Returns a supplier of processor that do nothing. They consume all input
```
mtopolnik: Must be rewritten to singular form. ("does nothing", "It consumes", etc.).

viliam-durina: Pushed a fix. Hope I won't cause too many conflicts with your changes. I only changed javadoc of sources/sinks, that you are not going to change.

PR #449, opened 2017-05-26
Reorganize processor factory utility classes

mtopolnik:
* Sources
* Sinks
* Processors
* DiagnosticProcessors


hazelcast-jet-core/src/main/java/com/hazelcast/jet/processor/package-info.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Contains static utility classes with factories of Jet processors. The
+ * factories are organized into several categories:
+ * <ul><li>
+ *     {@link com.hazelcast.jet.processor.Sources}
+ * </li><li>
+ *     {@link com.hazelcast.jet.processor.Sinks}
+ * </li><li>
+ *     {@link com.hazelcast.jet.processor.Processors Internal DAG vertices}
```
mtopolnik: https://en.wikipedia.org/wiki/Directed_graph

>If a vertex is neither a source nor a sink, it is called an _internal_.

PR #452, opened 2017-05-30
Refactor Kafka & Hadoop processor factories

viliam-durina:
- create factory classes in `com.hazelcast.jet.processor`
- move implementation classes under `com.hazelcast.jet.impl` package
- move javadoc to factory class

Update of samples and refman will follow.


viliam-durina:
verify

PR #453, opened 2017-05-30
Remove TODO regarding SkewReductionPolicy

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/SkewReductionPolicy.java
```diff
@@ -84,6 +84,10 @@
     private final long priorityDrainingThreshold;
     private final boolean forceAdvancePunc;
 
+    public SkewReductionPolicy(int numQueues) {
```
viliam-durina: We should add comment here, something like: "Creates skew reduction policy that does nothing (is disabled)".

cangencer:
verify

PR #457, opened 2017-06-02
Support window aggregation in batch

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Traverser.java
```diff
@@ -172,4 +172,30 @@ public T next() {
     static <T> Traverser<T> over(T... items) {
         return Traversers.traverseArray(items);
     }
+
+    /**
+     * Returns a traverser traversing over the longest prefix of elements taken
```
mtopolnik: 1. Just "traverser over" would be enough.
2. There is no clear treatment of `null` returned from the wrapped traverser. `null` shouldn't be passed into the predicate, we should treat it as if the predicate returned `false`. We must also document that `takeWhile` handles only`null`-terminated sequences.

viliam-durina: Thanks, I completely overlooked it. I also added test for it.

mtopolnik:
verify

mtopolnik:
verify

mtopolnik:
verify

PR #461, opened 2017-06-06
Add test support utilities

viliam-durina:


hazelcast-jet-test-support/src/main/java/com/hazelcast/jet/testsupport/TestSupport.java
```diff
@@ -0,0 +1,278 @@
...
+            @Nonnull ProcessorMetaSupplier supplier,
+            @Nonnull List<T> input,
+            @Nonnull List<U> expectedOutput,
+            boolean assertProgress
+    ) {
+        testProcessor(processorFrom(supplier), input, expectedOutput, assertProgress);
+    }
+
+    /**
+     * An utility to test processors. It will initialize the processor
```
mtopolnik: `An utility` -> `A utility`

hazelcast-jet-test-support/src/main/java/com/hazelcast/jet/testsupport/TestSupport.java
```diff
@@ -0,0 +1,278 @@
...
+     * instance, pass input items to it and assert the outbox contents.
+     * <p>
+     * For cooperative processors a 1-capacity outbox will be provided, which
+     * will additionally be full on every other processing method call. This
```
mtopolnik: `additionally` is redundant. The effect of being full on every other call is the result of capacity 1, not an additional effect.

hazelcast-jet-test-support/src/main/java/com/hazelcast/jet/testsupport/TestSupport.java
```diff
@@ -0,0 +1,278 @@
...
+     * will test edge cases in cooperative processors.
+     * <p>
+     * This method does the following:<ul>
+     *     <li>initializes the processor by calling {@link Processor#init(
+     *     com.hazelcast.jet.Outbox, com.hazelcast.jet.Processor.Context)}
+     *
+     *     <li>calls {@link Processor#process(int, Inbox)} with inbox
```
mtopolnik: Directly saying `calls process(0, inbox)` instead of referring to the method and then explaining arguments passed is more natural.

hazelcast-jet-test-support/src/main/java/com/hazelcast/jet/testsupport/TestSupport.java
```diff
@@ -0,0 +1,278 @@
...
+     *
+     *     <li>asserts the progress of the {@code process()} call: that
+     *     something was taken from the inbox or put to the outbox
+     *
+     *     <li>calls {@link Processor#complete()} until it returns {@code true}
+     *
+     *     <li>asserts the progress of the {@code complete()} call if it
+     *     returned {@code false}: something must have been put to the outbox.
+     * </ul>
+     * Note that this method never calls {@link Processor#tryProcess()}.
```
mtopolnik: I would expand this to explain why it's important. `Note that this method only uses the top-level Processor API and doesn't directly call AbstractProcessor#tryProcess()`.

viliam-durina: This doesn't refer to the `AbstractProcessor.tryProcess(ordinal, item)`, but to the one without inbox in `Processor`.

hazelcast-jet-test-support/src/main/java/com/hazelcast/jet/testsupport/TestSupport.java
viliam-durina: It's an additional effect: I put `DUMMY_OBJ` to outbox to make it full, before every other process call.

mtopolnik: Then it's unclear; it managed to confuse me. It should document this peculiar behavior of inserting a dummy item to make the outbox full.

cangencer:
verify

mtopolnik:
verify

PR #463, opened 2017-06-07
Fail fast when adding non-serializable processor supplier

viliam-durina:
ProcessorMetaSupplier must be serializable. Out-of-the-box metasuppliers can become non-serializable if they are passed a non-serializable lambda. Currently the job fails if run on multiple instances and the problem is harder to spot, because it fails when serializing the ExecutionPlan, which might contain lot of lambda expressions.

This PR makes this problem easier to spot and will also make the job fail even when using single Jet member.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
@@ -33,6 +33,11 @@ import javax.annotation.Nonnull;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InvalidClassException;
+import java.io.NotSerializableException;
+import java.io.ObjectOutputStream;
+import java.io.OutputStream;
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Callable;
@@ -157,5 +162,37 @@ public final class Util {
     // opposite signs and result has opposite sign of left-hand operand
     public static boolean diffHadOverflow(long a, long b, long diff) {
         return ((a ^ b) & (a ^ diff)) < 0;
+    }
+
+    /**
+     * Checks, that {@code argument} implements {@link Serializable}.
+     * It also checks, if the {@code argument} is actually serializable by trying to serialize it.
+     * This will reveal early, if all it's fields are serializable.
+     *
+     * @param argument Object to check
+     * @param argumentName Argument name for the exception
+     * @throws IllegalArgumentException If {@code argument} is not serializable.
+     */
+    public static void checkSerializable(Object argument, String argumentName) {
+        if (argument != null) {
+            if (!(argument instanceof Serializable)) {
+                throw new IllegalArgumentException("\"" + argumentName + "\" must be serializable");
+            }
+            try  (ObjectOutputStream os  = new ObjectOutputStream(new NullOutputStream())) {
+                os.writeObject(argument);
+            } catch (NotSerializableException | InvalidClassException e) {
+                throw new IllegalArgumentException("\"" + argumentName + "\" must be serializable", e);
+            } catch (IOException e) {
+                // never really thrown, as the underlying stream never throws it
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    private static class NullOutputStream extends OutputStream {
+        @Override
+        public void write(int b) throws IOException {
+            // do nothing
+        }
     }
 }
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
@@ -165,23 +165,23 @@ public final class Util {
     }
 
     /**
-     * Checks, that {@code argument} implements {@link Serializable}.
-     * It also checks, if the {@code argument} is actually serializable by trying to serialize it.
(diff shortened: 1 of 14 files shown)


cangencer:
It should say `ProcessorMetaSupplier` should be serializable instead, the `Processor` itself is never serialized. `ProcessorSupplier` created from `ProcessorMetaSupplier` also needs to be serializable. This could be checked when creating execution plan, in `ExecutionPlanBuilder`.

viliam-durina:
verify

cangencer:
verify

viliam-durina:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -158,4 +163,36 @@ public static boolean sumHadOverflow(long a, long b, long sum) {
     public static boolean diffHadOverflow(long a, long b, long diff) {
         return ((a ^ b) & (a ^ diff)) < 0;
     }
+
+    /**
+     * Checks, that {@code argument} implements {@link Serializable}.
```
mtopolnik: Overuse of punctuation :) It should also be line-wrapped at 72 chars.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/WriteBufferedPTest.java
```diff
@@ -120,14 +108,46 @@ public boolean complete() {
...
-                (buffer, item) -> events.add("add:" + item),
-                buffer -> events.add("flush"),
-                buffer -> events.add("dispose")
+                (buffer, item) -> localEvents.add("add:" + item),
+                buffer -> localEvents.add("flush"),
+                buffer -> localEvents.add("dispose")
         );
     }
+
+    private static class MyAbstractProcessor extends AbstractProcessor {
```
mtopolnik: Is this an unused class? If it's used, it should be renamed.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
viliam-durina: The comment was written before you mentored me about commas :). Fixed

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/WriteBufferedPTest.java
viliam-durina: An oversight, fixed.

mtopolnik:
We should also add fail-fast behavior to `Edge` partitioning lambdas.

viliam-durina:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -158,4 +163,36 @@ public static boolean sumHadOverflow(long a, long b, long sum) {
...
+     * @param object object to check
+     * @param objectName object description for the exception
+     * @throws IllegalArgumentException if {@code object} is not serializable
+     */
+    public static void checkSerializable(Object object, String objectName) {
+        if (object != null) {
+            if (!(object instanceof Serializable)) {
+                throw new IllegalArgumentException("\"" + objectName + "\" must be serializable");
+            }
+            try  (ObjectOutputStream os  = new ObjectOutputStream(new NullOutputStream())) {
```
mtopolnik: Two extra spaces on this line.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -158,4 +163,36 @@ public static boolean sumHadOverflow(long a, long b, long sum) {
...
+            } catch (IOException e) {
+                // never really thrown, as the underlying stream never throws it
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    private static class NullOutputStream extends OutputStream {
+        @Override
+        public void write(int b) throws IOException {
```
mtopolnik: `throws IOException` is redundant.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -158,4 +163,36 @@ public static boolean sumHadOverflow(long a, long b, long sum) {
...
+        if (object != null) {
+            if (!(object instanceof Serializable)) {
+                throw new IllegalArgumentException("\"" + objectName + "\" must be serializable");
+            }
+            try  (ObjectOutputStream os  = new ObjectOutputStream(new NullOutputStream())) {
+                os.writeObject(object);
+            } catch (NotSerializableException | InvalidClassException e) {
+                throw new IllegalArgumentException("\"" + objectName + "\" must be serializable", e);
+            } catch (IOException e) {
+                // never really thrown, as the underlying stream never throws it
```
mtopolnik: This comment is incorrect because `ObjectOutputStream` may throw `IOException` itself. Given this fact, it would be better to wrap in `JetException` instead of the generic `RuntimeException`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
viliam-durina: Sorry, I overlooked your review. I'll address these in some future PR.

PR #465, opened 2017-06-08
Fix WriteSocketTest.unitTest()

viliam-durina:


viliam-durina:
The `unitTest()` was run 500 times on PR builder successfully. It seems that the problem was that sending side of socket was able to do "connect+send+close" before the server side accepted the connection...

PR #466, opened 2017-06-09
Fix class unloading

viliam-durina:
There were 3 references to JetClassLoader kept:

- from `JetService` by not cleaning the `JetClassLoader` and
`resourceStores` map

- from `ClassLoaderWeakCache`, which keeps reference to class loader as
a key

- from MetricsRegistry, because it keeps references to Class objects

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/ResourceStore.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/ResourceStore.java
@@ -14,9 +14,11 @@
  * limitations under the License.
  */
 
-
 package com.hazelcast.jet.impl.deployment;
 
+import com.hazelcast.jet.JetException;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.logging.Logger;
 import com.hazelcast.nio.IOUtil;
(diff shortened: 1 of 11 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/ResourceStore.java
```diff
@@ -34,25 +36,37 @@
...
+    private Path storageDirectory;
 
     private final Map<ResourceDescriptor, File> resources = new ConcurrentHashMap<>();
     private final Map<String, ClassLoaderEntry> jarEntries = new ConcurrentHashMap<>();
     private final Map<String, ClassLoaderEntry> dataEntries = new ConcurrentHashMap<>();
     private final Map<String, ClassLoaderEntry> classEntries = new ConcurrentHashMap<>();
 
-    public ResourceStore(String storagePath) {
-        this.storageDirectory = createStorageDirectory(storagePath);
+    public ResourceStore(Path rootDir, NodeEngineImpl nodeEngine) {
```
cangencer: preferable to use `NodeEngine` here.

cangencer:
verify

PR #470, opened 2017-06-12
Watermark emission policy

mtopolnik:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/WatermarkEmissionPolicy.java
```diff
@@ -0,0 +1,44 @@
...
+package com.hazelcast.jet;
+
+import javax.annotation.Nonnull;
+import java.io.Serializable;
+
+/**
+ * A policy object that decides when when the watermark has advanced
+ * enough to emit a new watermark item.
+ */
+public interface WatermarkEmissionPolicy extends Serializable {
```
cangencer: can be `@FunctionalInterface`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/WatermarkEmissionPolicy.java
```diff
@@ -0,0 +1,44 @@
...
+     * lastEmittedWm}.
+     */
+    boolean shouldEmit(long currentWm, long lastEmittedWm);
+
+    /**
+     * Returns a policy that ensures that each emitted watermark has a higher
+     * value than the last one.
+     */
+    @Nonnull
+    static WatermarkEmissionPolicy suppressDuplicates() {
```
cangencer: is there a reason this is here, but not together with the rest of the policies? would it make sense to put all in a separate class?

mtopolnik: This is supposed to serve as the "null" policy that just protects the basic invariant of watermark emission. But we could put all three in here.

cangencer: I'd prefer them to all be in same place, preferably in this class ?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/InsertWatermarkP.java
```diff
@@ -39,34 +39,39 @@
 public class InsertWatermarkP<T> extends AbstractProcessor {
```
viliam-durina: This should also be renamed to `InsertWatermarksP` (plural)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/WatermarkEmissionPolicy.java
```diff
@@ -0,0 +1,76 @@
...
+     * value should be emitted, given the last emitted value {@code
+     * lastEmittedWm}.
+     */
+    boolean shouldEmit(long currentWm, long lastEmittedWm);
+
+    /**
+     * Returns a policy that ensures that each emitted watermark has a higher
+     * value than the last one.
+     */
+    @Nonnull
```
viliam-durina: We should say more clearly that this method is useful only in special cases: "Does minimum possible throttling causing lot of overhead from watermarks. Useful for edge cases and tests."

hazelcast-jet-core/src/main/java/com/hazelcast/jet/WatermarkEmissionPolicy.java
mtopolnik: This is already done.

cangencer:
verify

mtopolnik:
verify

PR #479, opened 2017-06-29
Upload job resources through IMap

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ResourceUploader.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.Map;
+import java.util.jar.JarEntry;
+import java.util.jar.JarInputStream;
+import java.util.zip.DeflaterOutputStream;
+
+import static com.hazelcast.jet.impl.util.ExceptionUtil.rethrow;
+
+public class ResourceUploader {
+
+    private static final int MAP_PUT_BATCH_SIZE_BYTES = 1 << 18; // 256kB
```
cangencer: Map.putAll already sends one message/operation per partition, so this value can be much higher or removed all together

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ResourceUploader.java
```diff
@@ -0,0 +1,113 @@
...
+                }
+                readStreamAndPutCompressedToMap(jis, jarEntry.getName());
+            }
+        }
+    }
+
+    private void readStreamAndPutCompressedToMap(InputStream in, String resourceId) throws IOException {
+        String key = "res:" + resourceId;
+        // ignore duplicates: the first resource in first jar takes precedence
+        if (targetMap.containsKey(key)) {
```
cangencer: we should probably avoid a remote RPC call here. We can keep track locally which ones have already been added.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetClientInstanceImpl.java
```diff
@@ -94,42 +84,14 @@ protected JobImpl(DAG dag, JobConfig config) {
         @Override
         public Future<Void> execute() {
             long executionId = getIdGenerator().newId();
-            deployResources(executionId);
+            new ResourceUploader(JetClientInstanceImpl.this, executionId).uploadMetadata(config);
```
cangencer: better to just pass the required map here

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ResourceUploader.java
```diff
@@ -0,0 +1,113 @@
...
+                if (jarEntry.isDirectory()) {
+                    continue;
+                }
+                readStreamAndPutCompressedToMap(jis, jarEntry.getName());
+            }
+        }
+    }
+
+    private void readStreamAndPutCompressedToMap(InputStream in, String resourceId) throws IOException {
+        String key = "res:" + resourceId;
```
cangencer: the prefix should be refactored to a util method, since it's duplicated in several places.

viliam-durina:
verify

viliam-durina:
verify

PR #480, opened 2017-07-03
Minor fixes to uploading resources directly to Map code

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetService.java
```diff
@@ -175,7 +175,7 @@ public ClientInvocationRegistry getClientInvocationRegistry() {
     }
 
     public ClassLoader getClassLoader(long executionId) {
-        IMap<String, Object> jobMetadataMap = getJetInstance().getMap(METADATA_MAP_PREFIX + executionId);
+        IMap<String, byte[]> jobMetadataMap = getJetInstance().getMap(METADATA_MAP_PREFIX + executionId);
```
viliam-durina: I intended the map to also contain all metadata for the job in the future, not just the resources, that's why resources are prefixed with the `res:` prefix and that's why I used `Object` for value. But anyway, this will work, it's just a compiler trick...

PR #487, opened 2017-07-18
Introduce stream event journal processor

gurbuzali:
depends on https://github.com/hazelcast/hazelcast/pull/11187

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
@@ -1,0 +1,359 @@
+/*
+ * Copyright (c) 2008-2017, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.jet.impl.connector;
+
+import com.hazelcast.cache.journal.EventJournalCacheEvent;
+import com.hazelcast.client.config.ClientConfig;
+import com.hazelcast.client.impl.HazelcastClientProxy;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.ICompletableFuture;
+import com.hazelcast.core.Partition;
+import com.hazelcast.jet.AbstractProcessor;
+import com.hazelcast.jet.Processor;
+import com.hazelcast.jet.ProcessorMetaSupplier;
+import com.hazelcast.jet.ProcessorSupplier;
+import com.hazelcast.jet.function.DistributedFunction;
+import com.hazelcast.jet.function.DistributedPredicate;
+import com.hazelcast.jet.impl.util.ExceptionUtil;
+import com.hazelcast.jet.processor.Processors;
+import com.hazelcast.jet.processor.Sources;
+import com.hazelcast.journal.EventJournalInitialSubscriberState;
+import com.hazelcast.journal.EventJournalReader;
+import com.hazelcast.map.journal.EventJournalMapEvent;
+import com.hazelcast.nio.Address;
+import com.hazelcast.projection.Projection;
+import com.hazelcast.ringbuffer.ReadResultSet;
+import com.hazelcast.ringbuffer.StaleSequenceException;
+import com.hazelcast.util.concurrent.BackoffIdleStrategy;
+import com.hazelcast.util.concurrent.IdleStrategy;
+
+import javax.annotation.Nonnull;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+import static com.hazelcast.client.HazelcastClient.newHazelcastClient;
+import static com.hazelcast.jet.impl.util.Util.processorToPartitions;
+import static java.util.concurrent.TimeUnit.MICROSECONDS;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.mapping;
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toMap;
+import static java.util.stream.IntStream.range;
+
+/**
+ * @see Sources#streamMap(String, DistributedPredicate, DistributedFunction, boolean)
+ */
+public final class StreamEventJournalP<E, T> extends AbstractProcessor {
+
+    private static final IdleStrategy IDLER =
+            new BackoffIdleStrategy(0, 0, MICROSECONDS.toNanos(1), MILLISECONDS.toNanos(100));
+    private static final int MIN_FETCH_SIZE = 0;
+    private static final int MAX_FETCH_SIZE = 100;
+
+    private final EventJournalReader<E> eventJournalReader;
+    private final List<Integer> partitions;
+    private final Map<Integer, Long> offsetMap = new HashMap<>();
+    private final SerializablePredicate<E> predicate;
+    private final Projection<E, T> projection;
+    private final boolean startFromLatestSequence;
+
+    private CompletableFuture<Void> jobFuture;
+    private boolean fetched;
+
+    private StreamEventJournalP(EventJournalReader<E> eventJournalReader,
+                                List<Integer> partitions,
+                                Predicate<E> predicate,
+                                Function<E, T> projectionF,
+                                boolean startFromLatestSequence) {
+        this.eventJournalReader = eventJournalReader;
+        this.partitions = partitions;
+        this.predicate = predicate != null ? predicate::test : null;
+        this.projection = createProjection(projectionF);
+        this.startFromLatestSequence = startFromLatestSequence;
+    }
+
+    @Override
+    protected void init(@Nonnull Context context) throws Exception {
+        jobFuture = context.jobFuture();
+
+        List<ICompletableFuture<EventJournalInitialSubscriberState>> futures = new ArrayList<>(partitions.size());
+        for (Integer partitionId : partitions) {
+            futures.add(eventJournalReader.subscribeToEventJournal(partitionId));
+        }
+        for (int i = 0; i < futures.size(); i++) {
+            ICompletableFuture<EventJournalInitialSubscriberState> future = futures.get(i);
+            offsetMap.put(partitions.get(i), getSequence(future.get()));
+        }
+    }
+
+    private long getSequence(EventJournalInitialSubscriberState state) {
+        if (startFromLatestSequence) {
+            return state.getNewestSequence() == -1 ? 0 : state.getNewestSequence();
+        }
+        return state.getOldestSequence();
+    }
+
+    @Override
+    public boolean complete() {
+        long idleCount = 0;
+        while (!jobFuture.isDone()) {
+            boolean fetched = poll();
+            if (fetched) {
+                idleCount = 0;
+            } else {
+                IDLER.idle(++idleCount);
+            }
+        }
+        return true;
+    }
+
+    private boolean poll() {
+        Map<Integer, ICompletableFuture<ReadResultSet<T>>> futureMap =
+                offsetMap.entrySet().stream().collect(toMap(Map.Entry::getKey,
+                        e -> eventJournalReader.readFromEventJournal(e.getValue(),
+                                MIN_FETCH_SIZE, MAX_FETCH_SIZE, e.getKey(), predicate, projection)
+                ));
+        fetched = false;
+        futureMap.forEach((key, future) -> {
+                    try {
+                        ReadResultSet<T> resultSet = future.get();
+                        resultSet.forEach(this::emitEvent);
+                        offsetMap.compute(key, (k, v) -> v + resultSet.readCount());
+                    } catch (ExecutionException e) {
+                        if (e.getCause() instanceof StaleSequenceException) {
+                            long headSeq = ((StaleSequenceException) e.getCause()).getHeadSeq();
+                            long oldOffset = offsetMap.put(key, headSeq);
+                            getLogger().severe("Stale sequence, requested: " + oldOffset + ", currentHead: " + headSeq);
+                            return;
+                        }
+                        throw ExceptionUtil.rethrow(e);
+                    } catch (InterruptedException e) {
+                        throw ExceptionUtil.rethrow(e);
+                    }
+                }
+        );
+        return fetched;
+    }
+
+    private void emitEvent(T event) {
+        emit(event);
+        fetched = true;
+    }
+
+    @Override
+    public boolean isCooperative() {
+        return false;
+    }
+
+    private static <E, T> Projection<E, T> createProjection(Function<E, T> projectionF) {
+        if (projectionF == null) {
+            return null;
+        }
+        return new Projection<E, T>() {
+            @Override public T transform(E input) {
+                return projectionF.apply(input);
+            }
+        };
+    }
+
+    private static <E, T> List<Processor> getProcessors(int count, List<Integer> ownedPartitions,
+                                                        EventJournalReader<E> eventJournalReader,
+                                                        Predicate<E> predicate,
+                                                        Function<E, T> projection,
+                                                        boolean startFromLatestSequence) {
+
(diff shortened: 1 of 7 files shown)


gurbuzali:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,422 @@
...
+    private final Projection<E, T> projection;
+
+    private final boolean newestSequence;
+
+    private CompletableFuture<Void> jobFuture;
+
+    private StreamEventJournalP(EventJournalReader<E> eventJournalReader, List<Integer> partitions,
+                                Predicate<E> predicate,
+                                Function<E, T> projectionF,
+                                boolean newestSequence) {
```
cangencer: this boolean should have a clearer name

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,422 @@
...
+import static com.hazelcast.jet.Util.entry;
+import static java.util.Collections.emptyList;
+import static java.util.concurrent.TimeUnit.MICROSECONDS;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.mapping;
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.IntStream.range;
+
+public final class StreamEventJournalP<E, T> extends AbstractProcessor {
```
cangencer: missing javadoc

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,422 @@
...
+    private StreamEventJournalP(EventJournalReader<E> eventJournalReader, List<Integer> partitions,
+                                Predicate<E> predicate,
+                                Function<E, T> projectionF,
+                                boolean newestSequence) {
+        this.eventJournalReader = eventJournalReader;
+        this.offsetMap = new HashMap<>();
+        this.predicate = predicate != null ? predicate::test : null;
+        this.projection = createProjection(projectionF);
+        this.newestSequence = newestSequence;
+        subscribe(partitions);
```
cangencer: subscription should be done on `init`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,422 @@
...
+                                                        DistributedFunction<EventJournalCacheEvent, T> projection,
+                                                        boolean newestSequence) {
+        return new RemoteClusterMetaSupplier<>(clientConfig,
+                instance -> (EventJournalReader<EventJournalCacheEvent>) instance.getCacheManager().getCache(cacheName),
+                predicate, projection, newestSequence);
+    }
+
+    @Override
+    protected void init(@Nonnull Context context) throws Exception {
+        super.init(context);
```
cangencer: not needed as `super.init()` doesn't do anything

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,422 @@
...
+        this.eventJournalReader = eventJournalReader;
+        this.offsetMap = new HashMap<>();
+        this.predicate = predicate != null ? predicate::test : null;
+        this.projection = createProjection(projectionF);
+        this.newestSequence = newestSequence;
+        subscribe(partitions);
+
+    }
+
+    public static <T> ProcessorMetaSupplier streamMap(String mapName,
```
cangencer: can we move the static utility methods to the end of the file, as they interfere with the flow of the file

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,422 @@
...
+        List<ICompletableFuture<EventJournalInitialSubscriberState>> futures = new ArrayList<>();
+        try {
+            for (Integer partitionId : partitions) {
+                futures.add(eventJournalReader.subscribeToEventJournal(partitionId));
+            }
+            for (int i = 0; i < futures.size(); i++) {
+                ICompletableFuture<EventJournalInitialSubscriberState> future = futures.get(i);
+                offsetMap.put(partitions.get(i), getSequence(future.get()));
+            }
+        } catch (Exception e) {
```
cangencer: should replace this with `uncheckedRun()` and limit it to where an exception is thrown rather than whole method

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,422 @@
...
+            if (fetched) {
+                idleCount = 0;
+            } else {
+                IDLER.idle(++idleCount);
+            }
+        }
+        return true;
+    }
+
+    private boolean poll(Consumer<T> consumer) {
```
cangencer: what's the use of `Consumer` here? you can just call `emit` directly.. as there is ever only one consumer

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,422 @@
...
+        for (Map.Entry<Integer, Long> entry : offsetMap.entrySet()) {
+            int partitionId = entry.getKey();
+            long offset = entry.getValue();
+            ICompletableFuture<ReadResultSet<T>> future = eventJournalReader.readFromEventJournal(offset,
+                    MIN_FETCH_SIZE, MAX_FETCH_SIZE, partitionId, predicate, projection);
+            futureMap.put(partitionId, future);
+        }
+        boolean fetched = false;
+        for (Map.Entry<Integer, ICompletableFuture<ReadResultSet<T>>> entry : futureMap.entrySet()) {
+            Integer partitionId = entry.getKey();
```
cangencer: variable can be inlined

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,422 @@
...
+            }
+        };
+    }
+
+    private static <E, T> List<Processor> getProcessors(int count, List<Integer> ownedPartitions,
+                                                        EventJournalReader<E> eventJournalReader,
+                                                        Predicate<E> predicate,
+                                                        Function<E, T> projection,
+                                                        boolean newestSequence) {
+        Map<Integer, List<Integer>> processorToPartitions = range(0, ownedPartitions.size())
```
cangencer: maybe this can be extracted to a utility function? same logic duplicated in `ReadWithPartitionIteratorP`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/processor/Sources.java
```diff
@@ -16,16 +16,21 @@
...
+import com.hazelcast.jet.function.DistributedFunction;
+import com.hazelcast.jet.function.DistributedPredicate;
 import com.hazelcast.jet.function.DistributedSupplier;
 import com.hazelcast.jet.impl.connector.ReadFilesP;
 import com.hazelcast.jet.impl.connector.ReadIListP;
 import com.hazelcast.jet.impl.connector.ReadWithPartitionIteratorP;
+import com.hazelcast.jet.impl.connector.StreamEventJournalP;
 import com.hazelcast.jet.impl.connector.StreamFilesP;
 import com.hazelcast.jet.impl.connector.StreamSocketP;
+import com.hazelcast.map.impl.journal.EventJournalMapEvent;
```
cangencer: EventJournalMapEvent should not be in impl package as it's part of public API

hazelcast-jet-core/src/main/java/com/hazelcast/jet/processor/Sources.java
```diff
@@ -41,6 +46,126 @@ private Sources() {
...
+     * no partitions assigned to them.
+     * <p>
+     * In order to stream from a map, event-journal should be configured
+     * Please see {@link com.hazelcast.config.EventJournalConfig}
+     */
+    @Nonnull
+    public static <T> ProcessorMetaSupplier streamMap(@Nonnull String mapName,
+                                                      DistributedPredicate<EventJournalMapEvent> predicate,
+                                                      DistributedFunction<EventJournalMapEvent, T> projection,
+                                                      boolean newestSequence) {
```
cangencer: name for `newestSequence` should be improved. Also missing javadoc for params.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
gurbuzali: this method was in another class, looks like I've forgot to remove the consumer

gurbuzali:
verify

gurbuzali:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,408 @@
...
+import static java.util.concurrent.TimeUnit.MICROSECONDS;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.mapping;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * @see Sources#streamMap(String, DistributedPredicate, DistributedFunction, boolean)
+ */
+public final class StreamEventJournalP<E, T> extends AbstractProcessor {
```
mmedenjak: I'm not a big fan of the size of this class but I guess that's how you roll

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,408 @@
...
+
+    private long getSequence(EventJournalInitialSubscriberState state) {
+        if (startFromLatestSequence) {
+            return state.getNewestSequence() == -1 ? 0 : state.getNewestSequence();
+        }
+        return state.getOldestSequence();
+    }
+
+    @Override
+    public boolean complete() {
```
mmedenjak: So an infinite source processor only needs to implement this method?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,408 @@
...
+
+    private boolean poll() {
+        Map<Integer, ICompletableFuture<ReadResultSet<T>>> futureMap = new HashMap<>();
+        for (Map.Entry<Integer, Long> entry : offsetMap.entrySet()) {
+            int partitionId = entry.getKey();
+            long offset = entry.getValue();
+            ICompletableFuture<ReadResultSet<T>> future = eventJournalReader.readFromEventJournal(offset,
+                    MIN_FETCH_SIZE, MAX_FETCH_SIZE, partitionId, predicate, projection);
+            futureMap.put(partitionId, future);
+        }
```
mmedenjak: If you're fond of j.u.s, you can just do :
```
final Map<Integer, ICompletableFuture<ReadResultSet<T>>> futureMap =
                offsetMap.entrySet().stream().collect(Collectors.toMap(
                        Entry::getKey,
                        e -> eventJournalReader.readFromEventJournal(e.getValue(),
                                MIN_FETCH_SIZE, MAX_FETCH_SIZE, e.getKey(), predicate, projection)
                ));
```

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,408 @@
...
+                for (T event : resultSet) {
+                    emit(event);
+                    fetched = true;
+                }
+                offsetMap.compute(entry.getKey(), (k, v) -> v + resultSet.readCount());
+            } catch (ExecutionException e) {
+                if (e.getCause() instanceof StaleSequenceException) {
+                    long headSeq = ((StaleSequenceException) e.getCause()).getHeadSeq();
+                    long oldOffset = offsetMap.put(entry.getKey(), headSeq);
+                    getLogger().severe("Stale sequence, requested: " + oldOffset + ", currentHead: " + headSeq);
```
mmedenjak: You also have a somewhat ok message in `((StaleSequenceException) e.getCause()).getMessage()` if you like

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,408 @@
...
+                    MIN_FETCH_SIZE, MAX_FETCH_SIZE, partitionId, predicate, projection);
+            futureMap.put(partitionId, future);
+        }
+        boolean fetched = false;
+        for (Map.Entry<Integer, ICompletableFuture<ReadResultSet<T>>> entry : futureMap.entrySet()) {
+            ICompletableFuture<ReadResultSet<T>> future = entry.getValue();
+            try {
+                ReadResultSet<T> resultSet = future.get();
+                for (T event : resultSet) {
+                    emit(event);
```
mmedenjak: So since this is a non-cooperative processor, we don't need to yield the thread back to the engine and keep the rejected event in a field?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,408 @@
...
+                    fetched = true;
+                }
+                offsetMap.compute(entry.getKey(), (k, v) -> v + resultSet.readCount());
```
mmedenjak: `v + resultSet.readCount()` - this is ok, you can also get the last sequence from the result set and put if you prefer (something like this) : 
`offsetMap.put(entry.getKey(), resultSet.getSequence(resultSet.size() - 1) + 1);`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/processor/Sources.java
```diff
@@ -44,6 +47,152 @@ private Sources() {
...
+    public static ProcessorMetaSupplier streamMap(@Nonnull String mapName) {
+        return StreamEventJournalP.streamMap(mapName, null, null, false);
+    }
+
+    /**
+     * Returns a meta-supplier of processor that will stream the events
+     * of the Hazelcast {@code IMap} with the specified name.
+     * Given predicate and projection will be applied to the events on the source.
+     * <p>
+     * The processors will only access data local to the
```
gurbuzali: we have it for local streaming/reading from map/cache, if we are streaming/reading from a remote cluster it is irrelevant

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
gurbuzali: regardless of finite(batch) or infinite, a `source` processors should implement just `complete` method. If you never return `true` then it is infinite.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
gurbuzali: since this is a non-cooperative processor, events won't get rejected, `emit` will block until there is room in the queue.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -0,0 +1,408 @@
...
+
+        @Override @Nonnull
+        public Function<Address, ProcessorSupplier> get(@Nonnull List<Address> addresses) {
+            // assign each remote partition to a member
+            Map<Address, List<Integer>> membersToPartitions =
+                    IntStream.range(0, remotePartitionCount)
+                             .boxed()
+                             .collect(groupingBy(partition -> addresses.get(partition % addresses.size())));
+
+            return address -> new RemoteClusterProcessorSupplier<>(membersToPartitions.get(address),
```
gurbuzali: not necessarily

mmedenjak: That's what I thought - but can we distribute the partitions then according to the partition table or do we not care about that?

gurbuzali: we don't care because it is already a remote call through client.

gurbuzali:
run-lab-run

PR #489, opened 2017-07-18
[WIP] Introduce Aeron source/sink

gurbuzali:


gurbuzali:
Yes, this is not a final work. I was planning to ask you guys `multiple stream as a source` approach. I'll add some tests to show the example usage

PR #490, opened 2017-07-19
Add projection and predicate capability to read map

gurbuzali:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/processor/Sources.java
```diff
@@ -74,6 +101,22 @@ public static ProcessorMetaSupplier readMap(@Nonnull String mapName, @Nonnull Cl
...
     /**
+     * Returns a meta-supplier of processor that will fetch entries from a
+     * Hazelcast {@code IMap} in a remote cluster. Processor will emit the
+     * entries as {@code Map.Entry}.
+     * <p>
+     * If the underlying map is concurrently modified, there may be missing or
+     * duplicate items.
+     */
+    @Nonnull
+    public static <K, V, T> ProcessorMetaSupplier readMap(@Nonnull String mapName,
```
cangencer: javadoc is incomplete

hazelcast-jet-core/src/main/java/com/hazelcast/jet/processor/Sources.java
```diff
@@ -61,9 +64,33 @@ public static ProcessorMetaSupplier readMap(@Nonnull String mapName) {
...
+     * The number of Hazelcast partitions should be configured to at least
+     * {@code localParallelism * clusterSize}, otherwise some processors will have
+     * no partitions assigned to them.
+     * <p>
+     * If the underlying map is concurrently being modified, there are no guarantees
+     * given with respect to missing or duplicate items.
+     */
+    @Nonnull
+    public static <K, V, T> ProcessorMetaSupplier readMap(@Nonnull String mapName,
+                                                          @Nonnull DistributedFunction<Map.Entry<K, V>, T> projectionF,
```
cangencer: can't either projection or predicate be null? if this is allowed, this should also be added as `@param` tags to the javadoc

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadWithPartitionIteratorP.java
```diff
@@ -143,6 +177,14 @@ public boolean isCooperative() {
                 .collect(toList());
     }
 
+    private static <I, O> Projection<I, O> projection(DistributedFunction<I, O> projectionF) {
```
cangencer: method can be renamed to `toProjection`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/processor/Sources.java
viliam-durina: We can also mention the full note from partition iterator: 
>Iterating the map should be done only when the {@link IMap} is not being
mutated and the *cluster is stable* (there are no migrations or membership changes).
In other cases, the iterator may not return some entries or may return an entry twice.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/processor/Sources.java
gurbuzali: I've checked the implementation on hazelcast side, both should be nonnull. Maybe we can provide it by passing a default implementation, wdyt ?

viliam-durina: We have similar concept here: https://github.com/hazelcast/hazelcast-jet/blob/master/hazelcast-jet-core/src/main/java/com/hazelcast/jet/processor/DiagnosticProcessors.java#L94-L97
You argued that requiring @Nonnull is better, but I don't like it at all. It's much easier to write `null` to predicate than `() -> true`.

gurbuzali: I was referring to have overloaded versions of `readMap` but I guess it'll be a lot of overloading.

cangencer: I think both should be `@Nullable`

gurbuzali: Having `@Nullable` is not inline with our other processors, why do we prefer `@Nullable` here but `@Nonnull` for other processors ? I don't have a strong opinion for both options but I favor `@Nonnull`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadWithPartitionIteratorP.java
```diff
@@ -86,6 +88,21 @@ public static ProcessorMetaSupplier readMap(String mapName) {
                         .iterator(fetchSize, partition, PREFETCH_VALUES));
     }
 
+    public static <K, V, T> ProcessorMetaSupplier readMap(String mapName,
```
cangencer: can we add `@Nonnull` annotations here as well?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/processor/Sources.java
```diff
@@ -52,28 +55,72 @@ private Sources() {
...
+     * {@code localParallelism * clusterSize}, otherwise some processors will have
+     * no partitions assigned to them.
+     * <p>
+     * Iterating the map should be done only when the {@link com.hazelcast.core.IMap} is not being
+     * mutated and the cluster is stable (there are no migrations or membership changes).
+     * In other cases, the iterator may not return some entries or may return an entry twice.
+     */
+    @Nonnull
+    public static <K, V, T> ProcessorMetaSupplier readMap(@Nonnull String mapName,
+                                                          @Nonnull DistributedFunction<Map.Entry<K, V>, T> projectionF,
```
cangencer: I would prefer the argument order to be predicate and then projection. Because the predicate is applied first, and then the projection.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/stream/IStreamMap.java
```diff
@@ -38,4 +42,17 @@
...
+     * Returns a parallel and distributed {@code Stream} with this map as its source.
+     * Entries will be filtered and mapped according to the given predicate and projection.
+     * <p>
+     * If the underlying map is concurrently being modified, there are no guarantees
+     * given with respect to missing or duplicate items in a stream operation.
+     *
+     * @return a parallel {@code Stream} over the elements in this collection
+     * @since 1.8
+     */
+    <T> DistributedStream<T> stream(@Nonnull DistributedFunction<Entry<K, V>, T> projectionF,
```
cangencer: see my comments about argument ordering under `Sources.readMap(..)`

PR #502, opened 2017-09-19
Add stream support for all sources

gurbuzali:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/stream/DistributedStream.java
```diff
@@ -60,6 +67,33 @@
...
+            protected String getName() {
+                return UuidUtil.newUnsecureUuidString();
+            }
+        };
+    }
+
+    /**
+     * Returns a distributed {@code Stream} with given {@code Source} as its source.
+     */
+    static <T> DistributedStream<T> fromSource(JetInstance instance, Source<T> source) {
```
cangencer: I thought about this and let's get rid of this link here. No need to intersect pipeline API with java.util.stream. What do you think?

gurbuzali: removed it

PR #505, opened 2017-09-20
Simplify adjacency map logic

mtopolnik:
TopologicalSorter now demands a fully populated map: every vertex appearing in some value must appear as a key as well.

PipelineImpl now makes a safe copy of its adjacency map.


mtopolnik:
verify

PR #506, opened 2017-09-21
Implement restoreFromSnapshot in AbstractProcessor

cangencer:
rename restoreSnapshot -> restoreFromSnapshot and
saveSnapshot -> saveToSnapshot


hazelcast-jet-core/src/main/java/com/hazelcast/jet/AbstractProcessor.java
```diff
@@ -306,6 +307,30 @@ protected boolean tryProcessWm4(@Nonnull Watermark wm) {
...
+     */
+    @Override
+    public final void restoreFromSnapshot(@Nonnull Inbox inbox) {
+        for (Map.Entry entry; (entry = (Map.Entry) inbox.poll()) != null; ) {
+            restoreFromSnapshot(entry.getKey(), entry.getValue());
+        }
+    }
+
+    /**
+     * Tries to restore the given key and value received from the
```
mtopolnik: Since it's `void`, it can't just "try to restore".

PR #511, opened 2017-09-22
Log if current cluster quorum value passes quorum value of a jet job

metanet:
If a job enables split brain protection, JobCoordinationService prints a log when the current quorum value of the cluster becomes bigger than quorum value of the job. The log is printed because the job loses its split brain protection feature in this case. If the cluster splits into partitions, the job can be successfully restarted on both sides.


metanet:
run-lab-run

metanet:
verify

metanet:
pr builder is still running => https://hazelcast-l337.ci.cloudbees.com/job/Jet-pr-builder/1822/console

metanet:
verify

PR #516, opened 2017-09-26
Reorganize serialization hooks

mtopolnik:
Put serializer hook classes in the same package as the type they are responsible for. Place hook classes into a containing utility class.

Also add serializer hooks for classes introduced with the Pipeline API.


mtopolnik:
verify

mtopolnik:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/datamodel/DataModelSerializerHooks.java
```diff
@@ -0,0 +1,364 @@
...
+                out.writeObject(val != NONE ? val : null);
+            }
+        }
+
+        @Override
+        public ItemsByTag read(ObjectDataInput in) throws IOException {
+            int size = in.readInt();
+            ItemsByTag ibt = new ItemsByTag();
+            for (int i = 0; i < size; i++) {
+                ibt.put(in.readObject(), in.readObject());
```
cangencer: shouldn't it be set to `NONE` when reading?

mtopolnik: No because that's done internally by `put()` There's an asymmetry between writing and reading because writing is done by iterating the raw internal map, whereas reading uses the public API.

PR #518, opened 2017-09-27
Delay job start if cluster is not safe

metanet:
If the cluster is not safe when a job is submitted, it could happen that the job could be started before partition re-distribution is completed. To prevent such non-optimal runs, we delay starting execution of a job if the cluster is not safe at job submission time.

Recent `warmUpPartitions()` changes in the test classes are also reverted.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
@@ -400,6 +400,18 @@ public class JobCoordinationService {
         }
     }
 
+    boolean shouldStartJobs() {
+        if (!(isMaster() && nodeEngine.isRunning())) {
+            return false;
+        }
+
+        Node node = nodeEngine.getNode();
+        InternalPartitionServiceImpl partitionService = (InternalPartitionServiceImpl) node.getPartitionService();
+        return partitionService.getPartitionStateManager().isInitialized()
+                && partitionService.isMigrationAllowed()
+                && !partitionService.hasOnGoingMigrationLocal();
+    }
+
     /**
      * Restarts a job for a new execution if the cluster is stable.
      * Otherwise, it reschedules the restart task.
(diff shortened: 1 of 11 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -400,6 +400,18 @@ void completeSnapshot(long jobId, long executionId, long snapshotId, boolean isS
         }
     }
 
+    boolean shouldStartJobs() {
```
viliam-durina: You can remove the `shouldStartJobs` from `restartJob` method in this class.

metanet: good spotting

metanet: your comment is addressed.

cangencer:
verify

cangencer:
verify

metanet:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -127,18 +127,31 @@ public JobResult getJobResult(long jobId) {
...
+                                + idToString(jobRecord.getJobId()));
+                    }
                 }
             }
+        } catch (Exception e) {
+            logger.fine("check quorum values task failed", e);
         }
     }
 
+    private boolean shouldCheckQuorumValues() {
```
viliam-durina: I don't like these single-use private methods just for the sake of using method name as a comment, but it seems everyone else does...

metanet: I understand your concern, but writing single-use methods for checking some conditions is a good use case.

PR #521, opened 2017-09-28
Add id field to BroadcastKey

cangencer:
SlidingWindowP instances should save unique ‘nextWmToEmit’ values and
 validate they are all same on snapshot restore.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
@@ -126,7 +126,7 @@ public interface Processor {
      * <p>
      * Non-cooperative processors are required to return from this method from
      * time to time to give the system a chance to check for snapshot requests
-     * job cancellation. The time the processor spends in this method affects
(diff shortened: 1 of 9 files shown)


cangencer:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
```diff
@@ -173,10 +173,11 @@ default boolean isCooperative() {
...
      * <p>
-     * Processors must ensure they give a chance to the system to call this
-     * method. Especially, non-cooperative source processors must make sure not
-     * to spend too much time per invocation of {@code complete()}, as this
-     * will prolong the snapshotting phase and impede overall progress.
+     * Processors which override this method must ensure they give a chance
+     * to the system to call this method. Especially, non-cooperative
+     * source processors must make sure not to spend too much time per invocation
+     * of {@code complete()}, as this will prolong the snapshotting phase
+     * and impede overall progress.
```
viliam-durina: We can remove this paragraph altogether. The need to return from `complete()` is already documented there.

PR #522, opened 2017-09-28
Make sure that all whenComplete calls wrap the callback in “withTryCatch”

cangencer:
This ensures that any exception which may occur during a whenComplete 
call is handled.


cangencer:
verify

PR #523, opened 2017-09-28
Allow snapshotDone before snapshotStarted

viliam-durina:
The `SnapshotContext.startNewSnapshot()` method can be called _after_ `taskletDone()` or `snapshotDoneForTasklet()` is called. This can happen in a situation when a processor only has input queues from remote members and the remote members happen to process `SnapshotOperation` and send barriers to such processor before the `SnapshotOperation` is called on this member.

### DIFF
--- a/hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/execution/SnapshotContextTest.java
+++ b/hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/execution/SnapshotContextTest.java
@@ -18,37 +18,98 @@ package com.hazelcast.jet.impl.execution;
 
 import com.hazelcast.jet.config.ProcessingGuarantee;
 import com.hazelcast.logging.ILogger;
-import org.junit.After;
-import org.junit.Before;
+import com.hazelcast.test.HazelcastParametersRunnerFactory;
+import com.hazelcast.test.annotation.ParallelTest;
+import com.hazelcast.test.annotation.QuickTest;
 import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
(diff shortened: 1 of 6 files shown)


viliam-durina:
verify
(The build was aborted)

viliam-durina:
verify

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/execution/SnapshotContextTest.java
```diff
@@ -18,37 +18,98 @@
...
+                }
+            }
+        }
+        return res;
     }
 
     @Test
-    public void when_taskletDoneBeforeItDidCurrentSnapshot_then_countedAsSnapshotWasDone() {
-        ssContext.taskletDone(9, false);
+    public void test() {
```
cangencer: more descriptive test name here would be good

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/execution/SnapshotContextTest.java
```diff
@@ -18,37 +18,98 @@
...
+                }
+            }
+        }
+        return res;
     }
 
     @Test
-    public void when_taskletDoneBeforeItDidCurrentSnapshot_then_countedAsSnapshotWasDone() {
-        ssContext.taskletDone(9, false);
+    public void test_snapShortStartAndDone() {
```
cangencer: This test should be broken down into several tests. I don't mind some duplication in test code, having so many parameters and conditionals makes the test hard to reason about.

cangencer: We'll consider this good enough for now. It was not trivial to break down the tests and test the same number of combinations.

PR #524, opened 2017-09-29
Fix snapshot in streamKafka & test

viliam-durina:
Also includes some clean-up in ArrayDequeInbox

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/CloseableProcessorSupplier.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/CloseableProcessorSupplier.java
@@ -37,10 +37,11 @@ import static java.util.stream.Collectors.toList;
 public class CloseableProcessorSupplier<E extends Processor & Closeable> implements ProcessorSupplier {
 
     static final long serialVersionUID = 1L;
+
     private final DistributedIntFunction<Collection<E>> supplier;
 
-    private ILogger logger;
-    private Collection<E> processors;
+    private transient ILogger logger;
+    private transient Collection<E> processors;
 
     /**
      * @param simpleSupplier Supplier to create processor instances.
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/CloseableProcessorSupplier.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/CloseableProcessorSupplier.java
@@ -14,23 +14,39 @@
  * limitations under the License.
  */
 
-package com.hazelcast.jet.impl.processor;
+package com.hazelcast.jet.core;
 
-import com.hazelcast.jet.core.Processor;
-import com.hazelcast.jet.core.ProcessorSupplier;
-import com.hazelcast.jet.function.DistributedIntFunction;
+import com.hazelcast.jet.function.DistributedSupplier;
 import com.hazelcast.logging.ILogger;
 
 import javax.annotation.Nonnull;
 import java.io.Closeable;
 import java.util.Collection;
+import java.util.stream.IntStream;
 
 import static com.hazelcast.jet.impl.util.ExceptionUtil.sneakyThrow;
+import static java.util.stream.Collectors.toList;
 
-public abstract class AbstractCloseableProcessorSupplier<E extends Processor & Closeable> implements ProcessorSupplier {
+/**
+ * A {@link ProcessorSupplier} which closes created processor instances
+ * when the job is complete.
+ *
+ * @param <E> Processor type
+ */
+public class CloseableProcessorSupplier<E extends Processor & Closeable> implements ProcessorSupplier {
+
+    static final long serialVersionUID = 1L;
+    private final DistributedSupplier<E> supplier;
 
     private ILogger logger;
     private Collection<E> processors;
+
+    /**
+     * @param supplier Supplier to create processor instances.
+     */
+    public CloseableProcessorSupplier(DistributedSupplier<E> supplier) {
+        this.supplier = supplier;
+    }
 
     @Override
     public final void init(@Nonnull Context context) {
@@ -39,7 +55,9 @@ public abstract class AbstractCloseableProcessorSupplier<E extends Processor & C
 
     @Nonnull @Override
     public final Collection<E> get(int count) {
-        return processors = getProcessors(count);
+        return processors = IntStream.range(0, count)
+                                     .mapToObj(i -> supplier.get())
+                                     .collect(toList());
     }
 
     @Override
@@ -62,22 +80,4 @@ public abstract class AbstractCloseableProcessorSupplier<E extends Processor & C
         }
     }
 
-    protected void initInternal(@Nonnull Context context) {
-    }
-
-    protected abstract Collection<E> getProcessors(int count);
-
-    protected void completeInternal(Throwable error) {
-    }
-
-    public static <E extends Processor & Closeable> ProcessorSupplier closeableSupplier(
-            DistributedIntFunction<Collection<E>> processorSupplier
-    ) {
-        return new AbstractCloseableProcessorSupplier<E>() {
-            @Override
-            protected Collection<E> getProcessors(int count) {
-                return processorSupplier.apply(count);
-            }
-        };
-    }
 }
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/CloseableProcessorSupplier.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/CloseableProcessorSupplier.java
@@ -78,7 +78,6 @@ public class CloseableProcessorSupplier<E extends Processor & Closeable> impleme
         }
 
         Throwable firstError = null;
-        // close all processors, ignoring their failures and throwing the first failure (if any)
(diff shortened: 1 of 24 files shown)


hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java
```diff
@@ -91,36 +96,45 @@
...
             List<PartitionInfo> partitionInfos = consumer.partitionsFor(topicName);
-            getLogger().finest("Num of ptions for topic " + topicName + ": " + partitionInfos.size());
-            while (nextAssignablePtions[topicIdx] < partitionInfos.size()) {
-                int partition = nextAssignablePtions[topicIdx];
-                nextAssignablePtions[topicIdx] += processorCount;
-                assert partitionInfos.get(partition).partition() == partition;
-                assignment.add(new TopicPartition(topicName, partition));
-                changed = true;
+            LoggingUtil.logFinest(getLogger(), "Num of ptions for topic '%s': %d", topicName, partitionInfos.size());
+            if (nextAssignablePtions[topicIdx] < partitionInfos.size()) {
```
mtopolnik: Here you can reduce the nesting level by inverting the `if` and using `continue`.

hazelcast-jet-kafka/src/test/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaPTest.java
```diff
@@ -94,6 +150,111 @@ public void run() throws Exception {
         assertTrueEventually(() -> assertTrue(jobFuture.isDone()));
     }
 
+    @Test
+    public void when_snapshotSaved_then_offsetsRestored() throws Exception {
+        StreamKafkaP[] processor = {new StreamKafkaP(properties, singletonList(topic1Name), 1, 0, 60000)};
+        TestOutbox[] outbox = {new TestOutbox(new int[] {10}, 10)};
```
mtopolnik: I can't make out a reason why these two are packed into an array.

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java
```diff
@@ -91,36 +96,45 @@
...
     }
 
     @Override
     public boolean complete() {
         if (System.nanoTime() >= nextPartitionCheck) {
-            reassignPartitions();
+            reassignPartitions(false);
         }
 
         if (!assignment.isEmpty()) {
```
mtopolnik: Here also the nesting level can be reduced by inverting the `if` and returning false after `parkNanos()`

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java
```diff
@@ -254,8 +272,8 @@ public MetaSupplier(Properties properties, List<String> topicIds) {
             } else {
                 metadataRefreshInterval = KAFKA_DEFAULT_REFRESH_INTERVAL;
             }
-            // Set metadata caching to 1 second: we get metadata for multiple partitions one by one, but
-            // internally it is fetched at once. This enables for the other calls to just query fetched metadata.
+            // Set metadata caching to 1 second: we read the metadata for multiple partitions one by one. If we
```
mtopolnik: We should consistently apply the 72-char text line length rule even to these comments.

hazelcast-jet-kafka/src/test/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaPTest.java
viliam-durina: They are both modified and used in lambdas.

viliam-durina:
verify

cangencer:
verify

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java
```diff
@@ -91,53 +96,64 @@
...
         }
 
-        if (changed) {
-            getLogger().info("Partition assignment changed: " + assignment);
+        if (!addedPartitions.isEmpty()) {
+            getLogger().info("Partition assignment changed, new partitions: " + addedPartitions);
+            assignment.addAll(addedPartitions);
             consumer.assign(assignment);
+            if (!isInitial) {
+                consumer.seekToBeginning(addedPartitions);
```
cangencer: does this call take `auto.offset.reset` into account?

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java
```diff
@@ -91,53 +96,64 @@
     protected void init(@Nonnull Context context) throws Exception {
         snapshottingEnabled = context.snapshottingEnabled();
         consumer = new KafkaConsumer<>(properties);
-        reassignPartitions();
+        reassignPartitions(true);
```
cangencer: wouldn't it make more sense to do partition assignment in the Supplier/MetaSupplier. It looks like we're relying on each node independently deciding what partitions it should subscribe to?

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java
```diff
@@ -155,17 +171,21 @@ public boolean isCooperative() {
     public boolean saveToSnapshot() {
         if (snapshotTraverser == null) {
             snapshotTraverser = Traversers.traverseIterable(offsets.entrySet())
-                    .onFirstNull(() -> snapshotTraverser = null);
+                                          .map(e -> entry(broadcastKey(e.getKey()), e.getValue()))
+                                          .onFirstNull(() -> snapshotTraverser = null);
         }
         return emitFromTraverserToSnapshot(snapshotTraverser);
```
cangencer: shouldn't the offsets be commited after they have been saved to snapshot?

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java
viliam-durina: 1. you only know the partitions when you have a `consumer`. You don't have one in Supplier. 
2. we regularly check for new partitions, kafka supports adding partitions during runtime (unlike HZ)

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java
viliam-durina: Intentionally not. This is for partitions added after the initial assignment: we want to read them from the beginning.

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java
viliam-durina: No need to: we don't use them. I commit them when snapshots are disabled to have "approximately once" guarantee.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/CloseableProcessorSupplier.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @Override
+    public void complete(Throwable error) {
+        if (processors == null) {
+            return;
+        }
+
+        Throwable firstError = null;
+        // close all processors, ignoring their failures and throwing the first failure (if any)
```
mtopolnik: This is a self-contradictory comment. If we throw, we don't ignore. I prefer to just delete the comment, the code speaks for itself.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlanBuilder.java
```diff
@@ -78,7 +79,9 @@ private ExecutionPlanBuilder() {
             final List<EdgeDef> outbound = toEdgeDefs(dag.getOutboundEdges(vertex.getName()), defaultEdgeConfig,
                     e -> vertexIdMap.get(e.getDestName()), isJobDistributed);
             final ProcessorMetaSupplier metaSupplier = vertex.getSupplier();
-            metaSupplier.init(new MetaSupplierCtx(instance, totalParallelism, localParallelism,
+            ILogger logger = nodeEngine.getLogger(metaSupplier.getClass().getName() + "." + vertex.getName() +
+                    "#MetaProcessorSupplier");
```
mtopolnik: `#ProcessorMetaSupplier`

PR #526, opened 2017-09-29
Add integration test for postponed snapshot

viliam-durina:


hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/PostponedSnapshotTest.java
```diff
@@ -0,0 +1,114 @@
...
+        Job job = instance.newJob(dag, config);
+
+        IStreamMap<Object, Object> snapshotsMap = instance.getMap(SnapshotRepository.snapshotsMapName(job.getJobId()));
+
+        // check, that snapshot starts, but stays in ONGOING state
+        assertTrueEventually(() -> assertTrue(existsSnapshot(snapshotsMap, ONGOING)), 5);
+        Thread.sleep(1000);
+        assertFalse(existsSnapshot(snapshotsMap, SUCCESSFUL));
+
+        latches.set(0, 1);
```
cangencer: I'd prefer to use two separate latches here, and then have just a single CountdownLatch passed into SourceP.
Looks good otherwise.

viliam-durina: You can't pass `CountdownLatch` or `AtomicInteger` through serialization. It has to be shared static field.

cangencer: ok, I see your point.

PR #527, opened 2017-09-29
Fix SourceProcessors to return from complete()

viliam-durina:
All non-cooperative processors are now required to return from complete()
from time to time to allow for snapshot creation. This PR modifies the
processors to do this.


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadFilesP.java
```diff
@@ -103,13 +141,21 @@ public static ProcessorSupplier supplier(@Nonnull String directory, @Nonnull Str
...
-                return IntStream.range(0, count)
+                processors = IntStream.range(0, count)
                         .mapToObj(i -> new ReadFilesP(directory, charsetObj, glob, count, i))
                         .collect(Collectors.toList());
+                return processors;
+            }
+
+            @Override
+            public void complete(Throwable error) {
+                processors.forEach(p -> uncheckRun(p::close));
```
cangencer: I think we have this pattern in quite a few places. Maybe we should have a utility like CloseableProcessorSupplier?

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/impl/connector/hadoop/ReadHdfsP.java
```diff
@@ -83,7 +83,12 @@ private ReadHdfsP(@Nonnull List<RecordReader> recordReaders, @Nonnull Distribute
 
     @Override
     public boolean complete() {
-        return emitFromTraverser(trav);
+        R nextItem = trav.next();
```
cangencer: we should consider going back to non-blocking outbox here

cangencer: I will also add that we will be yielding with every time, causing additional overhead inside the Tasklet.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/CancellationTest.java
```diff
@@ -305,30 +308,49 @@ public boolean complete() {
...
-            while (!jobFuture.isDone()) {
-                parkNanos(MILLISECONDS.toNanos(200));
-            }
+        public void close() throws IOException {
             isDone = true;
-            return true;
+        }
+    }
+
+    // TODO replace with standard CloseableProcessorSupplier
```
cangencer: leftover TODO

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java
```diff
@@ -112,28 +115,52 @@ public boolean complete() {
...
+        if (traverser == null) {
+            ConsumerRecords<Object, Object> records = consumer.poll(POLL_TIMEOUT_MS);
+            if (records.isEmpty()) {
+                return false;
             }
-            emit(entry(r.key(), r.value()));
+            traverser = traverseIterable(records);
         }
+
+        mapAndEmitFromTraverser(traverser, r -> entry(r.key(), r.value()),
```
cangencer: can we use another pattern here maybe?
```final Record[] currentRecord = new Record[1];```
```emitFromTraverser(traverser.peek(r -> { currentRecord[0] = r; }).map(entry(r.key(), r.value())), e -> offset.put(currentRecord[0]..) };```

File: hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java [hunk 5]
```diff
@@ -104,6 +115,15 @@ private void assignPartitions(boolean seekToBeginning) {
                 consumer.seekToBeginning(newAssignments);
             }
         }
+        // initialize offsets
+        for (int topicIdx = 0; topicIdx < partitionCounts.size(); topicIdx++) {
+            int count = partitionCounts.get(topicIdx);
+            offsets.merge(topics.get(topicIdx), LongStream.generate(() -> -1).limit(count).toArray(),
```
cangencer: would appreciate a small comment here, took me a little time to clarify what it was doing (copying old offsets over)

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java
```diff
@@ -104,6 +115,15 @@ private void assignPartitions(boolean seekToBeginning) {
                 consumer.seekToBeginning(newAssignments);
             }
         }
+        // initialize offsets
```
cangencer: can remove comment and extract into method

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
```diff
@@ -42,30 +41,6 @@
...
- * <p>
- * By default the processor declares itself as "cooperative" ({@link
- * #isCooperative()} returns {@code true}). It will be assigned an outbox
- * of finite capacity which is not emptied until the processor yields back
- * to the execution engine. As soon as the outbox refuses an offered item,
- * the processor should save its current state and return to the caller.
- * It should also limit the amount of time it spends in the invocation of
- * any of its methods because it will participate in a cooperative
- * multithreading scheme, sharing a thread with other processors.
- * <p>
```
mtopolnik: This part shouldn't have been deleted, only rephrased.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -118,49 +113,52 @@ private long getSequence(EventJournalInitialSubscriberState state) {
...
     public boolean complete() {
-        long idleCount = 0;
-        while (!jobFuture.isDone()) {
-            boolean fetched = poll();
-            if (fetched) {
-                idleCount = 0;
-            } else {
-                IDLER.idle(++idleCount);
+        pollIfNeeded();
+        if (traverser == null) {
```
mtopolnik: There is too much nesting in this block of code, and the body of the `if` statement is very long. This should be simplified.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/OutboxImpl.java
```diff
@@ -84,6 +89,10 @@ public final boolean offer(int ordinal, @Nonnull Object item) {
 
     @Override
     public final boolean offer(int[] ordinals, @Nonnull Object item) {
+        if (numRemainingInBatch <= 0) {
```
mtopolnik: Is there any legal way for `numRemainingInBatch` to become negative? If not, the condition should be `== 0` and it can be followed with an `assert numRemainingInBatch > 0`.

cangencer: It can be if the user ignores the return value and pushes items. Agree there should be some assert here.

cangencer: It doesn't seem possible since it's returned immediately if it's at 0. But agree with the assertion.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
viliam-durina: It's duplicate. I put everything to `isCooperative()` method. We can maybe refer from class javadoc to that method.

PR #536, opened 2017-10-06
EventJournal snapshotting support

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -68,139 +75,216 @@
...
+        Traverser<T> traverser = traverseIterable(resultSet);
+        final ReadResultSet<T> currentSet = resultSet;
+        return peekIndex(traverser, i -> {
+            pendingOffset = currentSet.getSequence(i) + 1;
+        });
+    }
+
+    private ReadResultSet<T> nextResultSet() {
+        Iterator<Entry<Integer, ICompletableFuture<ReadResultSet<T>>>> iterator = readFutures.entrySet().iterator();
+        while (iterator.hasNext()) {
```
viliam-durina: This is prone to starvation. If the first future always manages to be done since the last call, for example when the downstream is slow, we won't touch another futures.

PR #541, opened 2017-10-09
Multiple fixes to snapshotting code

viliam-durina:
* Change contract of Processor.finishSnapshotRestore: now it is called
each time the processor state is restored from snapshot. Previously, it
was only called if there was at least one key in the state.

* Simplifications in StreamEventJournalP: we don't need two places to
check for stale offset. We restore whatever offset there is in the
snapshot, it will fail later when trying to read from it.

* Initialize currInstream in ProcessorTasklet: having it uninitialized
caused the `tryProcess` method to by called once before
restoreFromSnapshot.

* Remove unused method from Traverser and outstanding TODOs


viliam-durina:
Included in #542 by mistake

PR #544, opened 2017-10-12
Update quorum size of jobs when the Jet cluster scales up

metanet:
When a new member is added to the cluster, quorum sizes of split-brain protection enabled jobs can be invalidated. To prevent issues, JobCoordinationService automatically updates quorums.


cangencer:
verify

PR #546, opened 2017-10-13
Fix failure in SliWP for at-least-once jobs & docs

viliam-durina:
Document the possibility of watermarks going back after restart of a job with at-least-once guarantee. Fix the assert in SlidingWindowP to allow it.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/AbstractProcessor.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/AbstractProcessor.java
@@ -272,6 +272,14 @@ public abstract class AbstractProcessor implements Processor {
      * The default implementation just emits the watermark downstream
      * to all ordinals.
      * <p>
+     * <i>Caution for jobs with at-least-once guarantee</i><br>
+     * In snapshotted jobs with <i>at least once</i> processing guarantee it
+     * can happen that the watermarks break the monotonicity requirement when
+     * the job is restarted. This is caused by the fact that watermark, like any
+     * other stream item, can be delivered duplicately after restart. This
+     * means that after a restart, a processor can be asked to process a
+     * watermark older than it already processed.
+     * <p>
      * <strong>NOTE:</strong> unless the processor doesn't differentiate between
      * its inbound edges, the first choice should be leaving this method alone
      * and instead overriding the specific {@code tryProcessN()} methods for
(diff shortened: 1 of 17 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/AbstractProcessor.java
```diff
@@ -272,6 +272,14 @@ protected boolean tryProcess4(@Nonnull Object item) throws Exception {
      * The default implementation just emits the watermark downstream
      * to all ordinals.
      * <p>
+     * <i>Caution for jobs with at-least-once guarantee</i><br>
+     * In snapshotted jobs with <i>at least once</i> processing guarantee it
```
cangencer: we should be consistent with usage of "at-least-once" vs "at least once" and "exactly-once". I think the one with dash is the correct usage.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/AbstractProcessor.java
```diff
@@ -272,6 +272,14 @@ protected boolean tryProcess4(@Nonnull Object item) throws Exception {
...
+     * can happen that the watermarks break the monotonicity requirement when
+     * the job is restarted. This is caused by the fact that watermark, like any
+     * other stream item, can be delivered duplicately after restart. This
+     * means that after a restart, a processor can be asked to process a
+     * watermark older than it already processed.
+     * <p>
      * <strong>NOTE:</strong> unless the processor doesn't differentiate between
      * its inbound edges, the first choice should be leaving this method alone
      * and instead overriding the specific {@code tryProcessN()} methods for
```
cangencer: this should say "tryProcessWmN"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/Processors.java
```diff
@@ -403,6 +427,16 @@ private Processors() {
      * it deletes from storage all the frames that trail behind the emitted
      * windows. The type of emitted items is {@link TimestampedEntry
      * TimestampedEntry&lt;K, A>} so there is one item per key per window position.
+     * <p>
+     * <i>Behavior on job restart</i><br>
+     * This processor saves its state to snapshot. After restart, it can
+     * continue accumulating where it left off.
+     * <p>
+     * In at-least-once mode, the watermarks are allowed to resume at lower
```
cangencer: I think this sentence is a bit hard to understand: My attempt:

"In at-least-once mode, after a restart the emitted windows can contain duplicate events that were already processed before the last snapshot. Furthermore, at-least-once allows watermarks to go back in time after a restart. If such a watermark is received, some windows that were emitted before the snapshot will be re-emitted. These windows might contain less events than before as some of them had already been evicted before the snapshot."

viliam-durina:
verify

mtopolnik:
verify

viliam-durina:
verify

PR #549, opened 2017-10-19
JetClassloader fix

gurbuzali:
I've changed `findResource` method of `JetClassloader` so that it does not throw exception. Instead I've created a custom URL which can be used to load the underlying resource. I've also implemented `findResources` method too, which returns enumeration of URLs for the requested resource.

While creating execution plan of the job, `init` method of `ProcessorMetaSupplier` is called. In this `init` method, user may need to use job resources (ReadHdfsP for example), so we should set `JetClassloader` before creating execution plan.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
@@ -175,17 +175,21 @@ public class MasterContext {
         }
 
         MembersView membersView = getMembersView();
+        ClassLoader contextClassLoader = swapContextClassLoader(coordinationService.getClassLoader(jobId));
         try {
             logger.info("Start executing " + jobAndExecutionId(jobId, executionId) + ", status " + jobStatus()
                     + "\n" + dag);
             logger.fine("Building execution plan for " + jobAndExecutionId(jobId, executionId));
             JobConfig jobConfig = jobRecord.getConfig();
+
             executionPlanMap = ExecutionPlanBuilder.createExecutionPlans(nodeEngine,
                     membersView, dag, jobConfig, lastSnapshotId);
         } catch (TopologyChangedException e) {
             logger.severe("Execution plans could not be created for " + jobAndExecutionId(jobId, executionId), e);
             scheduleRestart();
             return;
+        } finally {
+            Thread.currentThread().setContextClassLoader(contextClassLoader);
         }
 
         logger.fine("Built execution plans for " + jobAndExecutionId(jobId, executionId));
@@ -580,6 +584,13 @@ public class MasterContext {
                  .invoke();
             futures.put(member, future);
         }
+    }
+
+    private static ClassLoader swapContextClassLoader(ClassLoader jobClassLoader) {
+        Thread currentThread = Thread.currentThread();
+        ClassLoader contextClassLoader = currentThread.getContextClassLoader();
+        currentThread.setContextClassLoader(jobClassLoader);
+        return contextClassLoader;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/JetClassLoader.java
```diff
@@ -79,7 +88,73 @@ private InputStream resourceStream(String name) {
...
+        public boolean hasMoreElements() {
+            return url != null;
+        }
+
+        @Override
+        public URL nextElement() {
+            if (this.url == null) {
+                throw new NoSuchElementException();
+            }
+            URL url = this.url;
```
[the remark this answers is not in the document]
gurbuzali: 👍

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -582,6 +587,15 @@ private void invokeOnParticipants(Map<MemberInfo, InternalCompletableFuture<Obje
         }
     }
 
+    private ClassLoader swapJobClassLoader(NodeEngine nodeEngine, long jobId) {
```
cangencer: I think it would be better if this took ClassLoader as parameter instead, and make it a static method with name `swapContextClassLoader`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/JetClassLoader.java
```diff
@@ -79,7 +88,75 @@ private InputStream resourceStream(String name) {
         return new InflaterInputStream(new ByteArrayInputStream(classData));
     }
 
+    private URL createResourceURL(String name) {
+        try {
```
cangencer: you can inline this method and use `Util.uncheckCall`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/JetClassLoader.java
```diff
@@ -79,7 +88,75 @@ private InputStream resourceStream(String name) {
...
+        }
+
+        @Override
+        public boolean hasMoreElements() {
+            return url != null;
+        }
+
+        @Override
+        public URL nextElement() {
+            if (this.url == null) {
```
cangencer: no need for `this`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/JetClassLoader.java
```diff
@@ -79,7 +88,75 @@ private InputStream resourceStream(String name) {
...
+        public void connect() throws IOException {
+        }
+
+        @Override
+        public InputStream getInputStream() throws IOException {
+            return jetClassLoader.resourceStream(url.getFile());
+        }
+    }
+
+    private static final class SingleItemEnumeration implements Enumeration<URL> {
```
cangencer: it should either be made generic or called `SingleURLEnumeration`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/JetClassLoader.java
```diff
@@ -79,7 +88,75 @@ private InputStream resourceStream(String name) {
...
+        } catch (MalformedURLException e) {
+            throw ExceptionUtil.rethrow(e);
+        }
+    }
+
     private static boolean isEmpty(String className) {
         return className == null || className.isEmpty();
     }
+
+    private static final class JobResourceURLStreamHandler extends URLStreamHandler {
```
cangencer: you can make this a non-static class and avoid the constructor

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/JetClassLoader.java
```diff
@@ -79,7 +88,75 @@ private InputStream resourceStream(String name) {
...
+            this.jetClassLoader = jetClassLoader;
+        }
+
+        @Override
+        protected URLConnection openConnection(URL url) throws IOException {
+            return new JobResourceURLConnection(url, jetClassLoader);
+        }
+    }
+
+    private static final class JobResourceURLConnection extends URLConnection {
```
cangencer: this can also be made into a non-static class to avoid passing the instance parameter

eminn:
👍

PR #560, opened 2017-10-24
Make traverseStream close the stream when exhausted

### LINKED ISSUE #482: Traversers.traverseStream does not close the stream
The contract of `Stream` is such that the stream is not automatically closed when exhausted. When the stream's source is a file or other system resource, it doesn't get released.

Proposed fix:

```java
/**
 * Returns a traverser over the given stream of non-null elements. It will 
 * traverse the stream through its spliterator, which it obtains 
 * immediately. When it exhausts the stream, it will close it.
 */
@Nonnull
public static <T> Traverser<T> traverseStream(@Nonnull Stream<T> stream) {
    return spliterate(stream.spliterator()).onFirstNull(stream::close);
}
```

mtopolnik:
Fixes #482


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Traversers.java
```diff
@@ -75,26 +72,27 @@ private Traversers() {
...
+    public static <T> Traverser<T> traverseEnumeration(@Nonnull Enumeration<T> enumeration) {
         return () -> enumeration.hasMoreElements() ? enumeration.nextElement() : null;
     }
 
     /**
-     * Returns a traverser over the given stream. The stream is traversed through
-     * its spliterator, which is obtained immediately.
+     * Returns a traverser over the given stream of non-null elements. It will
+     * traverse the stream through its spliterator, which it obtains
+     * immediately. When it exhausts the stream, it closes it.
```
viliam-durina: I'd make the non-null requirement more obvious, similar to wording in `traverseSpliterator`: 
"Returns a traverser over the given stream. (...) The stream must not emit {@code null} items."

Issue #565, opened 2017-10-26
When a Jet job is cancelled and progress made on hdfs is lost


eminn:
Since the `complete()` method is not called on the `WriteHdfsP` the tasks' output cannot be committed correctly and the files cannot be written to the hdfs.

PR #572, opened 2017-10-27
Implement PMS.complete method for WriteHdfsP to commit the job

gurbuzali:
fixes https://github.com/hazelcast/hazelcast-jet/issues/565

### DIFF
--- a/hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/impl/connector/hadoop/WriteHdfsP.java
+++ b/hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/impl/connector/hadoop/WriteHdfsP.java
@@ -37,7 +37,6 @@ import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.util.List;
 
-import static com.hazelcast.jet.impl.util.Util.uncheckCall;
 import static com.hazelcast.jet.impl.util.Util.uncheckRun;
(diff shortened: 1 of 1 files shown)


cangencer:
can we add a test for this?

gurbuzali:
I did write a test for that but it is not failing even without the fix. I guess it is because our tests are using `LocalFileSystem` not `DistributedFileSystem`.

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/impl/connector/hadoop/WriteHdfsP.java
```diff
@@ -161,18 +169,15 @@ public void init(@Nonnull Context context) {
 
         @Override
         public void complete(Throwable error) {
-            if (commitJob) {
-                uncheckRun(() -> outputCommitter.commitJob(jobContext));
+            if (processorList != null) {
+                processorList.forEach(Processor::complete);
```
cangencer: wouldn't this cause Processor.complete() to be called twice? Once after DAG finishes, and once on PS.complete()?

gurbuzali: Yes, I've missed to change it to a private `close` method, fixed now.

PR #574, opened 2017-10-31
Fix job execution initialization bug

metanet:
If a job participant leaves after the local member initializes the execution, the local member can clean up the ExecutionContext object just before it receives the ExecuteOperation from the coordinator. In this case, the local member should fail with TopologyChangedException instead of IllegalStateException.


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobExecutionService.java
```diff
@@ -200,8 +200,10 @@ private void verifyClusterInformation(long jobId, long executionId, Address coor
         ClusterServiceImpl clusterService = (ClusterServiceImpl) nodeEngine.getClusterService();
         MembershipManager membershipManager = clusterService.getMembershipManager();
         int localMemberListVersion = membershipManager.getMemberListVersion();
+        Address thisAddress = nodeEngine.getThisAddress();
+
         if (coordinatorMemberListVersion > localMemberListVersion) {
-            assert masterAddress != nodeEngine.getThisAddress();
+            assert masterAddress != thisAddress;
```
cangencer: shouldn't it be .equals() here? Also can we add an assertion msg?

PR #576, opened 2017-11-02
WIP: Read Sensitivity Modifiers via reflection

jerrinot:
Java SE 7 does not specify any standard modifiers. Some implementations
use custom modifier, but we should not have a hard dependency on them.

This changeset will use the HIGH-sensitivity modifier whenever possible,
yet it will fallback to a default mode when the custom modifier is unavailable.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamFilesP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamFilesP.java
@@ -20,6 +20,7 @@ import com.hazelcast.jet.JetException;
 import com.hazelcast.jet.core.AbstractProcessor;
 import com.hazelcast.jet.core.CloseableProcessorSupplier;
 import com.hazelcast.jet.core.ProcessorMetaSupplier;
+import com.hazelcast.jet.impl.util.ReflectionUtils;
 import com.hazelcast.logging.ILogger;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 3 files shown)


jerrinot:
I stumbled upon this while reading Jet code. I wonder whether you would accept a patch like this. if so then I'll improve it further.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/FileSystemUtils.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private static WatchEvent.Modifier[] getHighSensitivityModifiersIfAvailable() {
+        WatchEvent.Modifier modifier = ReflectionUtils.readStaticFieldOrNull(SENSITIVITY_MODIFIER_CLASSNAME, "HIGH");
+        if (modifier == null) {
+            //bad luck, we did not find the modifier
+            return new WatchEvent.Modifier[0];
+        }
+        WatchEvent.Modifier[] modifiers = new WatchEvent.Modifier[1];
+        modifiers[0] = modifier;
+        return modifiers;
```
viliam-durina: We can do just `return new WatchEvent.Modifier[] {modifier}`.
Also, I would inline the whole `FileSystemUtils` class into `StreamFilesP`, it's too specific for this processor.

jerrinot: done

PR #587, opened 2017-11-08
Update Javadoc for map and remoteMap source

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sources.java
```diff
@@ -137,9 +137,10 @@ private Sources() {
      * singleAttribute()} and {@link
      * com.hazelcast.projection.Projections#multiAttribute(String...)
      * multiAttribute()}), depending on how the data is stored in the map.
-     * Similarly, using the {@link com.hazelcast.query.PredicateBuilder
-     * PredicateBuilder} will build a predicate that makes use of any indexes
-     * created on the map.
+     * Similarly, {@link com.hazelcast.query.PredicateBuilder
+     * PredicateBuilder} or {@link com.hazelcast.query.Predicates} can be used
+     * to only apply the predicate to a single attribute without deserializing the
+     * whole object. Use of map indices is not currently supported.
```
gurbuzali: should we add a comment about `InMemoryFormat.Portable` ? User will be surprised if there is still de-serialization going on even if s/he used `om.hazelcast.query.Predicates`

cangencer: You mean `InMemoryFormat.BINARY` ?

gurbuzali: Yes, to avoid deserializing one should use `Portable` objects if map configured with `InMemoryFormat.BINARY` or `InMemoryFormat.NATIVE`

cangencer: I updated the section before to add a link to `InMemoryFormat`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sources.java
```diff
@@ -136,10 +136,11 @@ private Sources() {
      * com.hazelcast.projection.Projections#singleAttribute(String)
      * singleAttribute()} and {@link
      * com.hazelcast.projection.Projections#multiAttribute(String...)
-     * multiAttribute()}), depending on how the data is stored in the map.
-     * Similarly, using the {@link com.hazelcast.query.PredicateBuilder
-     * PredicateBuilder} will build a predicate that makes use of any indexes
-     * created on the map.
+     * multiAttribute()}), depending on the {@link com.hazelcast.config.InMemoryFormat}
```
mtopolnik: I'm not sure if this is a good way to describe it since `InMemoryFormat` is either `BINARY` or `OBJECT` when using OSS HZ. With `OBJECT` you always have the advantage (even with plain lambdas) and with `BINARY` you may not get it either way (if your object doesn't use portable serialization).

Issue #583, opened 2017-11-07
Common prefix for internal objects


neilstevenson:
Can we expose a common prefix for internal objects please -- eg. "__jet"

This is nearly done in https://github.com/hazelcast/hazelcast-jet/blob/master/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java

But it's an implementation class and there's no single base prefix

This would be useful when working with `HazelcastInstance.getDistributedObjects` to be able to
filter out objects with a single prefix

PR #594, opened 2017-11-14
Fix default config & test

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Jet.java
```diff
@@ -38,7 +39,7 @@
  */
 public final class Jet {
 
-    private static final String INTERNAL_JET_MAPS = "__jet.*";
+    static final String INTERNAL_JET_MAPS = "__jet.*";
```
cangencer: could we make the Jet internal prefix public as well? There is another issue around this (#583)

PR #611, opened 2017-11-23
Fix Kafka writer issues

### LINKED ISSUE #605: WriteKafkaP might not meet at least once guarantee
`KafkaProducer.send` is asynchronous. The Kafka sink might theoretically lose records when snapshot is made, job fails just after it and the async call is not complete yet.

viliam-durina:
KafkaProducer.send() call is async. This scenario was possible:
- send a record
- do a successful snapshot
- member fails before the async record is sent
- after restart, the record is lost and the at-least-once guarantee broken.

This was resolved that the processor ensures that all async calls are
done at the time of snapshot.

Another fixes:
- calling `producer.flush()` after each batch of items
- checking for errors from async calls

Fixes #605

### DIFF
--- a/hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/KafkaSinks.java
+++ b/hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/KafkaSinks.java
@@ -39,14 +39,17 @@ public final class KafkaSinks {
      * supplied mapping function.
      * <p>
      * The source creates a single {@code KafkaProducer} per cluster
-     * member using the supplied properties.
+     * member using the supplied {@code properties}.
      * <p>
-     * Behavior on job restart: the processor is stateless. If the job is
-     * restarted, duplicate events can occur. If you need exactly-once
-     * behavior, you must ensure idempotence on the application level.
+     * Behavior on job restart: the processor is stateless. On snapshot we only
+     * make sure that all async operations are done. If the job is restarted,
+     * duplicate events can occur. If you need exactly-once behavior, you must
+     * ensure idempotence on the application level.
      * <p>
-     * IO failures are generally handled by Kafka producer and generally do not
-     * cause the processor to fail. Refer to Kafka documentation for details.
+     * IO failures are generally handled by Kafka producer and do not cause the
+     * processor to fail. Refer to Kafka documentation for details.
+     * <p>
+     * The
      *
      * @param properties     producer properties which should contain broker
      *                       address and key/value serializers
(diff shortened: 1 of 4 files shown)


hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/WriteKafkaP.java
```diff
@@ -52,28 +73,56 @@ public boolean isCooperative() {
...
+        private static final long serialVersionUID = 1L;
+        private static final String MAX_BLOCK_MS = Long.toString(1L << 32);
 
         private final Properties properties;
         private final Function<? super T, ProducerRecord<K, V>> toRecordFn;
 
         private transient KafkaProducer<K, V> producer;
 
         public Supplier(Properties properties, Function<? super T, ProducerRecord<K, V>> toRecordFn) {
+            // use a large enough value: we don't handle the TimeoutException and we don't want to fail due to it.
```
gurbuzali: I think it would be good to add this information(that we override `max.block.ms`) to the JavaDoc

cangencer:
Would there be any way to verify exactly-once behaviour with a test?

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/KafkaSinks.java
```diff
@@ -39,14 +39,17 @@ private KafkaSinks() {
...
+     * make sure that all async operations are done. If the job is restarted,
+     * duplicate events can occur. If you need exactly-once behavior, you must
+     * ensure idempotence on the application level.
      * <p>
-     * IO failures are generally handled by Kafka producer and generally do not
-     * cause the processor to fail. Refer to Kafka documentation for details.
+     * IO failures are generally handled by Kafka producer and do not cause the
+     * processor to fail. Refer to Kafka documentation for details.
+     * <p>
+     * The
```
cangencer: seems leftover "The"

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/WriteKafkaP.java
```diff
@@ -36,10 +40,20 @@
...
-public final class WriteKafkaP<T, K, V> extends AbstractProcessor {
+public final class WriteKafkaP<T, K, V> implements Processor {
 
     private final KafkaProducer<K, V> producer;
     private final Function<T, ProducerRecord<K, V>> toRecordFn;
+    private final AtomicInteger numPendingAsyncCalls = new AtomicInteger();
+    private final AtomicReference<Throwable> lastError = new AtomicReference<>();
+
+    private final Callback callback = (metadata, exception) -> {
+        // Note: this method is called on different thread.
```
cangencer: should say it *may* be called on a different thread instead.

viliam-durina:
verify

PR #616, opened 2017-11-25
Predictable vertex name in pipeline jobs

viliam-durina:
Fixes #606
Fixes #614 (Rename source/sink vertex names)

PR #621, opened 2017-11-28
 Introduced EntryProcessor sink

eminn:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +63,82 @@ private Sinks() {
...
+     * will be wrapped inside an {@link EntryProcessor}.
+     *
+     *  @param mapName name of the map
+     *  @param updateFn update function contains the update logic.
+     */
+    public static <K, V, E extends Map.Entry<K, V>> Sink<E> updateMap(
+            String mapName,
+            DistributedBiFunction<Entry<K, V>, ? super V, ? extends V> updateFn
+    ) {
+        return fromProcessor("updateMap(" + mapName + ')', updateMapP(mapName, updateFn));
```
viliam-durina: Use `updateMapSink(xxx)` here, this was changed in PR #616 that I just merged

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -55,6 +59,90 @@
...
+                    IMap map = instance.getMap(name);
+                    return buffer -> {
+                        try {
+                            if (buffer.isEmpty()) {
+                                return;
+                            }
+                            for (Entry<Object, Object> entry : buffer.entrySet()) {
+                                EntryProcessor<K, V> entryProcessor = toEntryProcessorFn.apply((T) entry);
+                                K key = toKeyFn.apply((T) entry);
+                                map.executeOnKey(key, entryProcessor);
```
viliam-durina: This will be very slow, it's processing entries one by one synchronously, with 1ms network latency 500 keys/s at most... If we keep this version, performance penalty should be at least document. But I think we can adapt all use cases of this version to the other version, which works in batches, so there's no need to keep this version

eminn: I know it's performance drawbacks. Let me discuss about the requirements with Vladimir and Can whether to keep this version or not. I agree with your point that if we keep this, we need to make aware the users that it makes a remote call for each item in the stream.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
eminn: Will adopt the naming convention from there.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
jerrinot: IMap also has `    ICompletableFuture submitToKey(K key, EntryProcessor entryProcessor);` cant you use this?

eminn: It will improve the latency but doesn't change the fact that we are making a network call per item. I am not strongly opinionated about this, so we could use that. The main discussion should be whether we need to support user provided entryprocessors or force users to implement the logic in the update functions.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +63,82 @@ private Sinks() {
         return new SinkImpl<>(sinkName, metaSupplier);
     }
 
+    /**
+     * Returns a sink that updates {@code Map.Entry}s it receives into a Hazelcast
+     * {@code IMap} with the specified name and update function. The function
+     * will be wrapped inside an {@link EntryProcessor}.
```
cangencer: this should include some code similar to `Map.compute`, it's not clear which parameter goes where currently. Also not clear what happens when the key doesn't exist in the map.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +63,82 @@ private Sinks() {
...
+     *
+     *  @param mapName name of the map
+     *  @param updateFn update function contains the update logic.
+     */
+    public static <K, V, E extends Map.Entry<K, V>> Sink<E> updateMap(
```
cangencer: We tend to use nouns for sinks. This could be called `updatingMap` ? 

There is also the case where you just want to merge the value with the new one, where they are of the same type. This is equivalent to `Collectors.toMap(Function, Function, BinaryOperator)`. This might also be a useful variation of this?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,197 @@ private Sinks() {
         return new SinkImpl<>(sinkName, metaSupplier);
     }
 
+    /**
+     * Returns a sink that updates items it receives into a Hazelcast
```
mtopolnik: Try to use the active voice as much as possible. It's a style recommendation.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,197 @@ private Sinks() {
...
+     * {@code IMap} with the specified name and merge function. The function
+     * will be wrapped inside an {@link EntryProcessor}. The value extracted
+     * from input item will be inserted to the map if there is no value associated
```
mtopolnik: "from _the_ input item". "inserted _into_ the map"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,197 @@ private Sinks() {
...
+     * with the key, otherwise {@code mergeFn} will be called with the associated value
+     * as oldValue and the value which we want insert as newValue. The return value
```
mtopolnik: `{@code oldValue}` and `{@code newValue}`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,197 @@ private Sinks() {
...
+     * of the {@code mergeFn} will be the new value of the key.
+     *
+     *  @param mapName name of the map
+     *  @param toKeyFn  function that extracts key from the input item.
```
mtopolnik: "the key". No period at the end (Javadoc style guide)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,197 @@ private Sinks() {
...
+     *  @param toValueFn  function that extracts value from the input item.
+     *  @param mergeFn merge function contains the merge logic. The function
```
mtopolnik: "~merge~ function _that_ contains the merge logic"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,197 @@ private Sinks() {
...
+            @Nonnull DistributedFunction<E, K> toKeyFn,
+            @Nonnull DistributedFunction<E, V> toValueFn,
+            @Nonnull DistributedBinaryOperator<V> mergeFn
+    ) {
+        return fromProcessor("mergingMapSink(" + mapName + ')',
+                mergeMapP(mapName, toKeyFn, toValueFn,  mergeFn));
+    }
+
+    /**
+     * Returns a sink that updates items it receives into a Hazelcast
```
mtopolnik: _the_ items

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,197 @@ private Sinks() {
...
+     * {@code IMap} with the specified name and merge function in a
+     * remote cluster identified by the supplied {@code ClientConfig}.
+     * The value extracted from input item will be inserted to the map
```
mtopolnik: _the_ input item, inserted _into_ the map

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,197 @@ private Sinks() {
...
+     *  @param mapName name of the map
+     *  @param toKeyFn  function that extracts key from the input item.
+     *  @param toValueFn  function that extracts value from the input item.
+     *  @param mergeFn merge function contains the merge logic. The function
+     *                 will be called with {@code mergeFn.apply(V oldValue, V newValue)}
+     *  @param <E> type of input item
+     *  @param <K> type of the key
+     *  @param <V> type of the value
+     */
+    public static <E, K, V> Sink<E> mergingMap(
```
mtopolnik: The naming should be discussed. Maybe `mapWithMerging` and `mapWithUpdating`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,197 @@ private Sinks() {
         return new SinkImpl<>(sinkName, metaSupplier);
     }
 
+    /**
+     * Returns a sink that updates items it receives into a Hazelcast
+     * {@code IMap} with the specified name and merge function. The function
```
mtopolnik: Wording must be improved.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,197 @@ private Sinks() {
...
+
+    ) {
+        return fromProcessor("remoteUpdatingMapSink(" + mapName + ')',
+                updateRemoteMapP(mapName, clientConfig, toKeyFn, updateFn));
+    }
+
+    /**
+     * Returns a sink that updates items it receives into a Hazelcast
+     * {@code IMap} with the specified name using {@code EntryProcessor}.
+     *
```
mtopolnik: Paragraph break must be indicated with `<p>`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,204 @@ private Sinks() {
         return new SinkImpl<>(sinkName, metaSupplier);
     }
 
+    /**
+     * Returns a sink that updates the items it receives into a Hazelcast
+     * {@code IMap} with the specified name and merge function. The function
+     * will be wrapped inside an {@link EntryProcessor}. The value extracted
```
cangencer: I'm not sure if we need to mention EntryProcessor here, it's an implementation detail. It should be mentioned that updates happen in partition thread and will not be concurrent with other updates to the same key.

It's also not clear what happens if the function returns null. Is the entry removed? This is a good example where they document what happens as code: https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#merge-K-V-java.util.function.BiFunction-

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,204 @@ private Sinks() {
...
+     *                      cluster
+     *  @param toKeyFn  function that extracts key from the input item
+     *  @param toValueFn  function that extracts value from the input item
+     *  @param mergeFn  function contains the merge logic. The function
+     *                 will be called with {@code mergeFn.apply(V oldValue, V newValue)}
+     *  @param <E> type of input item
+     *  @param <K> type of the key
+     *  @param <V> type of the value
+     */
+    public static <E, K, V> Sink<E> remoteMergingMap(
```
cangencer: let's put a link here to mergingMap() instead to avoid duplicate content.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,204 @@ private Sinks() {
...
+                mergeRemoteMapP(mapName, clientConfig, toKeyFn, toValueFn, mergeFn));
+    }
+
+    /**
+     * Returns a sink that updates the items it receives into a Hazelcast
+     * {@code IMap} with the specified name and update function. The function
+     * will be wrapped inside an {@link EntryProcessor}. The {@code updateFn}
+     * will be called with the input item as item and the {@code oldValue}, which is
+     * the associated value of the extracted key in Hazelcast Map. The return
+     * value of the {@code updateFn} will be the new value of the key.
```
cangencer: see my comments about mergingMap. It's not 100% clear how the update logic works.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +64,204 @@ private Sinks() {
...
+     * <p>
+     * This method internally uses {@code IMap.submitToKey() } method to
+     * invoke provided {@code EntryProcessor} on the key which does not
+     * utilize batching. There will be an invocation for each key.
+     * Main purpose of this method is to support {@code Offloadable}
+     * entry processors. If you don't have an {@code Offloadable}
+     * entry processor please use one of the following to implement
+     * your update/merge logic. The following methods will utilize
+     * batching while interacting with Hazelcast Cluster which is
+     * more efficient.
```
cangencer: I rewrote this as follows:

```
Returns a sink that updates the items it receives into a Hazelcast
{@code IMap} with the specified name using an {@code EntryProcessor}
per item.
<p>
Internally, the sink uses an entry processor per key is submitted. This method is not
as efficient as {@see Sinks#updatingMap} or {@see Sinks#mergingMap}
because it submits one {@link EntryProcessor} per key rather than in batches.
<p>
The supplied EntryProcessor can be made {@link Offloadable} to avoid
blocking the partition thread during large update operations.
```

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -55,6 +60,149 @@
...
+                    IMap map = instance.getMap(name);
+                    return buffer -> {
+                        try {
+                            if (buffer.isEmpty()) {
+                                return;
+                            }
+                            for (Entry<Object, Object> entry : buffer.entrySet()) {
+                                EntryProcessor<K, V> entryProcessor = toEntryProcessorFn.apply((T) entry);
+                                K key = toKeyFn.apply((T) entry);
+                                map.submitToKey(key, entryProcessor);
```
cangencer: what happens to the result of the future? It doesn't look like it gets handled at all. It will also not work with snapshotting and processing guarantees. It should behave in a similar way to `WriteKafkaP`.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/HazelcastConnectorTest.java
```diff
@@ -262,4 +270,82 @@ public void when_readList_and_writeList() throws ExecutionException, Interrupted
         assertEquals(ENTRY_COUNT, jetInstance.getList(sinkName).size());
     }
 
+    @Test
+    public void when_readMap_and_updateMap_with_update_function() throws ExecutionException, InterruptedException {
```
cangencer: these tests seem redundant with their equivalents in Pipeline API. They should all use Pipeline API and preferable be in the same place.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -55,6 +60,149 @@
...
+    ) {
+        boolean isLocal = clientConfig == null;
+        return dontParallelize(new HazelcastWriterSupplier<>(
+                serializableConfig(clientConfig),
+                index -> new ArrayList<>(),
+                ArrayList::add,
+                instance -> {
+                    IMap map = instance.getMap(name);
+                    Map<K, T> tmpMap = new HashMap<>();
+                    AbstractEntryProcessor entryProcessor = new AbstractEntryProcessor() {
```
cangencer: could this not be generic here?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
cangencer: Having them all prefixed with `map` seems like a good idea so that they're more discoverable.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -55,6 +60,149 @@
...
+    ) {
+        boolean isLocal = clientConfig == null;
+        return dontParallelize(new HazelcastWriterSupplier<>(
+                serializableConfig(clientConfig),
+                index -> new ArrayList<>(),
+                ArrayList::add,
+                instance -> {
+                    IMap map = instance.getMap(name);
+                    Map<K, V> tmpMap = new HashMap<>();
+                    AbstractEntryProcessor entryProcessor = new AbstractEntryProcessor() {
```
cangencer: same, could be made generic

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -162,7 +301,8 @@ private static SerializableClientConfig serializableConfig(ClientConfig clientCo
             entries = new ArrayList<>();
         }
 
-        @Override @Nonnull
+        @Override
```
cangencer: you don't need to reformat these lines, you can configure intellij to put them on same line.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -189,6 +330,146 @@ public String toString() {
...
+        }
+
+        @Override
+        public boolean tryProcess() {
+            checkError();
+            return true;
+        }
+
+        @Override
+        protected boolean tryProcess0(@Nonnull Object object) throws Exception {
```
cangencer: this should implement `tryProcess(int ordinal,)`, not `tryProcess0`, sinks can have multiple inputs.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -238,7 +519,8 @@ private boolean isRemote() {
             return clientConfig != null;
         }
 
-        @Override @Nonnull
+        @Override
+        @Nonnull
```
cangencer: same about reformatting.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -97,6 +99,20 @@ public static void uncheckRun(@Nonnull RunnableExc r) {
         }
     }
 
+    public static <T> ExecutionCallback<T> callbackOf(Consumer<T> onResponse, Consumer<Throwable> onError) {
```
cangencer: can we write a brief javadoc here explaining what it does?

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/util/IncrementEntryProcessor.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.jet.impl.util;
+
+import com.hazelcast.map.AbstractEntryProcessor;
+import java.util.Map.Entry;
+
+public class IncrementEntryProcessor<K> extends AbstractEntryProcessor<K, Integer> {
```
cangencer: this can be made a private class I think, it's not used anywhere else.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -189,6 +328,145 @@ public String toString() {
         }
     }
 
+    public static class EntryProcessorWriter<T, K, V> extends AbstractProcessor {
```
viliam-durina: This class can be private.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -189,6 +328,145 @@ public String toString() {
...
+        @Override
+        public boolean tryProcess() {
+            checkError();
+            return true;
+        }
+
+        @Override
+        protected boolean tryProcess(int ordinal, @Nonnull Object object) throws Exception {
+            checkError();
+            if (numConcurrentOps.incrementAndGet() >= MAX_PARALLEL_ASYNC_OPS) {
```
viliam-durina: There's no matching decrement to this increment if you exceed the limit. You should do it similarly to [this code](https://github.com/hazelcast/hazelcast-jet/blob/master/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/AsyncMapWriter.java#L193-L204) and just increase by 1. Maybe also extract this method to `Util`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -55,6 +67,133 @@
     private HazelcastWriters() {
     }
 
+    @Nonnull
+    @SuppressWarnings("unchecked")
+    public static <T, K, V> ProcessorMetaSupplier mergeMapP(
```
viliam-durina: This method and the next `updateMapP` are exactly the same, except for the `entryProcessor`. It should be extracted out.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -189,6 +328,145 @@ public String toString() {
...
+        }
+
+        @Override
+        public boolean saveToSnapshot() {
+            return ensureAllWritten();
+        }
+
+        private boolean ensureAllWritten() {
+            checkError();
+            return numConcurrentOps.get() == 0;
```
viliam-durina: This two lines should be swapped: it can happen, that there are no errors, then last operation completes with error and the counter is 0 -> we'll miss the error.
Fix this also in `WriteKafkaP`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -58,6 +65,165 @@ private Sinks() {
...
+    ) {
+        return fromProcessor("mapWithMerging(" + mapName + ')',
+                mergeMapP(mapName, toKeyFn, toValueFn,  mergeFn));
+    }
+
+    /**
+     * Returns a sink equivalent to {@link #mapWithMerging}, but for a map
+     * in a remote Hazelcast cluster identified by the supplied {@code
+     * ClientConfig}.
+     */
```
viliam-durina: Document the possibility to configure `INVOCATION_TIMEOUT_SECONDS`.

cangencer: I think this is not needed here as this goes for all remote sinks. We could refer to it in the ref manual in the Hazelcast connectors section.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -57,7 +71,97 @@ private HazelcastWriters() {
...
+                    AbstractEntryProcessor<K, V> entryProcessor = new AbstractEntryProcessor<K, V>() {
+                        @Override
+                        public Object process(Entry<K, V> entry) {
+                            V oldValue = entry.getValue();
+                            T item = tmpMap.get(entry.getKey());
+                            V newValue = updateFn.apply(oldValue, item);
+                            entry.setValue(newValue);
+                            return null;
+                        }
+                    };
```
viliam-durina: You've moved this part to the inner lambda. Now, `tmpMap` and `entryProcessor` will be created every time the buffer is flushed. Before, they were recycled.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -328,7 +293,7 @@ public String toString() {
         }
     }
 
-    public static class EntryProcessorWriter<T, K, V> extends AbstractProcessor {
+    static final class EntryProcessorWriter<T, K, V> extends AbstractProcessor {
```
viliam-durina: This class can be private if you move the `EntryProcessorWriterSupplier` class nested in it out. Or when you remove the import and reference the `EntryProcessorWriterSupplier` as `EntryProcessorWriter.EntryProcessorWriterSupplier`.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/SinksTest.java
```diff
@@ -120,7 +120,23 @@ public void mergingMap_with_merging_function() {
     }
 
     @Test
-    public void remoteMergingMap_with_merging_function() {
+    public void mapWithMerging_key_is_removed_when_function_returns_null() {
```
viliam-durina: `mapWithMerging_when_functionReturnsNull_then_keyIsRemoved`

viliam-durina: And few more like this.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
viliam-durina: I meant documenting it at all methods. I personally never go to refman for such details. In refman I'm mostly looking for concepts and introductions, the rest should be in javadoc, which I read as necessary.

PR #628, opened 2017-11-29
Send join request for job only when Job.getFuture() is called

### LINKED ISSUE #602: JetInstance.getJobs() should not create pending invocations for each running Job
Currently when you call `JetInstance.getJobs()` it creates pending invocations for each job because it sends a `JoinJobOperation` per job.

cangencer:
Fixes #602


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJobProxy.java
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        private boolean isRestartable(Throwable t) {
+            return t instanceof MemberLeftException
+                    || t instanceof TargetDisconnectedException
+                    || t instanceof TargetNotMemberException;
+        }
+
+        private boolean isSplitBrainMerge(Throwable t) {
+            Throwable cause = peel(t);
```
viliam-durina: duplicate `peel`

PR #648, opened 2017-12-12
Fix Javadoc for mapWithEntryProcessor

mtopolnik:
The current doc erroneously states that the other map-updating sinks, `mapWithMerging` and `mapWithUpdating`, don't achieve data locality.

The PR also reorders methods in `Sinks`, putting the simple `map()` sink above the more advanced ones.


mtopolnik:
I'll merge this as a quick fix now. We plan to revisit the design of this API and perhaps eliminate `mapWithEntryProcessor`.

PR #652, opened 2017-12-13
Include Hazelcast Javadocs in Jet distribution

cangencer:


pom.xml
```diff
@@ -335,7 +335,7 @@
                     <maxmemory>1024</maxmemory>
                     <additionalparam>-Xdoclint:-missing -tag apiNote::a -tag implSpec:a -link https://docs.oracle.com/javase/8/docs/api/</additionalparam>
                     <excludePackageNames>
-                        *.impl:
+                        *.impl:*.internal
```
mtopolnik: `impl` and `internal` could appear in the middle of the package name.

PR #654, opened 2017-12-18
Added note about entry locking

eminn:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -270,6 +276,9 @@ private Sinks() {
      * avoid blocking the Hazelcast partition thread during large update
      * operations.
      *
+     * <b>Note:</b> This operation is lock-aware, if the key is locked,
+     * the EntryProcessor will wait until it acquires the lock.
+     *
```
viliam-durina: I'd say:
> Unlike {@link #mapWithUpdating} and {@link #mapWithMerging}, this operation <em>is</em> lock-aware.

Also in the other two. It's easy to miss the point; I at first moment thought the note is the same in all three instances and was about to comment about it. And it's very unexpected difference, it's surprising that so similar methods have this kind of difference.

We also should add a simple test: lock one entry and check that it is/is not processed. When doing changes it's easy to forget about this.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -115,6 +115,9 @@ private Sinks() {
      *     map.put(key, value);
      * </pre>
      *
+     * <b>Note:</b> This operation is <em>NOT</em> lock-aware, it will process the entries
+     * no matter if they are locked or not.
```
viliam-durina: Add here and to `mapWithMerging`:
> Use {@link #mapWithEntryProcessor} if you need locking.

cangencer:
Please label and add milestone to your PRs.

PR #657, opened 2017-12-20
[WIP] Kotlin Processors

mtopolnik:
Kotlin coroutines bring a dramatic simplification to the coding of cooperative processors. The magic is in a single method: `yield()`, which suspends the execution such that it can be resumed later.

In Java we need a ton of `Traverser`-based support code to achieve a mediocre level of handling the cooperative multithreading concern. The processor code must express all its computation declaratively, as higher-order function calls on `Traverser`.  In Kotlin all this turns into plain imperative idioms with the usual flow-control constructs (`if`, `while` etc.). A good example of this is `InsertWatermarksPK`.

Source processors must handle the concern of suspendable resource management on their own, remembering which resources are held open while the processor is suspended (returns from a `process()` call). In Kotlin this turns into a simple scoped resource management idiom (suspendable try-with-resources block). This difference can be observed in `ReadFilesP(K)` and `StreamFilesP(K)`.

Performance of Kotlin processors is identical to Java's. Measuring on single-member WordCount (unrealistic case that emphasizes within-processor overhead) there was no drop in performance.

The Jet JAR increases in size from the current 10 MB to 12 MB due to the Kotlin runtime dependency. There were no issues with shading it in.


gregrluck:
Could this be done as an Adaptor? KotlinProcessor or KotlinAdapter which lets you implement a processor using Kotlin.

We could package this as a module and advertise this on the website.

gregrluck:
See proposal to create a Kotlin Support module.

PR #666, opened 2018-01-05
Improve job lifecycle management

metanet:
* Cancel jobs by sending cancellation operations instead of relying on the operation cancellation api.
* Job.join() returns only when the job is completed on the cluster.
* Add COMPLETING status. During the job is being completed, its status is reported as COMPLETING.
* Name jobs
* Simplify method names in the Job interface
* Change the job submission phase. Once the job resources are uploaded, the job record is created first. Then, the caller joins to the job. This approach eliminates the race between submitting the job and querying it afterwards.


metanet:
run-lab-run

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetInstanceImpl.java
```diff
@@ -67,10 +69,25 @@ public Job newJob(@Nonnull DAG dag, @Nonnull JobConfig config) {
...
+    public Job getJob(long jobId) {
+        try {
+            Job job = new JobProxy((NodeEngineImpl) nodeEngine, jobId);
+            job.getStatus();
+            return job;
+        } catch (Exception e) {
+            if (peel(e) instanceof JobNotFoundException) {
+                return null;
+            }
+
```
cangencer: extra line break

cangencer: extra line break

metanet:
run-lab-run

metanet:
run-lab-run

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetClientInstanceImpl.java
```diff
@@ -58,24 +61,56 @@ public JetConfig getConfig() {
...
-    public Collection<Job> getJobs() {
+    public List<Job> getJobs() {
         ClientInvocation invocation = new ClientInvocation(
                 client, JetGetJobIdsCodec.encodeRequest(), null, masterAddress(client.getCluster())
         );
         return uncheckCall(() -> {
             ClientMessage response = invocation.invoke().get();
             Set<Long> jobs = serializationService.toObject(JetGetJobIdsCodec.decodeResponse(response).response);
-            return jobs.stream().map(jobId -> new ClientJobProxy(client, jobId)).collect(toList());
+            return jobs.stream().map(this::getJob).collect(toList());
```
cangencer: this would cause a RPC for every job, which is not probably what we want here.

cangencer:
verify

metanet:
@cangencer @viliam-durina thanks guys for your review efforts

lukasherman:
Great improvement, I will test it and provide feedback soon.

PR #673, opened 2018-01-12
Fix & document projection to null

### LINKED ISSUE #669: No results are written into sink when projected field is nullable
Consider the following example, Hazelcast Member:
```
        Config config = new Config();
        config.getSerializationConfig().addPortableFactory(1, new PortableFactoryImpl());
        HazelcastInstance hazelcastInstance = Hazelcast.newHazelcastInstance(config);

        IMap<String, User> mapTest1 = hazelcastInstance.getMap("test1");
        mapTest1.put("k0", new User("k0", "n0", LocalDate.now()));
        mapTest1.put("k1", new User("k1", "n1", LocalDate.now()));
        mapTest1.put("k2", new User("k2", null, LocalDate.now()));
        mapTest1.put("k3", new User("k3", null, LocalDate.now()));
        mapTest1.put("k4", new User("k4", "n4", LocalDate.now()));
        mapTest1.put("k5", new User("k5", "n5", LocalDate.now()));

        System.out.println("Users count: " + mapTest1.size());
```
and Jet Member:
```
        JetInstance jet = Jet.newJetInstance();
        Pipeline p1 = Pipeline.create();

        p1.drawFrom(Sources.<String, ImdgMember.User, String>remoteMap("test1", new ClientConfig(),
// Workaround is to filter out NULLs
//                Predicates.and(Predicates.in("id", "k1", "k2", "k3", "k4"), Predicates.not(Predicates.equal("name", null))),
                Predicates.in("id", "k1", "k2", "k3", "k4"),
                Projections.singleAttribute("name"))).drainTo(Sinks.list("test1"));

        jet.newJob(p1).join();

        IList<String> sink = jet.getList("test1");
        System.out.println("Fetched (count=" + sink.size() + "): " + sink.subList(0, sink.size()));
```

No items are written to the sink List when some of the matching entries have `null` as the value of field `name`.

viliam-durina:
Links in javadoc will work when hazelcast/hazelcast#12142 is merged and
pom.xml references imdg 3.10. Fixes #669.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sources.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sources.java
@@ -136,7 +136,7 @@ public final class Sources {
      *      if you want to use projection only
      * @param projection the projection to map the events, you may use
      *     {@link Projections#identity()} if you want just the predicate.
-     *     If the projection returns a {@code null} for some items, that items
+     *     If the projection returns a {@code null} for an item, that item
      *     will be filtered out.
      * @param <T> type of emitted item
      */
@@ -187,7 +187,7 @@ public final class Sources {
      *      ADDED} and {@link com.hazelcast.core.EntryEventType#UPDATED UPDATED} events
      * @param projectionFn the projection to map the events, you may use
      *     {@link Util#mapEventToEntry()} to project new value from the event
-     *     If the projection returns a {@code null} for some items, that items
+     *     If the projection returns a {@code null} for an item, that item
      *     will be filtered out.
      * @param initialPos describes which event to start receiving from
      * @param <T> type of emitted item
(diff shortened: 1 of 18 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sources.java
```diff
@@ -176,6 +187,8 @@ private Sources() {
      *      ADDED} and {@link com.hazelcast.core.EntryEventType#UPDATED UPDATED} events
      * @param projectionFn the projection to map the events, you may use
      *     {@link Util#mapEventToEntry()} to project new value from the event
+     *     If the projection returns a {@code null} for some items, that items
```
mtopolnik: Should be "If the projection returns {@code null} for an item, that item will be filtered out. There are some more instances of this text below.

PR #683, opened 2018-01-25
Do not restart auto-restart disabled job if the master node leaves

### LINKED ISSUE #682: `AutoRestartOnMemberFailure` is false and the job still restarts
1. start 2 members
2. submit a job with `JobConfig.setAutoRestartOnMemberFailure(false)`
3. shut down one of the members
4. the job restarts

metanet:
Fixes #682

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
@@ -25,6 +25,7 @@ import com.hazelcast.jet.config.JetConfig;
 import com.hazelcast.jet.config.JobConfig;
 import com.hazelcast.jet.core.JobNotFoundException;
 import com.hazelcast.jet.core.JobStatus;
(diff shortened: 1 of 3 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -248,6 +249,17 @@ private boolean completeMasterContextIfJobAlreadyCompleted(MasterContext masterC
             return masterContexts.remove(jobId, masterContext);
         }
 
+        if (!masterContext.getJobConfig().isAutoRestartOnMemberFailureEnabled()
+                && jobRepository.getExecutionIdCount(jobId) > 0) {
+            String coordinator = nodeEngine.getNode().getThisUuid();
+            Throwable result = new TopologyChangedException();
+            logger.info("Completing master context " + idToString(jobId) + " with " + result
+                    + " since it is auto-restart disabled and the job has been executed before");
```
viliam-durina: "since it is auto-restart disabled" -> "since auto-restart is disabled"

PR #685, opened 2018-01-29
Modify allOf to use TupleN or ItemsByTag

viliam-durina:
Before, `AggregateOperations.allOf` wrapped the result to a 
`List<Object>`, which is type-unsafe. This commit replaces it with 
a special versions that uses either `Tuple2`, `Tuple3` or `ItemsByTag` 
in the general case.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
@@ -32,6 +32,7 @@ import com.hazelcast.jet.function.DistributedFunction;
 import com.hazelcast.jet.function.DistributedSupplier;
 import com.hazelcast.jet.function.DistributedToDoubleFunction;
 import com.hazelcast.jet.function.DistributedToLongFunction;
+import com.hazelcast.jet.function.DistributedTriFunction;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -237,6 +238,10 @@ public final class AggregateOperations {
                 .andFinish(LinTrendAccumulator::finish);
     }
 
+    /**
+     * Convenience for {@link #allOf(AggregateOperation1, AggregateOperation1,
+     * DistributedBiFunction)} with identity finish.
+     */
     @Nonnull
     public static <T, A1, A2, R1, R2> AggregateOperation1<T, Tuple2<A1, A2>, Tuple2<R1, R2>> allOf(
             @Nonnull AggregateOperation1<? super T, A1, R1> op1,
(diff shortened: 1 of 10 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
```diff
@@ -232,55 +237,73 @@ private AggregateOperations() {
...
     ) {
-        AggregateOperation1[] untypedOps = operations;
+        return allOf(op1, op2, Tuple2::tuple2);
+    }
 
+    @Nonnull
+    public static <T, A1, A2, R1, R2, R> AggregateOperation1<T, Tuple2<A1, A2>, R> allOf(
+            @Nonnull AggregateOperation1<? super T, A1, R1> op1,
+            @Nonnull AggregateOperation1<? super T, A2, R2> op2,
+                @Nonnull DistributedBiFunction<? super R1, ? super R2, R> finisher
```
cangencer: indentation

hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AllOfAggregationBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+    AllOfAggregationBuilder() { }
+
+    public <R> Tag<R> add(AggregateOperation1<? super T, ?, R> operation) {
+        operations.add(operation);
+        Tag<R> tag = tag(tags.size());
+        tags.add(tag);
+        return tag;
+    }
+
+    public AggregateOperation1<T, Object[], ItemsByTag> build() {
```
cangencer: we could also add an optional finisher function here, something like `Function<ItemsByTag, R>`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
```diff
@@ -233,54 +239,159 @@ private AggregateOperations() {
...
+        return allOf(op1, op2, Tuple2::tuple2);
+    }
 
+    /**
+     * Returns composite aggregate operation from 2 other aggregate operations.
+     * It allows you to calculate multiple aggregations over the same items at once.
+     *
+     * @param op1 1st operation
+     * @param op2 2nd operation
+     * @param finishFn a function combining 3 results into single target instance
```
cangencer: should say 2 results

hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
```diff
@@ -233,54 +239,159 @@ private AggregateOperations() {
...
+                            op3.deductFn().accept(acc1.f2(), acc2.f2());
+                        })
+                .andFinish(acc -> finishFn.apply(
+                        op1.finishFn().apply(acc.f0()),
+                        op2.finishFn().apply(acc.f1()),
+                        op3.finishFn().apply(acc.f2())));
+    }
+
+    /**
+     * Returns a builder to create a composite aggregate operation, if the
```
cangencer: it might be better to just say this can be used with an arbitrary number of operations and then add a note to refer to one of the two or three-arity ones. If you have less than 4, you can still use this option (for whatever reason).

hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
viliam-durina: fixed

hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
viliam-durina: fixed

PR #687, opened 2018-01-29
Improve javadoc on statelessness, parallelism

### LINKED ISSUE #678: JavaDoc should clearly indicate which methods assume stateless functions
Example: `ComputeStage` has a method `filter()` which accept a function. This function is expected to be stateless. JavaDoc should explicitly say this.

viliam-durina:
Fixes #678

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
@@ -52,8 +52,8 @@ import static java.nio.charset.StandardCharsets.UTF_8;
  * as its transform has the type {@link SinkStage} and accepts no
  * downstream stages.
  * <p>
- * The default local parallelism for sources in this class is 1, check the
- * documentation of individual methods.
+ * The default local parallelism for sources in this class is typically 1,
+ * check the documentation of individual methods.
  */
 public final class Sinks {
 
@@ -64,7 +64,7 @@ public final class Sinks {
      * Returns a sink constructed directly from the given Core API processor
      * meta-supplier.
      * <p>
-     * Default local parallelism for this source is specified by the given
+     * The default local parallelism for this source is specified by the given
      * {@link ProcessorMetaSupplier#preferredLocalParallelism() metaSupplier}.
      *
      * @param sinkName user-friendly sink name
@@ -84,7 +84,7 @@ public final class Sinks {
      * but overwritten. After the job is restarted from snapshot, duplicate
      * items will not change the state in the target map.
      * <p>
-     * Default local parallelism for this processor is 1.
+     * The default local parallelism for this processor is 1.
      */
     @Nonnull
     public static <E extends Map.Entry> Sink<E> map(@Nonnull String mapName) {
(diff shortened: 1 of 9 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -60,6 +63,9 @@ private Sinks() {
     /**
      * Returns a sink constructed directly from the given Core API processor
      * meta-supplier.
+     * <p>
+     * Default local parallelism for this source is specified by the given
```
mtopolnik: Should be "The default..."

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
```diff
@@ -51,6 +51,9 @@
  * a sink is a transform that has no output. A pipeline stage with a sink
  * as its transform has the type {@link SinkStage} and accepts no
  * downstream stages.
+ * <p>
+ * The default local parallelism for sources in this class is 1, check the
```
mtopolnik: Is this necessary? Since there's documentation on each method. One of the methods actually has a different default parallelism. Maybe say "The default local parallelism for _most_ of the sources in this class is 1".

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sources.java
```diff
@@ -388,6 +409,9 @@ private Sources() {
      * If the {@code ICache} is modified while being read, or if there is a
      * cluster topology change (triggering data migration), the source may
      * miss and/or duplicate some entries.
+     * <p>
+     * Default local parallelism for this processor is 2 (or less if less CPUs
```
mtopolnik: "or less if less CPUs are available" sounds contrived given that there is only one choice. "Or 1 if there's just one CPU available" sounds more natural.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sources.java
```diff
@@ -73,6 +76,9 @@ private Sources() {
     /**
      * Returns a source constructed directly from the given Core API processor
      * meta-supplier.
+     * <p>
+     * Default local parallelism for this source is specified by the given
```
mtopolnik: The default...

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
viliam-durina: I think it's good to mention common distinctive feature of the processors here. I changed to "is typically 1".

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sinks.java
viliam-durina: fixed

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sources.java
viliam-durina: fixed

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Sources.java
viliam-durina: fixed

PR #689, opened 2018-01-30
Replace findbugs with spotbugs

cangencer:


cangencer:
verify

viliam-durina:
Can we now get rid of `Util.completeVoidFuture`?

cangencer:
verify

cangencer:
verify

PR #691, opened 2018-01-30
Fix cooperative emission in StreamFilesP

### LINKED ISSUE #658: StreamFilesP should retry emitting line if outbox full
Currently StreamFilesP doesn't retry emitting the current line if outbox is full.

viliam-durina:
Other fixes:

- Fix the case when EOF is encountered mid-line and the reading of the
line resumes later

- The processor tried to finish watcher events after it was closed,
which is nonsense. It's only closed when the job is cancelled or the
watched directory was deleted. In neither case we should continue the
work.

- Updated the test to check cooperative emission.

- Fix TestProcessorContext.getLogger: the logger was created in
constructor even though the vertexName or globalProcessorIndex could be
changed later

Fixes #658

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamFilesP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamFilesP.java
@@ -140,7 +140,7 @@ public class StreamFilesP extends AbstractProcessor implements Closeable {
             if (isClosed()) {
                 return;
             }
-            getLogger().info("Closing StreamFilesP");
+            getLogger().fine("Closing StreamFilesP");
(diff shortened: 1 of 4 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamFilesP.java
```diff
@@ -131,7 +140,7 @@ public void close() {
             if (isClosed()) {
                 return;
             }
-            getLogger().info("Closing StreamFilesP. Any pending watch events will be processed.");
+            getLogger().info("Closing StreamFilesP");
```
cangencer: should this be on info level?

PR #693, opened 2018-01-31
Introduce manual job restart

metanet:
A new method is added to the Job interface. If the user calls job.restart() while a job is running, the job is restarted on the current member list of the cluster from the last successful snapshot. The new method can be used as a non-graceful way of achieving elasticity.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
@@ -95,4 +95,14 @@ public interface Job {
         Util.uncheckRun(() -> getFuture().get());
     }
 
+    /**
+     * Cancels the current execution if the job is currently running and
+     * schedules a new execution with the current member list of the Jet cluster
+     *
+     * @throws IllegalStateException if the job has been already completed
+     *
+     * @return true if the current execution of the job is cancelled
+     */
+    boolean restart();
+
 }
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
@@ -29,6 +29,7 @@ import com.hazelcast.jet.core.JobStatus;
 import com.hazelcast.jet.core.TopologyChangedException;
 import com.hazelcast.jet.core.Vertex;
 import com.hazelcast.jet.function.DistributedFunction;
+import com.hazelcast.jet.impl.exception.JobRestartRequestedException;
 import com.hazelcast.jet.impl.execution.BroadcastEntry;
 import com.hazelcast.jet.impl.execution.init.ExecutionPlan;
 import com.hazelcast.jet.impl.operation.CancelExecutionOperation;
@@ -38,6 +39,7 @@ import com.hazelcast.jet.impl.operation.SnapshotOperation;
 import com.hazelcast.jet.impl.operation.StartExecutionOperation;
 import com.hazelcast.jet.impl.util.ExceptionUtil;
 import com.hazelcast.jet.impl.util.NonCompletableFuture;
+import com.hazelcast.jet.impl.util.CompletionToken;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.nio.Address;
 import com.hazelcast.spi.ExecutionService;
@@ -99,7 +101,7 @@ public class MasterContext {
     private final JobRecord jobRecord;
     private final long jobId;
     private final NonCompletableFuture completionFuture = new NonCompletableFuture();
-    private final CompletableFuture<Void> cancellationFuture = new CompletableFuture<>();
+    private final CompletionToken cancellationToken;
     private final AtomicReference<JobStatus> jobStatus = new AtomicReference<>(NOT_STARTED);
     private final SnapshotRepository snapshotRepository;
     private volatile Set<Vertex> vertices;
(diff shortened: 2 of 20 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -412,6 +422,19 @@ private void cancelExecute(long jobId, long executionId) {
         });
     }
 
+    /**
+     * Cancels the execution invocation in order to restart it afterwards if the job is currently being executed
+     */
+    boolean cancelCurrentExecution() {
+        ExecutionCallback<Object> callback = this.executionCancellationCallback;
+        if (callback != null) {
+            callback.onFailure(null);
```
cangencer: we should add a comment here. Do you think having an explict exception would be better? I.e. JobRestartRequestedException ? Also naming is still a bit ambiguous between `cancelExecute` and cancelCurrentExecution`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -398,7 +398,17 @@ public void onFailure(Throwable t) {
         }));
 
         Function<ExecutionPlan, Operation> operationCtor = plan -> new StartExecutionOperation(jobId, executionId);
-        invoke(operationCtor, this::onExecuteStepCompleted, callback);
+        Consumer<Map<MemberInfo, Object>> completionCallback = results -> {
+            executionCancellationCallback = null;
+            onExecuteStepCompleted(results);
```
cangencer: I can see that the job is only restarted when it's TopologyChangedException and auto-restart is enabled, otherwise a restart will not be scheduled.

cangencer:
verify

metanet:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -372,46 +375,55 @@ private Throwable getInitResult(Map<MemberInfo, Object> responses) {
...
             Function<ExecutionPlan, Operation> operationCtor = plan -> new CancelExecutionOperation(jobId, executionId);
             invoke(operationCtor, responses -> { }, null);
         });
     }
 
+    /**
+     * Cancels the execution invocation in order to restart it afterwards if the job is currently being executed
+     */
+    boolean cancelCurrentExecution() {
+        CompletionToken restartFuture = this.executionRestartToken;
```
cangencer: name still says "future"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/CompletionToken.java
```diff
@@ -0,0 +1,43 @@
...
+
+package com.hazelcast.jet.impl.util;
+
+import java.util.concurrent.CompletableFuture;
+
+import static com.hazelcast.jet.impl.util.Util.completeVoidFuture;
+
+/**
+ * Wraps a CompletableFuture that is expected to be completed
+ * only in a single way and provides methods for easy of use
```
cangencer: typo "ease"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/CompletionToken.java
```diff
@@ -0,0 +1,43 @@
...
+    public boolean complete() {
+        return completeVoidFuture(future);
+    }
+
+    public boolean isCompleted() {
+        return future.isDone();
+    }
+
+    public void whenCompleted(Runnable runnable) {
+        future.whenComplete((result, throwable) -> runnable.run());
```
cangencer: it should use `withTryCatch` here, otherwise any exceptions in callback are lost and won't be visible.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/CompletionToken.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Wraps a CompletableFuture that is expected to be completed
+ * only in a single way and provides methods for easy of use
+ */
+public class CompletionToken {
+
+    private final CompletableFuture<Void> future = new CompletableFuture<>();
+
+    public boolean complete() {
+        return completeVoidFuture(future);
```
cangencer: this method is now deleted after spotbugs change, you need to rebase

metanet:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
```diff
@@ -95,4 +95,14 @@ default void join() {
         Util.uncheckRun(() -> getFuture().get());
     }
 
+    /**
+     * Cancels the current execution if the job is currently in running and
+     * schedules a new execution with the current member list of the Jet cluster
+     *
+     * @throws IllegalStateException if the job has been already completed
+     *
+     * @return true if the current execution of the job is cancelled
```
viliam-durina: It would be good to mention when `false` is returned. I guess it's when job is not yet started or is already restarting.

metanet: We already mention when it returns true: "Cancels the current execution if the job is currently in running". Do we need to mention when it returns false?

viliam-durina: I think it would be good. It made me wonder why wouldn't it succeed.

Issue #698, opened 2018-02-02
Allow reusing 1st stage windowing accumulation for different slide lengths


viliam-durina:
Allow this sliding window calculation setup:
- 1st stage: accumulate into tumblingWindow(10sec)
- 2nd stages: combine into all:
  - slidingWindow(30sec, 10sec)
  - slidingWindow(60sec, 10sec)
  - slidingWindow(5min, 1min): here multiple 1st stage frames will be combined into single slide

PR #699, opened 2018-02-05
Group frames in 2nd stage sliding window

viliam-durina:
Will allow this scenario:
- 1st sliding window stage will accumulate into tumblingWindowDef(10sec)
- 2nd stage will combine into slidingWindowDef(5min, 1min)

So 1 2nd stage window will combine 6 1st stage frames into 1 frame.

This is useful if you want to connect multiple 2nd stages to single 1st
stage and use different `slideBy` parameter.


viliam-durina:
verify

viliam-durina:
Fixes #698

PR #701, opened 2018-02-05
Fix intermittent test failures

viliam-durina:
Fixed `TestSupport`: it now creates new outbox when it creates new
processor: after snapshot restore, the processor is allowed to offer
different item due to reordering.

Two failures:

- StreamEventJournalPTest.smokeTest:
```
java.lang.AssertionError: Different item provided after offer() was
rejected
```

- time-based failure in
InsertWatermarksPTest.when_idleTimeout_then_idleMessageAfterTimeout


viliam-durina:
verify

PR #707, opened 2018-02-09
Fix sonar issues

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/operation/InitExecutionOperation.java
```diff
@@ -53,7 +53,6 @@ public InitExecutionOperation(long jobId, long executionId, int coordinatorMembe
         this.executionId = executionId;
         this.coordinatorMemberListVersion = coordinatorMemberListVersion;
         this.participants = participants;
-        this.executionId = executionId;
```
viliam-durina: We can also make the variable `final`

cangencer:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/operation/InitExecutionOperation.java
cangencer: It's deserialized in `readInternal()` so these fields can't be made final

PR #720, opened 2018-02-16
Add automodule names and update javadoc plugin to 3.0

cangencer:


cangencer:
verify

jerrinot:
why not to add a name for the core module?

PR #725, opened 2018-02-19
Change hazelcast-jet-all pom.xml for generating OSGi manifest

renatobertacco:
Replace maven-jar-plugin with maven-bundle-plugin in pom.xml

### DIFF
--- a/hazelcast-jet-all/pom.xml
+++ b/hazelcast-jet-all/pom.xml
@@ -109,33 +109,12 @@
                             <instructions>
                                 <Embed-Dependency>hazelcast-jet-core;inline=true</Embed-Dependency>
                                 <Embed-Transitive>true</Embed-Transitive>
-                                <Implementation-Title>${project.name}</Implementation-Title>
-                                <Implementation-Version>${project.version}</Implementation-Version>
-                                <Implementation-Vendor-Id>${project.groupId}</Implementation-Vendor-Id>
-                                <Implementation-Vendor>${project.organization.name}</Implementation-Vendor>
-                                <Implementation-URL>${project.url}</Implementation-URL>
-                                <Specification-Title>${project.name}</Specification-Title>
-                                <Specification-Version>${project.artifact.selectedVersion.majorVersion}.${project.artifact.selectedVersion.minorVersion}</Specification-Version>
-                                <Specification-Vendor>${project.organization.name}</Specification-Vendor>
-                                <Bundle-Activator>com.hazelcast.osgi.impl.Activator</Bundle-Activator>
                                 <Main-Class>com.hazelcast.jet.server.StartServer</Main-Class>
                                 <Automatic-Module-Name>com.hazelcast.jet</Automatic-Module-Name>
                                 <Export-Package>
-                                    com.hazelcast.*
+                                    *
                                 </Export-Package>
                                 <Import-Package>
-                                    !org.junit,
-                                    !com.hazelcast.*,
-                                    !com.eclipsesource.json,
-                                    sun.misc;resolution:=optional,
-                                    javax.cache;resolution:=optional,
-                                    javax.cache.*;resolution:=optional,
-                                    org.apache.log4j;resolution:=optional,
-                                    org.apache.logging.log4j;resolution:=optional,
-                                    org.apache.logging.log4j.*;resolution:=optional,
-                                    org.slf4j;resolution:=optional,
-                                    org.codehaus.groovy.jsr223;resolution:=optional,
-                                    org.jruby.embed.jsr223;resolution:=optional,
                                     *
                                 </Import-Package>
                             </instructions>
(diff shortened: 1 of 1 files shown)


cangencer:
There is a conflict which needs to be resolved, as there's been a new line added to the manifest file.
I can also see that `HazelcastOSGiService` would need to be adapted to be used with Jet. Are you using Hazelcast as an OSGi service?

renatobertacco:
No, I do not need to use Hazelcast as OSGi service

hazelcast-jet-all/pom.xml
```diff
@@ -127,10 +90,80 @@
...
+                    <execution>
+                        <id>bundle-manifest</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>bundle</goal>
+                        </goals>
+                        <configuration>
+                            <supportedProjectTypes>jar</supportedProjectTypes>
+                            <instructions>
+                                <Embed-Dependency>hazelcast-jet-core;inline=true</Embed-Dependency>
```
cangencer: is this needed? since the JAR is already shaded, and we don't have hazelcast-jet-core as an OSGi module

hazelcast-jet-all/pom.xml
```diff
@@ -127,10 +90,80 @@
...
+                                <Embed-Transitive>true</Embed-Transitive>
+                                <Implementation-Title>${project.name}</Implementation-Title>
+                                <Implementation-Version>${project.version}</Implementation-Version>
+                                <Implementation-Vendor-Id>${project.groupId}</Implementation-Vendor-Id>
+                                <Implementation-Vendor>${project.organization.name}</Implementation-Vendor>
+                                <Implementation-URL>${project.url}</Implementation-URL>
+                                <Specification-Title>${project.name}</Specification-Title>
```
cangencer: I believe these are not needed.

hazelcast-jet-all/pom.xml
```diff
@@ -127,10 +90,80 @@
...
+                                <Specification-Version>${project.artifact.selectedVersion.majorVersion}.${project.artifact.selectedVersion.minorVersion}</Specification-Version>
+                                <Specification-Vendor>${project.organization.name}</Specification-Vendor>
+                                <Bundle-Activator>com.hazelcast.osgi.impl.Activator</Bundle-Activator>
```
cangencer: we will leave the bundle activator out for now, as it can only create HZ instances

hazelcast-jet-all/pom.xml
```diff
@@ -127,10 +90,80 @@
...
+                                <Main-Class>com.hazelcast.jet.server.StartServer</Main-Class>
+                                <Automatic-Module-Name>com.hazelcast.jet</Automatic-Module-Name>
+                                <Export-Package>
+                                    com.hazelcast.*
```
cangencer: Although this seems to be the way Hazelcast does it, we should instead only export public packages and not implementation ones.

hazelcast-jet-all/pom.xml
```diff
@@ -127,10 +90,80 @@
...
+                                </Export-Package>
+                                <Import-Package>
+                                    !org.junit,
```
cangencer: I went through these dependenceis and not all of them are needed here. These are created here if you leave it as `*` 

```com.hazelcast.aws,
javax.annotation,
javax.cache;version="[1.0,2)";resolution:=optional,
javax.cache.configuration;resolution:=optional,
javax.cache.event;version="[1.0,2)";resolution:=optional
javax.cache.expiry;resolution:=optional
javax.cache.integration;resolution:=optional
javax.cache.management;resolution:=optional
javax.cache.processor;resolution:=optional
javax.cache.spi;version="[1.0,2)";resolution:=optional
javax.management
javax.management.openmbean
javax.net.ssl
javax.script
javax.security.auth
javax.security.auth.callback,javax.security.auth.login
javax.transaction.xa,javax.xml.namespace,javax.xml.parsers,javax.xml.transform
javax.xml.transform.dom
javax.xml.transform.sax,javax.xml.transform.stream
javax.xml.validation,javax.xml.xpath,org.apache.log4j,org.apache.log4j.spi,org.apache.logging.log4j
org.apache.logging.log4j.spi,org.osgi.framework
org.slf4j
org.w3c.dom
org.xml.sax
sun.misc```

hazelcast-jet-all/pom.xml
renatobertacco: Yes, it is. We need to do it twice because bnd-tool needs all dependencies for producing the Manifest file.

hazelcast-jet-all/pom.xml
renatobertacco: I added because these fields were present in the previous Manifest, but I can remove them

hazelcast-jet-all/pom.xml
renatobertacco: I copied the the hazelcast pom https://github.com/hazelcast/hazelcast/blob/master/hazelcast/pom.xml but I can leave *

cangencer:
Closing as the PR is stale and based on too old version.

PR #737, opened 2018-02-23
Add distributed() to single-stage aggregations

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/transform/WindowAggregateTransform.java
```diff
@@ -146,6 +145,6 @@ private void addSessionWindow(Planner p, SessionWindowDef wDef) {
                         nCopies(aggrOp.arity(), constantKey()),
                         aggrOp,
                         mapToOutputFn.toKeyedWindowResultFn()));
-        p.addEdges(this, pv.v, Edge::allToOne);
+        p.addEdges(this, pv.v, edge -> edge.distributed().allToOne());
```
cangencer: might be good to add diagram for single stage as well (even if just two vertices)

PR #744, opened 2018-02-26
Add mapWithContext operations

viliam-durina:
Also:
- TestSupport calls ProcessorSupplier.complete(), if verifyProcessor was
given a processor supplier


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/Processors.java
```diff
@@ -675,26 +679,82 @@ private Processors() {
...
+     * @param contextFactory the context factory
+     * @param mapFn a stateless mapping function
+     * @param <C> type of context object
+     * @param <T> type of received item
+     * @param <R> type of emitted item
+     */
+    @Nonnull
+    public static <C, T, R> ProcessorSupplier mapUsingContextP(
+            @Nonnull ContextFactory<C> contextFactory,
+            @Nonnull DistributedBiFunction<C, ? super T, R> mapFn
```
cangencer: minor but missing variance, `? super C, ? extends R`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/transform/MapUsingContextTransform.java
```diff
@@ -0,0 +1,48 @@
...
+import static com.hazelcast.jet.core.processor.Processors.mapUsingContextP;
+
+public class MapUsingContextTransform<C, T, R> extends AbstractTransform {
+    private final ContextFactory<C> contextFactory;
+    private final DistributedBiFunction<C, T, R> mapFn;
+
+    public MapUsingContextTransform(
+            @Nonnull Transform upstream,
+            @Nonnull ContextFactory<C> contextFactory,
+            @Nonnull DistributedBiFunction<C, T, R> mapFn
```
cangencer: missing variance here as well

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/Processors.java
viliam-durina: fixed

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/CloseableProcessorSupplier.java
```diff
@@ -57,13 +65,27 @@ public CloseableProcessorSupplier(DistributedSupplier<E> simpleSupplier) {
...
      */
     public CloseableProcessorSupplier(DistributedIntFunction<Collection<E>>  supplier) {
         this.supplier = supplier;
     }
 
+    /**
+     * Initializes this object with the given supplier. May only be invoked if
+     * the supplier property hasn't been enitialized yet.
+     */
+    public void setSupplier(DistributedIntFunction<Collection<E>> newSupplier) {
```
cangencer: I don't quite like this.. why not have the supplier take as constructor `Function<JetInstance, E>` instead? or `Function<Context, E>` ?

viliam-durina: It will need to be `Function<JetInstance, Supplier<E>>` because I need to create the shared context based on JetInstance and share it for all processor instances. I wanted to avoid supplier of supplier.

cangencer: I think it's fine, I find this code hard to follow and it requires that we extend the class rather than just using it.

viliam-durina: yes.. We might refactor everything to use setSupplier. Super-constructor call is bad in that you can't use any local state in it.

viliam-durina: refactored it to only have `setSupplier`, it's simpler now

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/CloseableProcessorSupplier.java
```diff
@@ -23,45 +23,56 @@
...
  * <p>
- * The {@code close()} method is called after all other calls on any local
- * processor have returned.
+ * The processors must implement {@link Closeable}. The call to {@code
+ * close()} will be the very last call on the processor.
  *
- * @param <E> Processor type
+ * @param <E> the processor type
  */
 public class CloseableProcessorSupplier<E extends Processor & Closeable> implements ProcessorSupplier {
```
cangencer: we should mark the class as final

viliam-durina: we can't - we need to override the `init` method

cangencer: I don't think this class should be extensible. I would prefer that instead we take a Function<Context, Supplier<Processor>.

viliam-durina: as per discussion, let's just make the class abstract, create factory methods and add first-class `Processor.close()` support in future PR

PR #748, opened 2018-02-27
Logging / toString improvements for windowing types

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/SlidingWindowP.java
```diff
@@ -309,6 +311,20 @@ private boolean flushBuffers() {
...
+    private void logLateEvent(long currentWm, @Nonnull Object item) {
+        if (item instanceof JetEvent) {
+            JetEvent event = (JetEvent) item;
+            getLogger().info(
+                    String.format("Event dropped, late by %dms. currentWatermark=%s, eventTime=%s, event=%s",
+                    currentWm - event.timestamp(), toLocalTime(currentWm), toLocalTime(event.timestamp()), event.payload()
+            ));
+        } else {
+            getLogger().info(String.format(
+                    "Event dropped because it was late currentWatermark=%s, event=%s", new Watermark(currentWm), item
```
viliam-durina: Can also be simpler: "Late event dropped; currentWatermark=..."

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -72,6 +73,7 @@
 
     private static final int BUFFER_SIZE = 1 << 15;
     private static final char[] ID_TEMPLATE = "0000-0000-0000-0000".toCharArray();
+    private static final DateTimeFormatter LOCAL_TIME_FORMATTER = DateTimeFormatter.ofPattern("HH:mm:ss.SSS");
```
viliam-durina: We should use `DateTimeFormatter.ofPattern("HH:mm:ss.SSS", Locale.US);`, even though I'm not sure if it makes a difference in formatting time...

cangencer: no it won't make any difference.. it's just numbers :)

PR #750, opened 2018-02-27
Follow changes in EventJournal

### LINKED ISSUE #741: Follow event journal changes in Hazelcast
See PR: https://github.com/hazelcast/hazelcast/pull/12429

`StaleSequenceException` is no longer thrown but instead there is a way to see how many items have been dropped.

If a stale seq is requested, you will get the lost item count as well as the current head.
If a sequence greater than newest is requested, you will get the tail + 1

cangencer:
Fixes #741

Should be merged after https://github.com/hazelcast/hazelcast/pull/12429

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
@@ -25,7 +25,6 @@ import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.core.Partition;
 import com.hazelcast.internal.journal.EventJournalInitialSubscriberState;
 import com.hazelcast.internal.journal.EventJournalReader;
-import com.hazelcast.jet.pipeline.JournalInitialPosition;
 import com.hazelcast.jet.Traverser;
 import com.hazelcast.jet.Traversers;
 import com.hazelcast.jet.core.AbstractProcessor;
@@ -40,6 +39,7 @@ import com.hazelcast.jet.core.processor.Processors;
 import com.hazelcast.jet.core.processor.SourceProcessors;
 import com.hazelcast.jet.function.DistributedFunction;
 import com.hazelcast.jet.function.DistributedPredicate;
+import com.hazelcast.jet.pipeline.JournalInitialPosition;
 import com.hazelcast.map.journal.EventJournalMapEvent;
 import com.hazelcast.nio.Address;
 import com.hazelcast.projection.Projection;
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/StreamEventJournalPTest.java
```diff
@@ -57,9 +57,13 @@
 
     private static final int NUM_PARTITIONS = 2;
     private static final int JOURNAL_CAPACITY = 10;
+    private static final int CAPACITY_PER_PARTITION = JOURNAL_CAPACITY / NUM_PARTITIONS;
```
viliam-durina: I'd rather do 
```
private static final int CAPACITY_PER_PARTITION = 5;
private static final int JOURNAL_CAPACITY = CAPACITY_PER_PARTITION * NUM_PARTITIONS;
```
With division there could be remainder.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/StreamEventJournalPTest.java
```diff
@@ -169,17 +163,15 @@ public void when_staleSequence_afterRestore() {
...
             outbox.drainQueueAndReset(0, output, true);
-            assertTrue("consumed more items than expected", output.size() == 0);
+            assertTrue("consumed different number of items than expected", output.size() == 0);
         }, 3);
 
         assertTrueEventually(() -> {
             assertTrue("Processor did not finish snapshot", p.saveToSnapshot());
         }, 3);
 
         // overflow journal
```
viliam-durina: "overflow _the_ journal"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -291,15 +299,6 @@ private void tryGetNextResultSet() {
                 // This exception can be safely ignored - it means the instance was shutting down,
                 // so we shouldn't unnecessarily throw an exception here.
                 return null;
-            } else if (ex instanceof StaleSequenceException) {
```
mmedenjak: You can still get the exception if the operation is parked on the member side (and may have read some items) after which the partition gets lost so you might want to keep this block.

viliam-durina: we use minFetchSize=1, so it should not get parked after after it read some items

PR #753, opened 2018-02-27
Add automatic WM throttling to Pipeline API

### LINKED ISSUE #743: Planner should use more optimal default watermark emission policy
Currently the planner always uses the `suppressDuplicates` policy which can result in redundant watermarks when a larger window size than 1 is used. The planner needs to figure out the optimal watermark emission rate based on the types of window being used.

viliam-durina:
Also modify WatermarkEmissionPolicy: it will tell what the new watermark
should be, not just whether to forward or not. For example, if
frameSize=1000 and new WM from the policy is 1234, it will not just tell
to forward 1234, but it will tell to forward 1000. This way WMs from all
parallel source instances will align and the WMs will be really
throttled.

It's implemented in `Planner.createdDag`: it checks frameDefs from all
transforms, calculates GCD and then changes emission policies in all WM
generating transforms.

Fixes  #743


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/Planner.java
```diff
@@ -55,13 +62,69 @@
...
+
         Iterable<Transform> sorted = topologicalSort(adjacencyMap, Object::toString);
         for (Transform transform : sorted) {
             transform.addToDag(this);
         }
         return dag;
     }
 
+    // package-visible for tests
+    static long[] calculateGcd(List<long[]> frameDefs) {
```
cangencer: this could be encapsulated in another class I think, as it's not specific to planner

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/pipeline/PlannerTest.java
```diff
@@ -0,0 +1,102 @@
...
+        assertArrayEquals(frameDef(0, 0), Planner.calculateGcd(asList(frameDef(0, 0))));
+    }
+
+    /**
+     * Generate random frame requirements and assert that their watermarks are
+     * on the verge of calculated frame length.
+     */
+    @Test
+    public void calculateGcd_randomTest() {
+        Random rnd = new Random(123); // use fixed seed to have repeatable test
```
cangencer: not fan of huge random tests, wouldn't it be better to have more deterministic values here?

PR #754, opened 2018-02-27
Make parallelism available to ProcessorSupplier.Context and Processor.Context

cangencer:
New additional fields of `localParallelism` and `totalParallelism` are now available in `Processor `and `ProcessorSupplier` level. 

Also made them extend the upstream interfaces, so that gradually more information is available in each context as you drill down.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
@@ -17,7 +17,6 @@
 package com.hazelcast.jet.core;
 
 import com.hazelcast.jet.JetException;
-import com.hazelcast.jet.JetInstance;
 import com.hazelcast.jet.config.ProcessingGuarantee;
 import com.hazelcast.logging.ILogger;
 
@@ -234,13 +233,7 @@ public interface Processor {
      * Context passed to the processor in the
      * {@link #init(Outbox, Context) init()} call.
      */
-    interface Context {
-
-        /**
-         * Returns the current Jet instance
-         */
-        @Nonnull
-        JetInstance jetInstance();
+    interface Context extends ProcessorSupplier.Context {
 
         /**
          *  Return a logger for the processor
@@ -253,12 +246,6 @@ public interface Processor {
          * this vertex on all nodes: its unique cluster-wide index.
          */
         int globalProcessorIndex();
-
-        /***
-         * Returns the name of the vertex associated with this processor.
-         */
-        @Nonnull
-        String vertexName();
 
         /**
          * Returns true, if snapshots will be saved for this job.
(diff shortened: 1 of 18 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
```diff
@@ -254,12 +247,6 @@ default boolean finishSnapshotRestore() {
          */
         int globalProcessorIndex();
```
viliam-durina: We should document here that it's zero-based and that it's in the range 0 .. totalParallelism-1

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
```diff
@@ -234,13 +233,7 @@ default boolean finishSnapshotRestore() {
...
-
-        /**
-         * Returns the current Jet instance
-         */
-        @Nonnull
-        JetInstance jetInstance();
+    interface Context extends ProcessorSupplier.Context {
 
         /**
          *  Return a logger for the processor
```
viliam-durina: The `logger()` is also in `PS.Context`, we can delete it here maybe. Is it only due to javadoc?
Also, the `processingGuarantee` and `snapshottingEnabled` methods can be pushed up to PMS level.

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java
```diff
@@ -271,43 +266,16 @@ public void restoreFromSnapshot(@Nonnull Object key, @Nonnull Object value) {
...
-            );
-        }
+    @Nonnull
+    public static <K, V, T> ProcessorSupplier processorSupplier(
+            @Nonnull Properties properties,
+            @Nonnull List<String> topics,
+            @Nonnull DistributedFunction<ConsumerRecord<K, V>, T> projectionFn,
+            @Nonnull WatermarkGenerationParams<T> wmGenParams
+    ) {
+        return new CloseableProcessorSupplier<StreamKafkaP>(() -> new StreamKafkaP<>(
```
viliam-durina: We've lost the preferred local parallelism of 2 here.
And the explicit type parameter is not needed.

cangencer: I've moved the preferred parallelism to `KafkaProcessors.streamKafkaP`. Will fix the explicit type params.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
cangencer: Yes they end up all being the same thing. Only ProcCtx has the unique index so maybe we should just have two types rather than three?

hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/impl/connector/kafka/StreamKafkaP.java
viliam-durina: ok, missed that

PR #755, opened 2018-02-28
Add and fix Javadoc

mtopolnik:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/Sources.java
```diff
@@ -97,7 +93,15 @@ private Sources() {
     }
 
     /**
-     * Javadoc pending
+     * Returns an unbounded (event stream) source that will use the supplied
+     * function to create processor meta-suppliers as required by the Core API.
+     * Jet will call the function you supply with the watermark generation
+     * paramters and it must return a meta-supplier of processors that will
```
cangencer: typo "paramters"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/Sources.java
```diff
@@ -97,7 +93,15 @@ private Sources() {
...
+     * act according to these parameters and emit the watermark items as they
+     * specify.
+     *
```
cangencer: we could also mention here that the source should make use of `WatermarkSourceUtil`

PR #760, opened 2018-03-01
Fix intermittent failure in SnapshotFailureTest

### LINKED ISSUE #759: com.hazelcast.jet.core.SnapshotFailureTest.when_snapshotFails_then_jobShouldNotFail
https://hazelcast-l337.ci.cloudbees.com/job/Jet-pr-builder/com.hazelcast.jet$hazelcast-jet-core/2484/testReport/junit/com.hazelcast.jet.core/SnapshotFailureTest/when_snapshotFails_then_jobShouldNotFail/

```
Error Message

no failed snapshot appeared in snapshotsMap

Stacktrace

java.lang.AssertionError: no failed snapshot appeared in snapshotsMap
	at com.hazelcast.jet.core.SnapshotFailureTest.when_snapshotFails_then_jobShouldNotFail(SnapshotFailureTest.java:121)
```

viliam-durina:
The test was racy, records for failed snapshots are deleted right after
they are marked as FAILED. This commit rewrites it to check using map
journal where we shouldn't miss events.

Fixes #759

### DIFF
--- a/hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/SnapshotFailureTest.java
+++ b/hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/SnapshotFailureTest.java
@@ -17,9 +17,10 @@
 package com.hazelcast.jet.core;
 
 import com.hazelcast.client.map.helpers.AMapStore;
+import com.hazelcast.config.EventJournalConfig;
 import com.hazelcast.config.MapConfig;
 import com.hazelcast.config.MapStoreConfig;
-import com.hazelcast.core.IMap;
+import com.hazelcast.jet.IMapJet;
 import com.hazelcast.jet.JetInstance;
 import com.hazelcast.jet.Job;
 import com.hazelcast.jet.config.JetConfig;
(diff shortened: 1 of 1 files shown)


hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/SnapshotFailureTest.java
```diff
@@ -97,28 +103,25 @@ public void when_snapshotFails_then_jobShouldNotFail() throws Exception {
...
+        p.drawFrom(Sources.mapJournal(snapshotsMapName(job.getId()),
+                event -> event.getNewValue() instanceof SnapshotRecord
+                        && ((SnapshotRecord) event.getNewValue()).status() == SnapshotStatus.FAILED,
+                EventJournalMapEvent::getNewValue,
+                JournalInitialPosition.START_FROM_OLDEST))
+         .peek()
+         .drainTo(Sinks.list("failed_snapshot_records"));
+        instance1.newJob(p);
 
         job.join();
```
eminn: Could it be happen that before second job is written it's results to the list, first job finishes and make the following assertions earlier and fails ?

viliam-durina: The first job takes 5 seconds to complete, it does snapshot every 100ms. In this case there were 6 failed snapshots made. If the second job is not able to process any event from the journal during 5 seconds, it will fail. If it ever fails, i'll increase the limits.

eminn: 👍

PR #770, opened 2018-03-09
[MERGE TO 0.6.1] Add custom class loader support

viliam-durina:


viliam-durina:
Should be ready to merge, but deferred to 0.6.1

PR #794, opened 2018-03-23
Fix deserialization error reporting from master

### LINKED ISSUE #786: Job failure due to missing class not reported to client
Steps to reproduce:
1. start standalone jet instance
2. in another jvm, create a client instance and submit a job, without adding the classes to `JobConfig`
3. the job fails on member due to the missing class, as expected
4. call `Job.join()` on the client, it blocks indefinitely. 

Expected behavior: `Job.join()` should throw an exception.

viliam-durina:
Fixes #786

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
@@ -209,7 +209,7 @@ public class MasterContext {
             executionPlanMap = createExecutionPlans(nodeEngine, membersView, dag, getJobConfig(), lastSnapshotId);
         } catch (Exception e) {
             logger.severe("Exception creating execution plan for " + jobIdString(), e);
-            onCompleteStepCompleted(e);
+            finalizeJob(e);
             return;
         } finally {
             Thread.currentThread().setContextClassLoader(previousCL);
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -170,7 +170,14 @@ void tryStartJob(Function<Long, Long> executionIdSupplier) {
             return;
         }
 
-        DAG dag = deserializeDAG();
+        DAG dag;
+        try {
+            dag = deserializeDAG();
+        } catch (RuntimeException e) {
+            logger.warning("DAG deserialization failed", e);
+            finalizeJob(e);
```
metanet: shouldn't we call `onCompleteStepCompleted()` here?

cangencer: I don't think you'll be able to complete, because verticies won't be init

cangencer: Ok I see, then it should be safe to call `onCompleteStepCompleted` instead of creating a finalizeJob method.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -170,7 +170,14 @@ void tryStartJob(Function<Long, Long> executionIdSupplier) {
             return;
         }
 
-        DAG dag = deserializeDAG();
+        DAG dag;
+        try {
+            dag = deserializeDAG();
+        } catch (RuntimeException e) {
```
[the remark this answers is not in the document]
viliam-durina: `Exception` is not thrown there, unless there's a `sneakyThrow`. But I replaced it.

PR #806, opened 2018-04-05
Minor logging improvements

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamEventJournalP.java
```diff
@@ -275,13 +277,14 @@ private void tryGetNextResultSet() {
                 if (lostCount > 0) {
                     getLogger().warning(lostCount +  " events lost for partition " +
                             partitionId + " due to journal overflow when reading from event journal." +
-                            " Increase journal size to avoid this error.");
+                            " Increase journal size to avoid this error. "
+                            + "getNextSequenceToReadFrom=" + resultSet.getNextSequenceToReadFrom() +
+                            "readCount=" + resultSet.readCount() + " prevSeq=" + prevSequence);
```
viliam-durina: there should be `+` sign at the beginning of all lines, here it's mixed

PR #808, opened 2018-04-06
Assert same item offered to Outbox after rejection

viliam-durina:


viliam-durina:
Test will succeed after #809 is merged

PR #809, opened 2018-04-06
Source processors should emit any pending items before starting a snapshot

cangencer:
Source processors are required to emit any pending items that were not 
fully emitted before starting a snapshot. This prevents a snapshot barrier
being emitted while an item has only been emitted to a subset of outbound
queues. However this was not respected by the current streaming source
processors.

As a consequence, the afterEmit callback on AbstractProcessor is also
removed as it is now redundant.


cangencer:
merged into #808

PR #858, opened 2018-04-26
Use hostname equality if reverse lookup throws UnknownHostException exception

### LINKED ISSUE #810: Do no fail job if HDFS Source reverse DNS lookup fails
Currently when the Jet HDFS source tries to distribute HDFS splits to all the members it tries to make use of data locality by matching the IP address of the member to the hostname of the datanode where a split is present. The member IP address is converted to a hostname using a reverse DNS lookup. However if the lookup fails with `UnknownHostException` the job will fail to start. Even though this typically points to a configuration error, it might be better to print a warning instead and not fail the job. The caveat will be that Jet will not be able to make use of data locality if reverse DNS resolution isn't working.

Alternatively, we could only use forward lookups and match HDFS hosts against IP addresses instead.

gurbuzali:
fixes #810

### DIFF
--- a/hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsP.java
+++ b/hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsP.java
@@ -276,6 +276,9 @@ public final class ReadHdfsP<K, V, R> extends AbstractProcessor {
         }
 
         private static boolean isSplitLocalForMember(InputSplit split, String hostName) {
+            if (hostName == null) {
+                return false;
+            }
             try {
                 return Arrays.stream(split.getLocations())
                              .anyMatch(hostName::equalsIgnoreCase);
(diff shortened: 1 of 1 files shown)


hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsP.java
```diff
@@ -266,6 +267,18 @@ private static boolean isSplitLocalForMember(InputSplit split, Address memberAdd
                 return Arrays.stream(split.getLocations())
                         .flatMap(loc -> Arrays.stream(uncheckCall(() -> InetAddress.getAllByName(loc))))
                         .anyMatch(inetAddr::equals);
+            } catch (IOException e) {
+                if (e instanceof UnknownHostException) {
+                    return isSplitLocalForMember(split, memberAddr.getScopedHost());
```
viliam-durina: We should log this at warning level if it happens.

PR #883, opened 2018-05-28
Close jms objects explicitly, without casting to AutoClosable for backward compatibility

gurbuzali:


viliam-durina:
The problem was that in JMS 1.1 `Connection`, `Session` and `Consumer` classes don't implement `AutoCloseable`, but in `2.0` they do. The PR replaces the `closeIfNotNull(AutoCloseable)` method with direct class to `close` method, which, as per ali's testing, fixes the problem.

Issue #884, opened 2018-05-29
SinkBuilder should take unique processor number


cangencer:
For `SinkBuilder.createFn` you should be able to supply a unique processor index so that it can be used for file type sinks with a unique number.

PR #886, opened 2018-05-29
Convert argument of SinkBuilder createFn to Processor.Context

gurbuzali:
Also add `preferredLocalParallelism` option to `SinkBuilder`
fixes https://github.com/hazelcast/hazelcast-jet/issues/884

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/SinkBuilder.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/SinkBuilder.java
@@ -16,13 +16,12 @@
 
 package com.hazelcast.jet.pipeline;
 
-import com.hazelcast.jet.JetInstance;
 import com.hazelcast.jet.core.Processor;
 import com.hazelcast.jet.core.ProcessorMetaSupplier;
 import com.hazelcast.jet.core.Vertex;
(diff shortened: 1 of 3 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/SinkBuilder.java
```diff
@@ -99,6 +101,18 @@
...
     }
 
+    /**
+     * Sets the local parallelism of the sink, default value is {@code 2}
+     *
+     * @param preferredLocalParallelism the local parallelism of the sink
+     */
+    @Nonnull
+    public SinkBuilder<W, T> preferredLocalParallelism(int preferredLocalParallelism) {
+        checkPositive(preferredLocalParallelism, "Preferred local parallelism should be a positive number");
```
viliam-durina: You should use `Vertex.checkLocalParallelism` here

viliam-durina:
What about passing entire `Processor.Context` object? It contains JetInstance and also the index, and few more.

gurbuzali:
Yes, I've thought about that too.`WriteBufferedP` expects `Processor.Context` but we change it for `SinkBuilder`. I don't know what was the reason behind this choice, that's why I've kept it as it is

viliam-durina:
I guess there was no specific reason. First we needed JetInstance, now the index. Next... ?

PR #892, opened 2018-05-31
Add transform-using-context for grouped streams

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/TransformUsingKeyedContextP.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Map;
+import java.util.Map.Entry;
+
+import static com.hazelcast.jet.Traversers.traverseIterable;
+
+/**
+ * Processor which, for each received item, emits all the items from the
+ * traverser returned by the given item-to-traverser function, using a context
+ * object.
+ */
```
mtopolnik: This Javadoc doesn't explain the "keyed" aspect.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/StreamStageImpl.java
```diff
@@ -97,6 +100,11 @@ public StreamStageImpl(
         return attachFlatMapUsingContext(contextFactory, flatMapFn);
     }
 
+    @Nonnull @Override
+    public <R> StreamStage<R> rollingAggregation(@Nonnull AggregateOperation1<? super T, ?, ? extends R> aggrOp) {
```
mtopolnik: The name `rollingAggregation` is not in line with other methods that are called `aggregate` and not `aggregation`. It is generally not in line with operation names being verbs.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/GeneralStage.java
```diff
@@ -162,6 +164,45 @@
...
+     * stage continues to accumulate to the accumulator after finishing
+     * function was called. This stage throws an exception if the finish
+     * function returns the same instance for any accumulator at runtime.
+     * <p>
+     * For example, {@link
+     * com.hazelcast.jet.aggregate.AggregateOperations#summingLong
+     * summingLong()} is OK because its result is an immutable {@code Long}.
+     * {@link com.hazelcast.jet.aggregate.AggregateOperations#toSet() toSet()}
+     * is not because its result is the same {@code Set} instance to which it
+     * accumulates.
```
mtopolnik: We've got a pretty lame story here. We're basically violating the contract of `AggregateOperation` as it has been defined so far.

viliam-durina: Yes, we are. But what else? Shall we serde the result? Shall we require the finishFn to return an object that will not refer to the accumulator so that the accumulator can be further accumulated to? I think neither of this... Both will harm performance for an edge case. We add a requirement and check it (partly, only for identity).

PR #893, opened 2018-06-01
Javadoc cleanup in aggregate operation builders

mtopolnik:
Removed leftover duplicate Javadoc

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/CoAggregateOperationBuilder.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/CoAggregateOperationBuilder.java
@@ -20,7 +20,6 @@ import com.hazelcast.jet.datamodel.ItemsByTag;
 import com.hazelcast.jet.datamodel.Tag;
 import com.hazelcast.jet.function.DistributedBiConsumer;
 import com.hazelcast.jet.function.DistributedFunction;
-import com.hazelcast.jet.pipeline.StageWithWindow;
 import com.hazelcast.util.Preconditions;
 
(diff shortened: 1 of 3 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/CoAggregateOperationBuilder.java
```diff
@@ -39,17 +39,9 @@
...
- * <p>
- * To obtain this builder, call {@link AggregateOperations#coAggregateOperationBuilder()}.
- * This builder is suitable only if you'll use independent aggregate
- * operations on each input and expect to receive a separate result for
- * each of them. You can combine the individual results in the final
- * step by providing a {@code finishFn}.
+ * accepts multiple inputs. To obtain it, call {@link
+ * AggregateOperations#coAggregateOperationBuilder()}. and refer to that
+ * method's Javadoc for further details.
  */
```
viliam-durina: I'd say only
```
/** See {@link AggregateOperations#coAggregateOperationBuilder()} */
```
in all three cases in this PR. Much quicker to read and you have to go there anyway. No need to repeat the first sentence. Same as [here](https://github.com/hazelcast/hazelcast-jet/blob/master/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/SinkBuilder.java#L35).

mtopolnik: I prefer it this way because it tells you something about the class when you're browsing around. So you know whether you want to follow another link or you're not interested. In any case, these changes just align the descriptions with everything else.

PR #900, opened 2018-06-05
Usability improvements

viliam-durina:
- kafka source requires to give at least one topic (uses `String topic,
String ... moreTopics` method contract). Also better documentation of
offset committing.

- add javadoc warnings about null key, null timestamps or nonsensical
timestamps returned in lambdas

- check for null key returned from edge partitioner key extractor and
print a better error message

- remove the constructor in TimestampedEntry and TimestampedItem with
(with `ignored` parameter). It was hard to figure out what this is for.
Replaced with a static method.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/datamodel/TimestampedEntry.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/datamodel/TimestampedEntry.java
@@ -47,17 +47,6 @@ public final class TimestampedEntry<K, V> implements Map.Entry<K, V> {
     }
 
     /**
-     * This constructor exists in order to match the shape of the functional
-     * interface {@link KeyedWindowResultFunction}.
-     * <p>
-     * Constructs a timestamped entry with the supplied field values. Ignores
-     * the first argument.
-     */
-    public TimestampedEntry(long ignored, long timestamp, @Nonnull K key, @Nonnull V value) {
-        this(timestamp, key, value);
-    }
-
-    /**
      * Returns the timestamp of this entry.
      */
     public long getTimestamp() {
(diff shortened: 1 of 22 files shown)


viliam-durina:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/datamodel/TimestampedEntry.java
```diff
@@ -102,4 +91,16 @@ public boolean equals(Object obj) {
...
     }
+
+    /**
+     * This method matches the shape of the functional interface {@link
+     * KeyedWindowResultFunction}.
+     * <p>
+     * Constructs a {@code TimestampedEntry} using the window end time as the
+     * timestamp.
+     */
+    public static <K, V> TimestampedEntry<K, V> mapWindowResult(long winStart, long winEnd, @Nonnull K key,
```
cangencer: I think `fromWindowResult` would be a better name

hazelcast-jet-core/src/main/java/com/hazelcast/jet/datamodel/TimestampedItem.java
```diff
@@ -91,4 +80,16 @@ public String toString() {
...
+
+
+    /**
+     * This method matches the shape of the functional interface {@link
+     * WindowResultFunction}.
+     * <p>
+     * Constructs a {@code TimestampedItem} using the window end time as the
+     * timestamp.
+     */
+    public static <V> TimestampedItem<V> mapWindowResult(long winStart, long winEnd, @Nonnull V value) {
```
cangencer: same as previous comment

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/GeneralStage.java
```diff
@@ -339,9 +343,16 @@
      * watermark without causing dropped events. If you add the timestamps
      * later on, events from different partitions may be mixed, increasing
      * the perceived event lag and causing more dropped events.
+     * <p>
+     * <b>Warning:</b> make sure the property you access in {@code timestampFn}
+     * isn't null, it would fail the job. Also that there are no nonsensical
+     * values such as -1, MIN_VALUE, 2100-01-01 etc - we'll treat those as real
+     * timestamps and they can mess the logic.
```
cangencer: can we rephase this? "mess the logic" isn't a good term

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/GeneralStage.java
```diff
@@ -339,9 +343,16 @@
...
      *
      * @param timestampFn a function that returns the timestamp for each item
-     * @param allowedLag the allowed lag behind the top observed timestamp
+     * @param allowedLag the allowed lag behind the top observed timestamp.
```
cangencer: we should say "typically milliseconds"

vladoschreiner:
See also https://github.com/hazelcast/hazelcast-jet-reference-manual/issues/80

Issue #899, opened 2018-06-04
com.hazelcast.jet.core.JobRestartWithSnapshotTest.when_nodeDown_then_jobRestartsFromSnapshot_twoStage


cangencer:
https://hazelcast-l337.ci.cloudbees.com/view/Jet/job/jet-jdk9/com.hazelcast.jet$hazelcast-jet-core/209/testReport/junit/com.hazelcast.jet.core/JobRestartWithSnapshotTest/when_nodeDown_then_jobRestartsFromSnapshot_twoStage/

```Error Message
No snapshot record found - job likely finished
Stacktrace
java.lang.AssertionError: No snapshot record found - job likely finished
	at com.hazelcast.jet.core.JobRestartWithSnapshotTest.lambda$waitForNextSnapshot$9(JobRestartWithSnapshotTest.java:364)
	at com.hazelcast.jet.core.JobRestartWithSnapshotTest.waitForNextSnapshot(JobRestartWithSnapshotTest.java:362)
	at com.hazelcast.jet.core.JobRestartWithSnapshotTest.when_nodeDown_then_jobRestartsFromSnapshot(JobRestartWithSnapshotTest.java:218)
	at com.hazelcast.jet.core.JobRestartWithSnapshotTest.when_nodeDown_then_jobRestartsFromSnapshot_twoStage(JobRestartWithSnapshotTest.java:117)```

**Comment by cangencer** [comment]
possibly related.. 


com.hazelcast.jet.core.JobRestartWithSnapshotTest.when_nodeDown_then_jobRestartsFromSnapshot_singleStage

```
Error Message
No snapshot record found - job likely finished
Stacktrace
java.lang.AssertionError: No snapshot record found - job likely finished
	at com.hazelcast.jet.core.JobRestartWithSnapshotTest.lambda$waitForNextSnapshot$9(JobRestartWithSnapshotTest.java:364)
	at com.hazelcast.jet.core.JobRestartWithSnapshotTest.waitForNextSnapshot(JobRestartWithSnapshotTest.java:362)
	at com.hazelcast.jet.core.JobRestartWithSnapshotTest.when_nodeDown_then_jobRestartsFromSnapshot(JobRestartWithSnapshotTest.java:218)
	at com.hazelcast.jet.core.JobRestartWithSnapshotTest.when_nodeDown_then_jobRestartsFromSnapshot_singleStage(JobRestartWithSnapshotTest.java:112)
```

[...]

PR #907, opened 2018-06-07
Add RestartableException & fix HazelcastWriters

viliam-durina:
We ignored the `HazelcastInstanceNotActiveException` in
`HazelcastWriters`, if the object written to was local. We relied that
the job will restart anyway and writing will work after that. But if the
job completed after such failure, but before a restart, the writing
error was silently ignored.

This commit fixes the problem by introducing `RestartableException` that
can be thrown from processors to indicate a failure that can be resolved
by a restart.


hazelcast-jet-core/src/main/java/com/hazelcast/jet/RestartableException.java
```diff
@@ -0,0 +1,59 @@
...
+ *          ProcessorSupplier} and {@link ProcessorMetaSupplier} (but not their
+ *          {@code close()} methods)
+ *     <li>...
+ * </ul>
+ *
+ * If this exception is caught, the job will be terminated abruptly and
+ * restarted (if {@link JobConfig#setAutoRestartOnMemberFailure(boolean)
+ * enabled}).
+ */
+public class RestartableException extends JetException {
```
cangencer: we also have `JobRestartRequestedException`, should these be consolidated somehow?

viliam-durina: `JobRestartRequestedException` is used internally that a job restart was requested (by calling `Job.restart()`). For example, it restarts the job even though `isAutoRestartOnMemberFailureEnabled` is disabled.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/HazelcastConnector_RestartTest.java
```diff
@@ -0,0 +1,142 @@
...
+        assertTrueEventually(() -> assertTrue("no output to sink", sinkList.size() >= 4), 5);
+
+        // When
+        // initiate the shutdown. Thanks to the MigrationBlockingService migration will not finish
+        // before the latch is counted down.
+        Future shutdownFuture = spawn(() -> instance2.shutdown());
+
+        // Then - assert that the job stopped producing output
+        sleepSeconds(2);
+        assertFalse("node engine is running",
```
metanet: `assertTrueEventually()` would be better here instead of a `sleep()` call.

viliam-durina:
Fixes #899

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/HazelcastConnector_RestartTest.java
viliam-durina: Doesn't work, the job continues a little while after `isRunning` becomes `true`. I think it becomes `true` quite quickly. It's here rather as a safety check, not really as a material conditional.

PR #912, opened 2018-06-08
Jet metrics, metric collection & API

viliam-durina:
- add background service for collecting HZ metrics
- add jet-specific metrics
- expose metrics through jet client


cangencer:
Metrics configuration is also missing XML support

viliam-durina:
>Metrics configuration is also missing XML support

Do we need to bother? It will be moved to imdg anyway.

viliam-durina:
verify

PR #890, opened 2018-05-31
Introduce Avro Connector

gurbuzali:
* make `ReadFilesP` generic so that it supports reading avro files
* add `localProcessorIndex` to `Processor.Context`
* add support for reading shared file system 
* introduce avro module

PR #913, opened 2018-06-11
Introduce file source/sink builders

gurbuzali:
Also add shared file system support to StreamFilesP

[Breaking Change for Public API]
- Adds a new boolean parameter `sharedFileSystem` to `SourceProcessors.streamFilesP`
- Changes `Sources.files` and `Sources.fileWatcher` to `Sources.filesBuilder`
- Changes `Sinks.files` to `Sinks.filesBuilder`

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/SourceProcessors.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/SourceProcessors.java
@@ -336,16 +336,17 @@ public final class SourceProcessors {
 
     /**
      * Returns a supplier of processors for
-     * {@link Sources#fileWatcher(String, Charset, String)}.
+     * {@link Sources#fileWatcher(String, Charset, String, boolean)}.
      */
     @Nonnull
     public static ProcessorMetaSupplier streamFilesP(
             @Nonnull String watchedDirectory,
             @Nonnull Charset charset,
             @Nonnull String glob,
-            @Nonnull DistributedBiFunction<String, String, ?> mapOutputFn
+            @Nonnull DistributedBiFunction<String, String, ?> mapOutputFn,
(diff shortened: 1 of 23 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/SourceProcessors.java
```diff
@@ -336,16 +336,17 @@ public static ProcessorMetaSupplier streamSocketP(
...
+     * {@link Sources#fileWatcher(String, Charset, String, boolean)}.
      */
     @Nonnull
     public static ProcessorMetaSupplier streamFilesP(
             @Nonnull String watchedDirectory,
             @Nonnull Charset charset,
             @Nonnull String glob,
-            @Nonnull DistributedBiFunction<String, String, ?> mapOutputFn
+            @Nonnull DistributedBiFunction<String, String, ?> mapOutputFn,
+            boolean sharedFileSystem
```
cangencer: shouldn't be last param.
What about `Sources.files`?

gurbuzali: I've already added shared files system support to `Sources.files` with Avro PR
https://github.com/hazelcast/hazelcast-jet/pull/890/files#diff-ebaff9dcd50ec0708250477d8b796454R771
it was the last parameter there too

cangencer: ok, can we make it so that it's not after the lambda for both? code looks generally nicer if the lambda is the last parameter.

viliam-durina:
I think it's time to convert this to builder. File source has 5 parameters, 4 of which are optional and used separately.

PR #924, opened 2018-06-15
More convenience for enrichment from Hazelcast maps

mtopolnik:
Adds these methods:

- `nonKeyedStage.mapUsingReplicatedMap()`
- `nonKeyedStage.mapUsingIMap()`
- `keyedStage.mapUsingIMap()` -- this one will achieve data locality through partitioning

Declares `{map,filter,flatMap}UsingContext` on both non-keyed and keyed stages. Expresses the above enrichment operations in terms of `mapUsingContext`.

Reformulates rolling aggregation to uses just the `AggregateOperation`. To support it, `AggregateOperation` got another primitive: `export`, which is like `finish` but preserves the state of the accumulator.


cangencer:
verify

cangencer:
TODO: breaking changes need to be documented

PR #926, opened 2018-06-18
Expose metrics through JMX

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/CompressingProbeRenderer.java
```diff
@@ -182,4 +209,14 @@ private void moveNext() {
             }
         };
     }
+
+    private static class MorePublicByteArrayOutputStream extends ByteArrayOutputStream {
```
cangencer: We used this pattern somewhere else I think? Maybe should be a Util class?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/CompressingProbeRenderer.java
```diff
@@ -49,18 +58,41 @@
 
     private static final short BINARY_FORMAT_VERSION = 1;
 
+    private final ILogger logger;
+    private ObjLongConsumer<byte[]> consumer;
```
cangencer: can be made final

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsRenderPlugin.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.jet.impl.metrics;
+
+public interface MetricsRenderPlugin {
```
cangencer: can we extend ProbeRenderer instead (with default implementations for renderNoValue and renderException) and also add some javadoc for the methods?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/CompressingProbeRenderer.java
viliam-durina: I did it several times, but it was always removed. This is the only instance now in jet code. And as a rule of thumb, single instance doesn't make it to `Util`. Btw, we should also make it non-synchronized, i reuse the instance and i wouldn't bet the JVM is able to elide locking...

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsRenderPlugin.java
viliam-durina: We handle the exceptions in a common way. `renderException` is supposed to send the exception text to the target (to MC or to JMX bean). In both our cases we just log the exception, if each plugin was a `ProbeRenderer` itself, the exception would be logged twice. 

`renderNoValue` is used in very special case: when the metric was deregistered while iterating the metrics. There's no point in having it.

cangencer: can we rename MetricsRendererPlugin -> MetricsConsumer and the methods appropriately? Rendering implies showing something visually and plugin implies dynamic discovery/registration..

cangencer:
Also missing Spring and XML config

PR #934, opened 2018-06-22
Eagerly close processors

### LINKED ISSUE #867: Processor cleanup should be done eagerly
Currently `Processor` instances are cleaned up only on job completion. The cleanup involves calling `Processor.close()` and also deregistering the metrics from `MetricsRegistry`. This should instead be done after each individual processor is completed, so that they don't unnecessarily hold on to resources.

viliam-durina:
Fixes #867

Also contains:
- add `jobId()`, `executionId()` and `jobConfig()` to `Processor.Context`
- move `idToString` to public `Util` class (to format the ID from
`Processor.Context`)

Breaking change:
- `Processor.close()` methods no longer has `Throwable` parameter and it is called earlier (right after the processor completes, can be before the job completes).

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
@@ -235,7 +235,7 @@ public interface Processor {
     }
 
     /**
-     * Called as the last method in the processor lifecycle*. It is called
+     * Called as the last method in the processor lifecycle. It is called
      * whether the job was successful or not, and strictly before {@link
      * ProcessorSupplier#close} is called on this member. The method might get
      * called even if {@link #init} method was not yet called.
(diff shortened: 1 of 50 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
```diff
@@ -236,24 +235,19 @@ default boolean finishSnapshotRestore() {
     }
 
     /**
-     * Called after the execution has finished on all members - successfully or
-     * not, before {@link ProcessorSupplier#close} is called. If the execution
-     * was <em>aborted</em> due to a member leaving the cluster it is called
-     * immediately. Int this case, it can happen that the job is still running
-     * on some other member (but not on this member).
+     * Called as the last method in the processor lifecycle*. It is called
```
mtopolnik: Stray `*` here

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/test/TestProcessorMetaSupplierContext.java
```diff
@@ -50,6 +54,45 @@ public TestProcessorMetaSupplierContext setJetInstance(@Nonnull JetInstance jetI
         return this;
     }
 
+    @Override
+    public long jobId() {
+        return jobId;
+    }
+
+    /**
+     * Set the job ID.
```
mtopolnik: It should say "_Sets_ the job ID." Likewise in the methods below.

PR #939, opened 2018-06-26
Prefer using job name in logs

### LINKED ISSUE #911: Job name missing in a log
Cluster log contains just generated Job IDs. It would be handy to include job names in the log records to differentiate between various jobs running in the cluster.

viliam-durina:
The following is used in log messages:

`AbstractJobProxy`: logs on the submitting client or member
```
  xxxx-xxxx-xxxx-xxxx (name ??)   - if config not loaded yet
  xxxx-xxxx-xxxx-xxxx (name '')   - if loaded, but name null
  xxxx-xxxx-xxxx-xxxx (name 'abc')
```

`MasterContext`: logs on coordinator member. Name is always known.
```
  job 'abc', execution xxx-xxx-xxx-xxx                - if name set
  job 'xxx-xxx-xxx-xxx', execution xxx-xxx-xxx-xxx    - if name not set
```

`ExecutionContext`: on each member, name not known for a short while
initially
```
  job 'abc', execution xxx-xxx-xxx-xxx              - if name set
  job xxx-xxx-xxx-xxx, execution xxx-xxx-xxx-xxx    - if name not set
```

Initial log, printed on each member on INFO level to give association for
jobId, jobName and executionId:

```
  Execution plan for jobId=xxx-xxx-xxx-xxx, jobName=null/'abc',
  executionId=xxx-xxx-xxx-xxx initialized
```

The important change is that if job name is set, job ID is not always
displayed. If the job name is not unique, we won't be able to tell the
jobs apart in logs. This is documented at `JobConfig.setName()` method.

Fixes #911

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJobProxy.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJobProxy.java
@@ -63,21 +63,19 @@ public abstract class AbstractJobProxy<T> implements Job {
     private final AtomicBoolean cancelledJob = new AtomicBoolean();
     private final ExecutionCallback<Void> joinJobCallback = new JoinJobCallback();
 
-    private final Supplier<JobConfig> jobConfigSup;
+    private volatile JobConfig jobConfig;
     private final Supplier<Long> submissionTimeSup = memoizeConcurrent(this::doGetJobSubmissionTime);
 
     AbstractJobProxy(T container, long jobId) {
         this.jobId = jobId;
-        this.jobConfigSup = memoizeConcurrent(this::doGetJobConfig);
         this.container = container;
-        this.logger = createLogger();
+        this.logger = loggingService().getLogger(Job.class);
     }
 
     AbstractJobProxy(T container, long jobId, DAG dag, JobConfig config) {
         this.jobId = jobId;
-        this.jobConfigSup = () -> config;
         this.container = container;
-        this.logger = createLogger();
+        this.logger = loggingService().getLogger(Job.class);
 
         doSubmitJob(dag, config);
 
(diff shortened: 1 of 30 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJobProxy.java
```diff
@@ -63,21 +63,19 @@
     private final AtomicBoolean cancelledJob = new AtomicBoolean();
     private final ExecutionCallback<Void> joinJobCallback = new JoinJobCallback();
 
-    private final Supplier<JobConfig> jobConfigSup;
+    private volatile JobConfig jobConfig;
     private final Supplier<Long> submissionTimeSup = memoizeConcurrent(this::doGetJobSubmissionTime);
 
     AbstractJobProxy(T container, long jobId) {
         this.jobId = jobId;
-        this.jobConfigSup = memoizeConcurrent(this::doGetJobConfig);
```
cangencer: what was this inlined?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -457,7 +456,7 @@ SnapshotRepository snapshotRepository() {
     /**
      * Completes the job which is coordinated with the given master context object.
      */
-    void completeJob(MasterContext masterContext, long executionId, long completionTime, Throwable error) {
+    void completeJob(MasterContext masterContext, long completionTime, Throwable error) {
         // the order of operations is important.
 
         long jobId = masterContext.getJobId();
```
cangencer: should rename this method to `jobId()` as well

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRecord.java
```diff
@@ -108,6 +108,7 @@ public void readData(ObjectDataInput in) throws IOException {
     public String toString() {
         return "JobRecord{" +
                 "jobId=" + idToString(jobId) +
+                ", jobConfig.name=" + getConfig().getName() +
```
cangencer: we can just say `name=...`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJobProxy.java
viliam-durina: because i needed to get the value without triggering a lazy load. It's needed in case of various errors where the could fail too.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRecord.java
viliam-durina: This is more technically accurate and matches with `JobResult.toString()`...

cangencer:
I think we should always set a name for the job, if the name is missing on job submission, it could be set to the id string representation automatically. This should simplify a lot of the logic

PR #941, opened 2018-06-26
Add executionId to job metric names

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
```diff
@@ -172,7 +171,8 @@ public void initialize(NodeEngine nodeEngine, long jobId, long executionId, Snap
 
                 ProbeBuilder probeBuilder = this.nodeEngine.getMetricsRegistry().newProbeBuilder()
                         .withTag("module", "jet")
-                        .withTag("job", idToString(jobId))
+                        .withTag("job", Long.toHexString(jobId))
```
cangencer: it would be better to use idToString for readability and consistency with logs. Saving a few characters doesn't matter in this case.

PR #944, opened 2018-06-27
Keyed mapUsingContext operations should take key as one of the parameters

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/TriPredicate.java
```diff
@@ -0,0 +1,38 @@
...
+package com.hazelcast.jet.function;
+
+import java.util.function.Predicate;
+
+/**
+ * Represents a predicate which accepts three arguments. This
+ * is the three-arity specialization of {@link Predicate}.
+ **/
+@FunctionalInterface
+public interface TriPredicate<T, U, V> {
```
mtopolnik: `TriPredicate` is now lacking the composition methods.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/DistributedTriPredicate.java
```diff
@@ -0,0 +1,53 @@
...
+ * {@code Serializable} variant of {@link TriPredicate}.
+ */
+public interface DistributedTriPredicate<T, U, V> extends TriPredicate<T, U, V>, Serializable {
+
+    /**
+     * Returns a composite predicate which evaluates the
+     * equivalent of {@code this.test(t, u, v) && other.test(t, u, v)}.
+     */
+    default DistributedTriPredicate<T, U, V> and(DistributedTriPredicate<? super T, ? super U, ? super V> other) {
+        checkNotNull(other, "other");
```
mtopolnik: It should also `checkSerializable(other)`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/DistributedTriPredicate.java
```diff
@@ -0,0 +1,53 @@
...
+import static com.hazelcast.util.Preconditions.checkNotNull;
+
+/**
+ * {@code Serializable} variant of {@link TriPredicate}.
+ */
+public interface DistributedTriPredicate<T, U, V> extends TriPredicate<T, U, V>, Serializable {
+
+    /**
+     * Returns a composite predicate which evaluates the
+     * equivalent of {@code this.test(t, u, v) && other.test(t, u, v)}.
```
mtopolnik: Add "The `other` predicate must not be null".

hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/DistributedTriPredicate.java
viliam-durina: I don't think so, the method taking the resulting predicate should.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/DistributedTriPredicate.java
cangencer: better to add `@Nonnull` annotations I think

mtopolnik: Indeed :)

PR #946, opened 2018-06-28
Elasticity

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
```diff
@@ -81,33 +81,73 @@ default String getName() {
...
 
+    /**
+     * Gracefully suspend the current execution. Job status will become {@link
+     * JobStatus#SUSPENDED}. To resume the job, call {@link #resume()}.
+     *
+     * <p>If the job does not do {@linkplain JobConfig#setProcessingGuarantee
+     * state snapshots}, it will be suspended anyway. When resumed, it will
+     * start with an empty state.
+     *
+     * <p>This call returns quickly and a suspension process is initiated. This
```
metanet: maybe we can explicitly say: `This call returns quickly and a suspension process is initiated in the background`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -681,6 +805,67 @@ private static ClassLoader swapContextClassLoader(ClassLoader jobClassLoader) {
         return contextClassLoader;
     }
 
+    void resumeJob(Function<Long, Long> executionIdSupplier) {
+        synchronized (lock) {
+            if (jobStatus.compareAndSet(SUSPENDED, NOT_STARTED)) {
```
metanet: should we print a debug log here for the "else" branch?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/TerminationMode.java
```diff
@@ -0,0 +1,112 @@
...
+     * If true, job resources and snapshots should be deleted after
+     * termination. Otherwise the job will remain ready to be restarted. It's
+     * true only for cancellation.
+     */
+    public boolean isDeleteData() {
+        return deleteData;
+    }
+
+    @CheckReturnValue
+    public TerminationMode withoutStopWithSnapshot() {
```
metanet: should we rename this method to: `withoutTerminalSnapshot`

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/CancellationTest.java
```diff
@@ -297,6 +297,7 @@ private static void assertExecutionTerminated() {
                 long last = previous[0];
                 previous[0] = current;
                 assertTrue("Call counter should eventually stop being incremented.", current == last);
+                sleepMillis(200);
```
metanet: I don't think we need a `sleep` call here. `assertTrueEventually()` already sleeps internally.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/GracefulShutdownTest.java
```diff
@@ -0,0 +1,393 @@
...
+            System.out.println("aaa");
+            sleepMillis(200);
+        }
+
+        // after that, the job should eventually complete
+        assertTrueEventually(() -> assertTrue(jobFuture.isDone()), 5);
+    }
+
+    @Test
+    @Repeat(10)
```
metanet: There is a left-over `@Repeat` here.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/SplitBrainTest.java
```diff
@@ -81,7 +76,7 @@ public void when_quorumIsLostOnMinority_then_jobRestartsUntilMerge() {
             MockPS processorSupplier = new MockPS(StuckProcessor::new, clusterSize);
             DAG dag = new DAG().vertex(new Vertex("test", processorSupplier));
             jobRef[0] = instances[0].newJob(dag, new JobConfig().setSplitBrainProtection(true));
-            assertOpenEventually(StuckProcessor.executionStarted);
+            assertOpenEventually(StuckProcessor.executionStarted, 10);
```
metanet: why do we use a 10-sec hard limit for eventual assertions here? It can be quite problematic when the pr builder slows down. The default timeout must be just fine.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -394,127 +449,178 @@ private void invokeStartExecution() {
...
 
         Map<Boolean, List<Entry<MemberInfo, Object>>> grouped = groupResponses(responses);
         Collection<MemberInfo> successfulMembers = grouped.get(false).stream().map(Entry::getKey).collect(toList());
 
+        // there was no user exception. If the termination was requested, return appropriate exception
         if (successfulMembers.size() == executionPlanMap.size()) {
-            logger.fine("Execute of " + jobIdString() + " is successful.");
+            logger.fine(opName + " of " + jobIdString() + " was successful");
+
+            // TODO [viliam] This is a race:
```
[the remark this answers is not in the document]
viliam-durina: No, the race is still there. I'm not fixing it now, but we might do it some day...

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/CancellationTest.java
viliam-durina: That's true, but it's not part of its contract. We require longer pauses for correct functionality, so why not do them ourselves?

viliam-durina:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/DAG.java
```diff
@@ -390,11 +389,6 @@ public String toDotString() {
         Pattern stepPattern = Pattern.compile("(?<stepName>.+)-step[12]");
         int clusterCount = 0;
         for (Vertex v : this) {
-            // skip snapshot vertices
```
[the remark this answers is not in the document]
viliam-durina: It's no longer needed, we convert to DOT string before adding the restore vertices. And there was a bug in it: if there was just one vertex, with snapshot-restore vertices it didn't go into the "dangling vertex" branch, so I fixed it this way.

PR #950, opened 2018-07-02
Add JetClientConfig which uses the correct group name and password

cangencer:
Currently when you create a new `ClientConfig` object it doesn't have the correct group name and password to connect to the Jet cluster. This class is a convenience for overriding the default group name and password when creating a client config to connect to a Jet cluster.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JetClientConfig.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JetClientConfig.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
+ *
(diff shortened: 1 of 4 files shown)


cangencer:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JetClientConfig.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.hazelcast.client.config.ClientConfig;
+
+/**
+ * A config object which is used for configuring the Jet client to
+ * connect to the Jet cluster.
+ * <p>
+ * See {@link ClientConfig} for actual config options.
+ */
+public class JetClientConfig extends ClientConfig {
```
viliam-durina: This class is unused.

PR #967, opened 2018-07-13
bufferFn should be created once per processor

cangencer:
Causes mapWithMerging and mapWithUpdating to fail with parallelism > 1

Test will follow later.


viliam-durina:
The problem is in mapWithUpdating, they should not have stateful lambdas. We should fix those.

PR #976, opened 2018-07-23
Do not tag snapshot restore vertices with source=yes

cangencer:
Additional do not print them when restarting the job


hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/DotTest.java
```diff
@@ -76,6 +77,22 @@ public void when_pipelineToDotString() {
         System.out.println(p.toDag().toDotString());
     }
 
+    @Test
+    public void when_snapshotVertex_then_skip() {
```
viliam-durina: This test actually shows the name-clash...

PR #991, opened 2018-08-14
Fix race in AsyncMapWriter

### LINKED ISSUE #975: com.hazelcast.jet.impl.util.AsyncMapWriterTest.when_writeErrorOnlyOnce_then_retrySuccess
```
Error Message
expected:<0> but was:<null>
Stacktrace
java.lang.AssertionError: expected:<0> but was:<null>
	at com.hazelcast.jet.impl.util.AsyncMapWriterTest.when_writeErrorOnlyOnce_then_retrySuccess(AsyncMapWriterTest.java:260)
```

https://hazelcast-l337.ci.cloudbees.com/view/Jet/job/jet/com.hazelcast.jet$hazelcast-jet-core/825/testReport/junit/com.hazelcast.jet.impl.util/AsyncMapWriterTest/when_writeErrorOnlyOnce_then_retrySuccess/

viliam-durina:
The writer uses `PartitionIteratingOperation`, which can partly fail:
some partitions succeed, some fail. The failed partitions are retried.
The retrial is executed from an `ExecutionCallback`, which is called
once for each member. Both attempts use `invokeOnCluster`, which used a
`doneLatch` to track callbacks from members and when it got to 0, it
completed the future. The problem was if in the first attempt, two
members failed. It will cause two retry calls to `invokeOnCluster`. But
the first completed retry call completed the future before the second
one finishes, causing the test to (correctly) fail.

The solution is to use one `doneLatch` for both the initial attempt and
all retries. `invokeOnCluster` only increments and decrements it. If a
partition is retried, it is first incremented by the number of retry
operations so that it gets to 0 only when all retrials are successfully
done.

The test failed in about 0.1% of runs.

Fixes #975

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/AsyncMapWriter.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/AsyncMapWriter.java
@@ -119,7 +119,7 @@ public class AsyncMapWriter {
      */
     public boolean tryFlushAsync(CompletableFuture<Void> completionFuture) {
         Map<Address, List<Integer>> memberPartitionsMap = partitionService.getMemberPartitionsMap();
-        AtomicInteger doneLatch = new AtomicInteger(0);
+        AtomicInteger pendingOps = new AtomicInteger(0);
         List<PartitionOpBuilder> ops = memberPartitionsMap.entrySet()
                                                           .stream()
                                                           .map(e -> opForMember(e.getKey(), e.getValue(), outputBuffers))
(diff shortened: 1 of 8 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/AsyncMapWriter.java
```diff
@@ -230,15 +234,17 @@ private boolean invokeOnCluster(List<PartitionOpBuilder> opBuilders,
...
                     executionService.schedule(() -> {
                         try {
-                            if (!tryRetry(partitions, entries, completionFuture)) {
+                            if (!tryRetry(partitions, entries, doneLatch, completionFuture)) {
                                 completionFuture.completeExceptionally(originalErr);
+                                return;
                             }
-
+                            int value = doneLatch.decrementAndGet();
+                            assert value > 0 : "value=" + value;
```
cangencer: couldn't value be 0 here if the retry completes before the assert?

viliam-durina: yes, that's true. But then I also need to do the same action as the other place if the count gets to 0. I'll fix it.

Issue #988, opened 2018-08-14
@SpringAware annotation not processed for sinks


PeeWee2201:
When a custom sink annotated by @SpringAware is added to the pipeline, it is wrapped in an AdaptingProcessor. Therefore, the @SpringAware annotation is not seen and autowiring is not performed.

Sink can be defined as:

```
@SpringAware
public class WriteSink extends AbstractProcessor
{
   public WriteNotificationSink()
   {
      setCooperative(false);
   }

   @Override
   protected boolean tryProcess(int oridinal, Object item)
   {
      Assert.assertNotNull(mDatabase); // This will fail.
   }

   @Resource
   private Database mDatabase;

   private static class Supplier implements ProcessorSupplier
   {
      @Override
      public Collection<? extends Processor> get(int count)
      {
         return Stream.generate(() -> new WriteSink()).limit(count).collect(Collectors.toList());
      }
   }

   private static final class MetaSupplier implements ProcessorMetaSupplier
   {
      static final long serialVersionUID = 1L;

      private MetaSupplier()
      {
      }

      @Override
      public Function<Address, ProcessorSupplier> get(List<Address> addresses)
      {
         return address -> new Supplier();
      }
   }

   public static Sink<Object> notificationSink()
   {
      return new SinkImpl<>("writeSink", new MetaSupplier());
   }
}
```

And used as:

```
List<String> text = jet.getList("text");
text.add("One");
text.add("Two");
Pipeline p = Pipeline.create();
p.drawFrom(Sources.<String>list("text")).drainTo(writeSink());
jet.newJob(p);
```

PR #999, opened 2018-08-16
Inject spring resources to wrapped processors

gurbuzali:
fixes https://github.com/hazelcast/hazelcast-jet/issues/988

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
@@ -25,6 +25,7 @@ import com.hazelcast.jet.config.ProcessingGuarantee;
 import com.hazelcast.jet.core.Processor;
 import com.hazelcast.jet.core.Processor.Context;
 import com.hazelcast.jet.core.Watermark;
+import com.hazelcast.jet.impl.processor.ProcessorWrapper;
 import com.hazelcast.jet.impl.util.ArrayDequeInbox;
 import com.hazelcast.jet.impl.util.CircularListCursor;
 import com.hazelcast.jet.impl.util.ProgressState;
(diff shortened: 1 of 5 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/ProcessorWrapper.java
```diff
@@ -28,12 +28,17 @@
  * processor.
  */
 public class ProcessorWrapper implements Processor {
```
cangencer: should also be abstract class

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
```diff
@@ -190,8 +191,12 @@ private OutboxImpl createOutbox(@Nonnull OutboundCollector ssCollector) {
     @Override
     public void init() {
         if (serializationService.getManagedContext() != null) {
-            Object processor2 = serializationService.getManagedContext().initialize(processor);
-            assert processor2 == processor : "different object returned";
+            Processor toInit = processor;
```
cangencer: can be replaced with `? :` expression

hazelcast-jet-spring/src/test/java/com/hazelcast/jet/spring/TestCustomProcessor.java
```diff
@@ -0,0 +1,99 @@
...
+    @BeforeClass
+    @AfterClass
+    public static void start() {
+        Jet.shutdownAll();
+    }
+
+    @Test
+    public void test() {
+        Pipeline p = Pipeline.create();
+        p.drawFrom(Sources.batchFromProcessor("source", preferLocalParallelismOne(CustomSourceP::new)))
```
cangencer: is this test actually testing the wrapping?

gurbuzali: not the source but the sink is wrapped

cangencer: OK, I see the sink is always wrapped.

PR #1002, opened 2018-08-17
Don't store JobResults forever

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Jet.java
```diff
@@ -54,6 +56,8 @@
      */
     public static final String INTERNAL_JET_OBJECTS_PREFIX = "__jet.";
 
+    private static final int JOB_RESULTS_TTL_SECONDS = (int) DAYS.toSeconds(7);
```
cangencer: can we make this a group property instead?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Jet.java
```diff
@@ -130,6 +134,9 @@ static void configureJetService(JetConfig jetConfig) {
                         .setStatisticsEnabled(false)
                         .setMergePolicyConfig(
                                 new MergePolicyConfig().setPolicy(IgnoreMergingEntryMapMergePolicy.class.getName()))
+                )
+                .addMapConfig(new MapConfig(JOB_RESULTS_MAP_NAME)
```
cangencer: will this config take precedence over the wildcard?

viliam-durina: Exact match always takes precedence.

PR #1007, opened 2018-08-20
Test fixes

viliam-durina:


hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/TopologyChangeTest.java
```diff
@@ -427,7 +427,7 @@ public void when_nodeIsShuttingDownDuringInit_then_jobRestarts() {
 
         assertTrueEventually(() -> assertFalse(jetService.getJobCoordinationService().getMasterContexts().isEmpty()));
 
-        spawn(instances[2]::shutdown);
+        spawn(() -> instances[2].getHazelcastInstance().shutdown());
```
cangencer: how does this fix the test?

viliam-durina: It uses the forceful shutdown instead of graceful. With graceful, there is a race: The job is not restarted but oscillates between STARTING and NOT_RUNNING. With forceful, it becomes STARTING and remains so, because the InitExecutionOp is rejected, so it should be stable.

PR #1011, opened 2018-08-21
Elasticity fixes

viliam-durina:
Contains:
- a snapshot could be begun if job was not running (due to upscaling...)
- if we fail to restart to scale-up, we'll schedule a retry
- we called `handleTermination` even if job was not running

The effect of the above was minor, either the 
`CompleteExecutionOperation` or `SnapshotOperation` were rejected by the 
target, which polluted the logs, but no real harm was done.


viliam-durina:
verify

PR #1025, opened 2018-08-30
Update IMDG to 3.11

viliam-durina:


viliam-durina:
Should compile after hazelcast/hazelcast#13636.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetService.java
```diff
@@ -243,8 +244,12 @@ public ClassLoader getClassLoader(long jobId) {
     }
 
     @Override
-    public void handle(Packet packet) throws IOException {
-        networking.handle(packet);
+    public void accept(Packet packet) {
+        try {
```
cangencer: we could use `uncheckedRun()`

viliam-durina: It's quite a hot path here, `uncheckRun` will create one throwaway lambda for each packet, I wouldn't do it.

PR #1031, opened 2018-09-03
Remove dead code

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -741,7 +741,7 @@ void finalizeJob(@Nullable Throwable failure) {
                 jobRecord = jobRecord.withSuspended(true);
                 nonSynchronizedAction = () -> coordinationService.suspendJob(this);
             } else if (terminationModeAction == NO_ACTION) {
-                jobStatus = NOT_RUNNING;
+                throw new UnsupportedOperationException("Should not have received " + terminationModeAction + " here");
```
cangencer: would it better to change this to a switch statement? I found the logic rather confusing here.

viliam-durina:
I have better PR that improves this.

PR #1038, opened 2018-09-07
Don't throw HzRetryableExc from terminateJob

viliam-durina:
If JobRecord exists, we'll eventually learn of it, but otherwise there's
no point in retrying.

Also includes test for calling methods on stale Job instance.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
@@ -409,7 +409,7 @@ public class JobCoordinationService {
                 // we'll eventually learn of the job through scanning of records or from a join operation
                 throw new RetryableHazelcastException(message);
             } else {
-                throw new JobNotFoundException(message);
+                throw new JobNotFoundException(jobId);
             }
         }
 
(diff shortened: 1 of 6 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -403,8 +403,14 @@ public void terminateJob(long jobId, TerminationMode terminationMode) {
...
         if (masterContext == null) {
-            throw new RetryableHazelcastException("No MasterContext found for job " + idToString(jobId)
-                    + " for " + terminationMode);
+            JobRecord jobRecord = jobRepository.getJobRecord(jobId);
+            String message = "No MasterContext found for job " + idToString(jobId) + " for " + terminationMode;
+            if (jobRecord != null) {
+                // we'll eventually learn of the job through scanning of records or from a join operation
+                throw new RetryableHazelcastException(message);
+            } else {
+                throw new JobNotFoundException(message);
```
cangencer: the message doesn't seem right here, it's not that there is no master context but the record doesn't exist either?

viliam-durina: It is correct. MasterContext wasn't found, but we throw different type of exception based on whether the JobRecord exists or not.

viliam-durina: Thinking of it more, you're right: the error will be reported to the user. He has no idea what MasterContext is. I'll update it to the default message.

PR #1040, opened 2018-09-07
Add jet shutdown hook, disable imdg one

gurbuzali:
By default, imdg adds its own shutdown hook, which calls either shutdown or terminate according to the shutdown-hook-policy. We've introduced graceful shutdown of jobs but it is only called if explicitly called `JetInstance.shutdown()`. If the processor is killed externally jobs are not shutdown gracefully.

With this PR, we disable the imdg shutdown hook and add our own hook. In the hook, if the policy is graceful then the jobs and underlying imdg instance are shutdown gracefully otherwise we terminate the imdg instance without shutting down the jobs.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Jet.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Jet.java
@@ -68,9 +68,8 @@ public final class Jet {
         configureJetService(config);
         HazelcastInstanceImpl hazelcastInstance = ((HazelcastInstanceProxy)
                 Hazelcast.newHazelcastInstance(config.getHazelcastConfig())).getOriginal();
-        JetInstanceImpl jetInstance = new JetInstanceImpl(hazelcastInstance, config);
-        jetInstance.registerShutdownHook();
-        return jetInstance;
+        JetService service = hazelcastInstance.node.nodeEngine.getService(JetService.SERVICE_NAME);
+        return service.getJetInstance();
     }
 
     /**
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Jet.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Jet.java
@@ -117,9 +117,12 @@ public final class Jet {
 
         Properties jetProps = jetConfig.getProperties();
         Properties hzProperties = hzConfig.getProperties();
+
+        // copy Jet Config properties as HZ properties
         for (String prop : jetProps.stringPropertyNames()) {
             hzProperties.setProperty(prop, jetProps.getProperty(prop));
         }
+
         HazelcastProperties properties = new HazelcastProperties(hzProperties);
 
         ServicesConfig servicesConfig = hzConfig.getServicesConfig();
@@ -127,7 +130,7 @@ public final class Jet {
                 .addServiceConfig(new ServiceConfig().setEnabled(true)
                         .setName(JetService.SERVICE_NAME)
                         .setClassName(JetService.class.getName())
-                        .setProperties(jetServiceProperties(properties.getString(SHUTDOWNHOOK_ENABLED)))
+                        .setProperties(jetServiceProperties(properties))
                         .setConfigObject(jetConfig));
 
         servicesConfig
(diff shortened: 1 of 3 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Jet.java
```diff
@@ -67,6 +71,7 @@ public static JetInstance newJetInstance(JetConfig config) {
         configureJetService(config);
         HazelcastInstanceImpl hazelcastInstance = ((HazelcastInstanceProxy)
                 Hazelcast.newHazelcastInstance(config.getHazelcastConfig())).getOriginal();
+        Runtime.getRuntime().addShutdownHook(shutdownHookThread(hazelcastInstance));
```
viliam-durina: We should also remove the shutdown hook.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Jet.java
```diff
@@ -68,9 +68,8 @@ public static JetInstance newJetInstance(JetConfig config) {
         configureJetService(config);
         HazelcastInstanceImpl hazelcastInstance = ((HazelcastInstanceProxy)
                 Hazelcast.newHazelcastInstance(config.getHazelcastConfig())).getOriginal();
-        JetInstanceImpl jetInstance = new JetInstanceImpl(hazelcastInstance, config);
-        jetInstance.registerShutdownHook();
-        return jetInstance;
+        JetService service = hazelcastInstance.node.nodeEngine.getService(JetService.SERVICE_NAME);
```
cangencer: you can also use `Util.getJetInstance`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetInstanceImpl.java
```diff
@@ -51,6 +51,7 @@ public JetInstanceImpl(HazelcastInstanceImpl hazelcastInstance, JetConfig config
         this.nodeEngine = hazelcastInstance.node.getNodeEngine();
         this.config = config;
         this.shutdownHookThread = shutdownHookThread(nodeEngine);
+        Runtime.getRuntime().addShutdownHook(shutdownHookThread);
```
cangencer: the hook shouldn't be created here as JetInstanceImpl is just a wrapper and can be created several times. It should be created in JetService instead.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Jet.java
```diff
@@ -150,5 +151,6 @@ static void configureJetService(JetConfig jetConfig) {
         for (String prop : jetProps.stringPropertyNames()) {
             hzConfig.getProperties().setProperty(prop, jetProps.getProperty(prop));
         }
+        hzConfig.setProperty(GroupProperty.SHUTDOWNHOOK_ENABLED.getName(), "false");
```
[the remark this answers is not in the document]
gurbuzali: should we introduce a jet-specific property then ? Or we can check the `SHUTDOWNHOOK_ENABLED` property if it is false, do nothing if it is true add our own hook and set it to false to disable imdg hook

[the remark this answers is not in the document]
cangencer: I think we'll need to disable the original one somehow, and copy the original setting to another property and check that one during startup.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetService.java
```diff
@@ -51,21 +52,25 @@
...
+import static com.hazelcast.spi.properties.GroupProperty.SHUTDOWNHOOK_POLICY;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 public class JetService
         implements ManagedService, ConfigurableService<JetConfig>, PacketHandler, MembershipAwareService,
         LiveOperationsTracker {
 
+    public static final String SERVICE_NAME = "hz:impl:jetService";
     public static final int MAX_PARALLEL_ASYNC_OPS = 1000;
+    public static final HazelcastProperty JET_SHUTDOWN_HOOK_ENABLED =
```
cangencer: can we move this `GroupProperty` ?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetService.java
```diff
@@ -279,4 +291,15 @@ public void populate(LiveOperations liveOperations) {
         }
         return keys;
     }
+
+    private Thread shutdownHookThread(NodeEngine nodeEngine) {
+        return new Thread(() -> {
```
cangencer: would be good to give the thread a name

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetService.java
gurbuzali: moved it. I was not aware we have jet specific group property class

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetService.java
gurbuzali: "jet.ShutdownThread" ? imdg has "hz.ShutdownThread"

cangencer: Yes that's fine.

PR #1062, opened 2018-09-22
Convert ReadMetricsOperation from blocking to async operation

gurbuzali:
ReadMetricsOperation was using wait/notify mechanism but it is not thread-safe, it is designed for partition-thread operations. With the change, `readMetrics` method returns a future which is used to return the response once a new metric arrived.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/JetMetricsService.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/JetMetricsService.java
@@ -45,6 +45,7 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
 import static com.hazelcast.jet.Util.entry;
+import static com.hazelcast.jet.impl.util.ExceptionUtil.withTryCatch;
 import static java.util.stream.Collectors.joining;
 
 /**
@@ -60,7 +61,7 @@ public class JetMetricsService implements ManagedService, ConfigurableService<Me
     private final LiveOperationRegistry liveOperationRegistry;
     // Holds futures for pending read metrics operations
     private final ConcurrentMap<CompletableFuture<RingbufferSlice<Map.Entry<Long, byte[]>>>, Long>
-            futureMap = new ConcurrentHashMap<>();
+            pendingReads = new ConcurrentHashMap<>();
 
     /**
      * Ringbuffer which stores a bounded history of metrics. For each round of collection,
@@ -133,15 +134,15 @@ public class JetMetricsService implements ManagedService, ConfigurableService<Me
             throw new IllegalArgumentException("Metrics collection is not enabled");
         }
         CompletableFuture<RingbufferSlice<Map.Entry<Long, byte[]>>> future = new CompletableFuture<>();
-        future.whenComplete((s, e) -> futureMap.remove(future));
-        futureMap.put(future, startSequence);
+        future.whenComplete(withTryCatch(logger, (s, e) -> pendingReads.remove(future)));
+        pendingReads.put(future, startSequence);
 
-        readFromJournal(future, startSequence);
+        tryCompleteRead(future, startSequence);
 
         return future;
     }
 
-    private void readFromJournal(CompletableFuture<RingbufferSlice<Map.Entry<Long, byte[]>>> future, long sequence) {
+    private void tryCompleteRead(CompletableFuture<RingbufferSlice<Map.Entry<Long, byte[]>>> future, long sequence) {
         try {
             RingbufferSlice<Map.Entry<Long, byte[]>> slice = metricsJournal.copyFrom(sequence);
             if (!slice.isEmpty()) {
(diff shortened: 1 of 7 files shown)


gurbuzali:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/management/ReadMetricsOperation.java
```diff
@@ -16,22 +16,20 @@
...
-import com.hazelcast.spi.WaitNotifyKey;
 
 import java.util.Map.Entry;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 
-public class ReadMetricsOperation extends Operation implements BlockingOperation, ReadonlyOperation {
+public class ReadMetricsOperation extends Operation implements ReadonlyOperation {
 
     private static final int MIN_TIMEOUT_SECONDS = 5;
```
eminn: I think we need a decision to keep timeout on the operation or not ?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/JetMetricsService.java
```diff
@@ -111,21 +116,41 @@ public void init(NodeEngine nodeEngine, Properties properties) {
...
      * Read metrics from the journal from the given sequence
      */
-    public ConcurrentArrayRingbuffer.RingbufferSlice<Map.Entry<Long, byte[]>> readMetrics(long startSequence) {
+    public CompletableFuture<RingbufferSlice<Map.Entry<Long, byte[]>>> readMetrics(long startSequence) {
         if (!config.isEnabled()) {
             throw new IllegalArgumentException("Metrics collection is not enabled");
         }
-        return metricsJournal.copyFrom(startSequence);
+        CompletableFuture<RingbufferSlice<Map.Entry<Long, byte[]>>> future = new CompletableFuture<>();
+        future.whenComplete((s, e) -> futureMap.remove(future));
```
cangencer: it's better to use `ExceptionUtil.withTryCatch` for all whenComplete calls, otherwise any exception gets swallowed.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/JetMetricsService.java
```diff
@@ -47,16 +51,16 @@
...
     private final NodeEngineImpl nodeEngine;
     private final ILogger logger;
-
-    // keys used for synchronization of read operation
-    private final MetricsNotifyKey notifyKey = new MetricsNotifyKey();
-    private final Notifier notifier = new MetricsNotifier();
+    private final LiveOperationRegistry liveOperationRegistry;
+    // Holds futures for pending read metrics operations
+    private final ConcurrentMap<CompletableFuture<RingbufferSlice<Map.Entry<Long, byte[]>>>, Long>
+            futureMap = new ConcurrentHashMap<>();
```
cangencer: let's call this `pendingReads`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/management/ReadMetricsOperation.java
cangencer: yes timeout should be removed

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/management/ReadMetricsOperation.java
```diff
@@ -40,31 +38,39 @@ public ReadMetricsOperation(long offset, int collectionIntervalSeconds) {
...
+        service.getLiveOperationRegistry().register(this);
     }
 
     @Override
-    public String getServiceName() {
-        return JetMetricsService.SERVICE_NAME;
+    public void run() {
+        JetMetricsService service = getService();
+        CompletableFuture<RingbufferSlice<Entry<Long, byte[]>>> future = service.readMetrics(offset);
+        future.whenComplete((slice, error) -> doSendResponse(error != null ? error : slice));
```
cangencer: see previous comment about catch exceptions in whenComplete calls.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/management/ReadMetricsOperation.java
```diff
@@ -40,31 +38,39 @@ public ReadMetricsOperation(long offset, int collectionIntervalSeconds) {
...
     }
 
     @Override
-    public void onWaitExpire() {
-        sendResponse(new OperationTimeoutException("No metrics were retrieved for " + getWaitTimeout() + "ms"));
+    public String getServiceName() {
+        return JetMetricsService.SERVICE_NAME;
     }
 
+    final void doSendResponse(Object value) {
```
cangencer: seems duplicate of AsyncJobOperation , but I see that also extends AbstractJobOperation so probably it's not possible to refactor it cleanly.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/JetMetricsService.java
```diff
@@ -111,21 +116,41 @@ public void init(NodeEngine nodeEngine, Properties properties) {
...
+
+        return future;
     }
 
-    /**
-     * key used for signalling pending read operations
-     */
-    public WaitNotifyKey waitNotifyKey() {
-        return notifyKey;
+    private void readFromJournal(CompletableFuture<RingbufferSlice<Map.Entry<Long, byte[]>>> future, long sequence) {
```
cangencer: method should be called "tryCompleteRead" or something similar I think

PR #1065, opened 2018-09-24
Properly Adapt Rolling Aggregation to JetEvent

mtopolnik:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/Processors.java
```diff
@@ -804,34 +809,32 @@ private Processors() {
...
-     * If the mapping result is {@code null}, the vertex emits nothing.
-     * Therefore it can be used to implement filtering semantics as well.
+     * If the result after applying `mapToOutputFn` is {@code null}, the vertex
+     * emits nothing. Therefore it can be used to implement filtering semantics
+     * as well.
      * <p>
-     * This vertex saves the state to snapshot so the context objects will
-     * survive a job restart.
-     *  @param <T> type of the input item
+     * This vertex saves the state to snapshot so the context objects will survive
```
cangencer: instead of context objects we should say the accumulators/state

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/FunctionAdapter.java
```diff
@@ -109,30 +109,42 @@
 
     @SuppressWarnings("unchecked")
     public <T, T1, R> DistributedBiFunction<?, ? super T1, ?> adaptHashJoinOutputFn(
```
cangencer: we could add `@Nonnull` to output values as well

PR #1066, opened 2018-09-24
Add serializer for DONE_ITEM, collapse BroadcastKey

cangencer:


hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/execution/WatermarkCoalescer_IntegrationTest.java
```diff
@@ -307,7 +308,7 @@ public boolean complete() {
                     nextItemAt = System.nanoTime() + MILLISECONDS.toNanos(((Delay) item).millis);
                     pos++;
                     return false;
-                } else if (item == DONE_ITEM) {
+                } else if (item.equals(DONE_ITEM)) {
```
viliam-durina: I'd rename this to `DONE_ITEM_STR` or use the `"DONE_ITEM"` string directly, as this is likely to be mistaken for the actual DONE_ITEM.

PR #1083, opened 2018-09-28
Traverser.over -> Traversers.traverseItems

### LINKED ISSUE #902: Confusing default impl of Traverses
We have `static <T> Traverser<T> over(T... items)` directly in the `Traverser` interface and then a bunch of other impls in `Traversers` 

I found it confusing. It creates an impression `over(array)` is the only impl we ship with Jet. 

For a reference:
image

mtopolnik:
Fixes #902

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Traverser.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Traverser.java
@@ -221,14 +221,4 @@ public interface Traverser<T> {
         };
     }
 
-    /**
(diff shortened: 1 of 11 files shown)


viliam-durina:
`traverseArray` and `traverseItems` are almost the same. Can we convert `traverseArray` to a vararg version?

mtopolnik:
I would prefer to keep them separate. You may want to traverse over an array as a single item, then you'll get the usual trouble with vararg ambiguity.

mtopolnik:
Change-following samples PR:

https://github.com/hazelcast/hazelcast-jet-code-samples/pull/142

PR #1091, opened 2018-10-01
Remove watermarks from a pipeline with no windowing stages

### LINKED ISSUE #1063: Timestamped stream with no windowing does noThrottling
If you have a timestamped pipeline which doesn't do any windowing defaults to `noThrottling()` on the watermarks. This can cause generation of  large amount of WMs which do not serve any purpose.

mtopolnik:
Fixes #1063.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/Planner.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/Planner.java
@@ -43,7 +43,7 @@ import java.util.function.Consumer;
 import static com.hazelcast.jet.core.Edge.from;
 import static com.hazelcast.jet.core.SlidingWindowPolicy.tumblingWinPolicy;
 import static com.hazelcast.jet.core.WatermarkEmissionPolicy.emitByFrame;
-import static com.hazelcast.jet.core.WatermarkEmissionPolicy.noThrottling;
+import static com.hazelcast.jet.core.WatermarkEmissionPolicy.noWatermarks;
 import static com.hazelcast.jet.impl.TopologicalSorter.topologicalSort;
 import static java.util.stream.Collectors.toList;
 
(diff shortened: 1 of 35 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/Planner.java
```diff
@@ -67,15 +67,15 @@ DAG createDag() {
...
+                                                 .map(Transform::preferredWatermarkStride)
                                                  .filter(frameSize -> frameSize > 0)
                                                  .mapToLong(i -> i)
                                                  .toArray());
         // Update watermark emission policy on all wm gen params
         // with the GCD frame length
         WatermarkEmissionPolicy emitPolicy = frameSizeGcd > 0
                 ? emitByFrame(tumblingWinPolicy(frameSizeGcd))
-                : noThrottling();
+                : noWatermarks();
```
cangencer: we already have WatermarkGenerationParams.noWatermarks() , this seems like a duplication?

mtopolnik: This code was simpler with it, but now I'm replacing the entire `wmParams` instead.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/SessionWindowDef.java
```diff
@@ -28,7 +28,7 @@
  * @param <T> type of the stream item
  */
 public class SessionWindowDef<T> implements WindowDefinition {
-    private static final int MAX_FRAME_RATE = 100;
+    private static final int MAX_FRAME_SIZE = 100;
```
viliam-durina: It should be `MAX_WATERMARK_STRIDE`

PR #1102, opened 2018-10-04
Make single SnapshotRestore input higher priority

### LINKED ISSUE #1100: Snapshot restore edge to source processor is not higher priority
Snapshot is not started until all higher-priority edges are completed. Snapshot restoring is done using such edges. However, in the case of source, the restore edge isn't higher priority, because it's the only edge. It can happen that the first snapshot is done before the state of the source is restored.

It's very hard to run into this issue because:
- snapshot interval must be shorter than it takes to restore the snapshot
- all other vertices except the source one must be already restored

If the state is small, it's quick to restore. If the state is large, it's typically not the source which has large state.

viliam-durina:
Snapshot is not started until all higher-priority edges are completed.
Snapshot restoring is done using such edges. However, in the case of
source, the restore edge wasn't higher priority, because it's the only
edge. It can happen that the first snapshot is done before the state of
the source is restored.

Fixes #1100


viliam-durina:
Tests fail because issue fixed in #1101 now reproduces here. Tests should pass after #1101 is merged and this branch is rebased.

PR #1104, opened 2018-10-10
Reuse snapshot maps

viliam-durina:
Changes the mode of operation where we created new maps for each
snapshot and destroyed the maps when deleting the snapshot to a mode
where we have 2 maps and use one for writing and one for storing.

Notes:
- responsibility for updating JobRecord moved from JobCoordinationService
to MasterContext

- MasterContext contains final field with JobRecord. It mutates it
(therefore some fields in JobRecord were made volatile) and writes it
out after each change using an async call. Since it can be mutated in
parallel and parallel async writes can be reordered, we use timestamp to
ensure that an older update doesn't overwrite a newer one.

- One snapshot is stored in one map (previously, each vertex used its
own map). The chunk key now contains vertex name and snapshotId.
SnapshotId was added to it to distinguish zombie writes, should they
happen, or writing after a failed `IMap.clear`. When reading, unknown
vertices and snapshotIds are logged and ignored.

- Which data map is used for new snapshot is communicated to members
using SnapshotOp. Because target member must know this before it starts
writing, we've removed the (rare) possibility that snapshot is written
before the operation is received.


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/SnapshotContext.java
```diff
@@ -143,7 +166,8 @@ synchronized void initTaskletCount(int totalTaskletCount, int highPriorityTaskle
      * {@code SnapshotOperation} and send barriers to such processor before
      * the {@code SnapshotOperation} is executed on this member.
      */
-    synchronized CompletableFuture<SnapshotOperationResult> startNewSnapshot(long snapshotId, boolean isTerminal) {
+    synchronized CompletableFuture<SnapshotOperationResult> startNewSnapshot(
+            long snapshotId, int dataMapIndex, boolean isTerminal) {
```
cangencer: since snapshot can't start anymore just because of barrier from remote member we should remove the comment and also simplify the related logic

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/AsyncSnapshotWriterImpl.java
```diff
@@ -242,19 +245,28 @@ private boolean putAsyncToMap(int partitionId, Supplier<Data> dataSupplier) {
...
+     * offered until new snapshot is seen in {@link #snapshotContext}.
      *
      * @return {@code true} on success, {@code false} if we weren't able to
      * flush some partitions due to the limit on number of parallel async ops.
      */
     @Override
     @CheckReturnValue
-    public boolean flush() {
+    public boolean flushAndReset() {
         for (int i = 0; i < buffers.length; i++) {
```
cangencer: I think we should also return false here if map name is not set, since flush() can be called without any data.

viliam-durina: No, `flush` returns false if you need to call flush again. If no map was set, we're fine returning true.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/SnapshotContext.java
viliam-durina: fixed

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/AsyncSnapshotWriterImpl.java
cangencer: It'll be required if we want to prevent snapshot from completing before map is set.

viliam-durina: True, didn't realize it. Fixed.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -362,25 +353,26 @@ void tryStartJob(Function<Long, Long> executionIdSupplier) {
...
-            dag.edge(between(readSnapshotVertex, explodeVertex).isolated())
-               .edge(new SnapshotRestoreEdge(explodeVertex, vertex, destOrdinal));
+        String mapName = jobRecord.successfulSnapshotDataMapName();
+        logger.info("State of " + jobIdString() + " will be restored from snapshot " + snapshotId + ", map=" + mapName);
+
+        List<Vertex> originalVertices = new ArrayList<>();
+        dag.iterator().forEachRemaining(originalVertices::add);
+
+        Map<String, Integer> vertexToOrdinal = new HashMap<>();
+        Vertex readSnapshotVertex = dag.newVertex(SNAPSHOT_VERTEX_PREFIX + "read",
```
cangencer: what will be the parallelism here? since we only have one vertex, the throughput risks being much lower?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -808,8 +813,20 @@ void setFinalResult(Throwable failure) {
...
-            jobRecord = jobRecord.withQuorumSize(newQuorumSize);
+        if (jobRecord.getQuorumSize() < newQuorumSize) {
+            jobRecord.setQuorumSize(newQuorumSize);
+            writeJobRecord(true);
+            logger.info("Current quorum size: " + jobRecord.getQuorumSize() + " of job "
+                    + idToString(jobRecord.getJobId()) + " is updated to: " + newQuorumSize);
+        }
+    }
+
+    private void writeJobRecord(boolean async) {
```
cangencer: it's better to split this into two methods and call the method updateJobRecord/updateJobRecordAsync I think

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
viliam-durina: It will be the default, which is 2. 

The map reading is mostly io-bound. It reads the snapshot data chunks, which are `byte[]`, so they are easy to deserialize (`System.arraycopy`). The actual objects are deserialized in expand-snapshot-p, which is cooperative and uses max parallelism. Map reading uses `MapProxyImpl.iterator`, which is a sync operation, so there will be some waiting. Two threads will keep cpu and network busy enough.

However, this is just an "informed opinion", not based on benchmarks.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -383,51 +399,102 @@ public JobResult getJobResult(long jobId) {
...
+        dynamicData.updateTimestamp();
+        jobRecords.submitToKey(jobId, new UpdateJobRecordDynamicDataEntryProcessor(jobId, dynamicData),
+                writeJobRecordCallback);
+    }
 
-        private long jobId;
+    /**
+     * See {@link #writeJobRecordAsync}.
+     */
+    void writeJobRecordSync(long jobId, DynamicData dynamicData) {
```
metanet: Don't we have an isolation issue while using `writeJobRecordSync()` and `writeJobRecordAsync()` methods?

For instance, `JobRecord::ongoingSnapshotDone` method updates 7 fields of `JobRecord.DynamicData` and then calls `writeJobRecordSync()` to persist these changes.
In parallel, while `JobRecord::ongoingSnapshotDone` is updated some of the fields but not all of them, `writeJobRecordAsync()` method may kick in, say because of quorum value update, and persist a dirty `DynamicData` object.

metanet: I don't get why we have the async version in the first place. It is used in `setJobStatusToString`, `finalizeJob`, and `updateQuorumSize` for rare cases and it think these places can use the sync version. It is also called in `beginSnapshot`, which seems to me as the most important place. Since it is called asyncly, some snapshot data can be persisted using the new `ongoingSnapshotId` value. However, the master node can fail before it persists the `DynamicData` object. If this is a safe state for us, then we don't need the async persistence call here. Then, we can get rid of the async job record persistence method completely.

viliam-durina: The synchronization is described in the second bullet in the [initial comment](#issue-221797573). The `MasterContext` has final `jobRecord` field. The `JobRecord` has volatile fields that are used to store the state and updates to them are properly synchronized. `MasterContext` instance is created on master as soon as the JobRecord is discovered through scanning or through a submit request from the client, and the same instance is kept around until the job completes (this includes while the job is suspended).

Thus, the `MasterContext.jobRecord` field is the _master replica_ of the record. We only read it initially and write it out only for the case should the master fail. That's why we can do it in an async way, and we do so by default. Since two async `put`s can be reordered, we use the `timestamp` to ensure that older async write doesn't overwrite a newer one already written.

We write the JobRecord in a sync way only to store the information that a snapshot is completed. The reason is that after this we're going to delete the previous successful snapshot and we don't want to start deleting before `JobRecord.dataMapIndex` is updated to refer to the map we're not going to delete.

If the master fails after it writes some data to ongoing snapshot, but before it increments `ongoingSnapshotId`, the new master always clears the inactive snapshot map when starting the job (MasterContext:319). There could be some zombie write from the old master using the same `snapshotId` as the new master is going to use for the next snapshot, but this should be detected by the snapshot validation.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -808,11 +813,22 @@ void setFinalResult(Throwable failure) {
     }
 
     void updateQuorumSize(int newQuorumSize) {
-        synchronized (lock) {
-            jobRecord = jobRecord.withQuorumSize(newQuorumSize);
+        if (jobRecord.getQuorumSize() < newQuorumSize) {
```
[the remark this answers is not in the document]
viliam-durina: Fixed this. Just small difference: the smaller value will cause `setQuorumSize` to throw, not to overwrite. I also renamed the method to `setLargerQuorumSize` to reflect that.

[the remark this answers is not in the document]
metanet: 👍

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
metanet: as I told above, async calls are already used in the rarely-called places, like switching to / returning from pending state. The only "critical" place async call made is `beginSnapshot` method. My understanding is, if a snapshot begins without ensuring the change in `JobRecord.DynamicData` is persisted and this is ok, we don't need this async call. Then, we can get rid of the async call in `beginSnapshot()`, and convert the others to sync calls. wdyt?

cangencer: I would much prefer sync calls and only way to update the record. `beginSnapshot()` shouldn't be critical either, it will only slightly delay starting of the snapshot.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobExecutionRecord.java
```diff
@@ -0,0 +1,307 @@
...
+        return JetInitDataSerializerHook.FACTORY_ID;
+    }
+
+    @Override
+    public int getId() {
+        return JetInitDataSerializerHook.JOB_EXECUTION_RECORD;
+    }
+
+    @Override
+    public void writeData(ObjectDataOutput out) throws IOException {
```
cangencer: would it be a good idea to have a unit test for this, since a lot of fields are read/written?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobExecutionRecord.java
```diff
@@ -0,0 +1,307 @@
...
+        ongoingSnapshotStartTime = in.readLong();
+        lastFailureText = in.readBoolean() ? in.readUTF() : null;
+        quorumSize.set(in.readInt());
+        suspended = in.readBoolean();
+        timestamp.set(in.readLong());
+    }
+
+    @Override
+    public String toString() {
+        return "DynamicData{" +
```
cangencer: should say `JobExecutionRecord`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobExecutionRecord.java
```diff
@@ -0,0 +1,307 @@
...
+                ", suspended=" + suspended +
+                ", snapshotId=" + snapshotId +
+                ", dataMapIndex=" + dataMapIndex +
+                ", startTime=" + startTime +
+                ", endTime=" + endTime +
+                ", numBytes=" + numBytes +
+                ", numKeys=" + numKeys +
+                ", numChunks=" + numChunks +
+                ", ongoingSnapshotId=" + ongoingSnapshotId +
+                ", ongoingSnapshotStartTime=" + ongoingSnapshotStartTime +
```
cangencer: times should be converted to readable strings

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -383,51 +396,95 @@ public JobResult getJobResult(long jobId) {
                   .sorted(comparing(JobResult::getCreationTime).reversed()).collect(toList());
     }
 
-    public static class FilterExecutionIdByJobIdPredicate implements Predicate<Long, Long>, IdentifiedDataSerializable {
+    /**
+     * Write the JobRecord to the IMap.
```
cangencer: write the `JobExecutionRecord`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -383,51 +396,95 @@ public JobResult getJobResult(long jobId) {
...
+     * JobExecutionRecord} is older than the timestamp of the stored record.
+     * See {@link UpdateJobExecutionRecordEntryProcessor#process}. It will also
+     * be ignored if the key doesn't exist in the IMap.
+     */
+    void writeJobExecutionRecord(long jobId, JobExecutionRecord record, boolean canCreate) {
+        record.updateTimestamp();
+        String result = (String) jobExecutionRecords.executeOnKey(jobId,
+                new UpdateJobExecutionRecordEntryProcessor(jobId, record, canCreate));
+        if (result != null) {
+            logger.fine(result);
```
cangencer: result should be named something like `message`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -810,8 +825,26 @@ void setFinalResult(Throwable failure) {
...
+            logger.info("Current quorum size: " + jobExecutionRecord.getQuorumSize() + " of job "
+                    + idToString(jobRecord.getJobId()) + " is updated to: " + newQuorumSize);
+        }
+    }
+
+    private void writeJobExecutionRecord(boolean canCreate) {
+        try {
+            coordinationService.jobRepository().writeJobExecutionRecord(jobRecord.getJobId(), jobExecutionRecord,
+                    canCreate);
+            logger.info("aaa, written JER");
```
cangencer: seems like debug msg

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobExecutionRecord.java
viliam-durina: added

PR #1107, opened 2018-10-11
Use submitToKeys to merge to IMap

### LINKED ISSUE #970: mapWithMerging / mapWithUpdating sinks are inefficient with bulk updates
`mapWithMerging` and `mapWithUpdating` sinks can be inefficient when sending bulk updates because all of the data is sent to all the nodes regardless if they're needed or not.

A copy of the EntryProcessor is also created for each partition when taking backups, which adds another overhead.

viliam-durina:
IMap.executeOnKeys has an issue that a copy of the EntryProcessor (EP)
is sent to each member and, additionally, one copy is sent again for
backup. The EP used for `mapWithMerging` contained a HashMap with a
value for each key. The problem was, that the map got copied many times,
wasting a lot of memory.

This commit splits the HashMap into a submap, one for each partition,
and uses one `submitToKeys` call for each partition.

Fixes #970

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
@@ -143,7 +143,9 @@ public final class HazelcastWriters {
 
         private final Semaphore concurrentAsyncOpsSemaphore = new Semaphore(MAX_PARALLEL_ASYNC_OPS);
         private final AtomicReference<Throwable> firstError = new AtomicReference<>();
-        private final IMap map;
+        private final IMap<K, V> map;
+
+        // one map per partition to store the temporary values
         private final Map<Data, Object>[] tmpMaps;
 
         UpdateMapContext(
@@ -195,7 +197,7 @@ public final class HazelcastWriters {
                 partitionId = clientPartitionService.getPartitionId(keyData);
             }
             Data itemData = serializationService.toData(item);
-            tmpMaps[partitionId].merge(keyData, itemData, (o, n) -> MultiItem.merge(o, (Data) n));
+            tmpMaps[partitionId].merge(keyData, itemData, (o, n) -> ApplyFnEntryProcessor.append(o, (Data) n));
         }
 
         void flush() {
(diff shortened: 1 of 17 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -111,62 +116,129 @@ private HazelcastWriters() {
...
+                tmpMaps[i] = new HashMap<>();
+            }
         }
 
         void add(T item) {
-            buffer.add(item);
+            K key = toKeyFn.apply(item);
+            // Calculating partitionId requires serializing - we serialize the key twice unnecessarily, but I don't
+            // know a way to avoid it.
+            int partitionId = partitionService.getPartition(key).getPartitionId();
```
gurbuzali: this can be a problem for the maps which has a non-default partition strategy
```
        Config config = new Config();
        MapConfig mapConfig = config.getMapConfig("map");
        PartitioningStrategyConfig partitioningStrategyConfig = new PartitioningStrategyConfig();
        partitioningStrategyConfig.setPartitioningStrategyClass(StringPartitioningStrategy.class.getName());
        mapConfig.setPartitioningStrategyConfig(partitioningStrategyConfig);

        HazelcastInstanceProxy hz = ((HazelcastInstanceProxy)Hazelcast.newHazelcastInstance(config));
        PartitionService partitionService = hz.getPartitionService();

        String key = "key@actualPartition";

        Partition partition = partitionService.getPartition(key);
        System.out.println(partition);

        Data serializedKey = hz.getSerializationService().toData(key, StringPartitioningStrategy.INSTANCE);
        Partition partitionFromSerialized = partitionService.getPartition(serializedKey);
        System.out.println(partitionFromSerialized);
```

gurbuzali: created an issue for it 
https://github.com/hazelcast/hazelcast/issues/13950

viliam-durina: fixed: when writing locally, we'll use correct partitioning strategy. When writing remotely, we can't, there's no api on client

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -111,62 +124,179 @@ private HazelcastWriters() {
...
+                // is calculated and when the value for backup operation is re-serialized
+                keyData = serializationService.toData(key, ((MapProxyImpl) map).getPartitionStrategy());
+                partitionId = internalPartitionService.getPartitionId(keyData);
+            } else {
+                // We ignore partition strategy for remote connection, the client doesn't know it.
+                // TODO we might be able to fix this after https://github.com/hazelcast/hazelcast/issues/13950 is fixed
+                // The functionality should work, but will be ineffective: the submitOnKey calls will have wrongly
+                // partitioned data.
+                keyData = serializationService.toData(key);
+                partitionId = partitionService.getPartition(key).getPartitionId();
```
gurbuzali: we should use `keyData` here ?

viliam-durina:
verify

Issue #1105, opened 2018-10-10
TopologyChangeTest.when_nonCoordinatorLeavesDuringExecution_then_jobRestarts failure


viliam-durina:
https://hazelcast-l337.ci.cloudbees.com/job/jet-pr-builder/com.hazelcast.jet$hazelcast-jet-core/3259/testReport/junit/com.hazelcast.jet.core/TopologyChangeTest/when_nonCoordinatorLeavesDuringExecution_then_jobRestarts_1_/

```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.hazelcast.jet.core.TopologyChangeTest.lambda$when_nonCoordinatorLeavesDuringExecution_then_jobRestarts$2(TopologyChangeTest.java:196)
	at com.hazelcast.jet.core.JetTestSupport$1.run(JetTestSupport.java:168)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1294)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1320)
	at com.hazelcast.jet.core.JetTestSupport.assertTrueEventually(JetTestSupport.java:129)
	at com.hazelcast.jet.core.TopologyChangeTest.when_nonCoordinatorLeavesDuringExecution_then_jobRestarts(TopologyChangeTest.java:191)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
[...]

PR #1114, opened 2018-10-17
Fix test failures

viliam-durina:


viliam-durina:
This PR also fixes #1105

PR #1117, opened 2018-10-18
Allow init() methods to throw Exception

### LINKED ISSUE #1116: Processor.init(), PS.init() and PMS.init() should throw Exception
This should be for convenience, as these exceptions are already handled downstream and it's common when opening connections and similar that the API used will also have throw in the API.

viliam-durina:
Fixes #1116


viliam-durina:
verify

PR #1121, opened 2018-10-25
Don't materialize `distinctKeys` in SessionWindowP

viliam-durina:
Materialization of `distinctKeys` isn't needed (it even materialized a
Set first inside the `distinct()` stage and then copied it to a List).
Most of the time, each key has one Window, and if not, we'll call
`closeWindows` again and it will find out that there's no window to
close.


viliam-durina:
verify

PR #1131, opened 2018-11-05
Document batch sources about reading mutating data

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/Sources.java
```diff
@@ -892,6 +892,9 @@ private Sources() {
      * IList}. All elements are emitted on a single member &mdash; the one
      * where the entire list is stored by the IMDG.
      * <p>
+     * If the {@code IList} is modified while being read, the source may miss
```
gurbuzali: we read all the items in the list to local with a single operation, so there shouldn't be any duplicate items.

gurbuzali: if list size is greater than `FETCH_SIZE` then we get the items with multiple operations. so 👍

PR #1147, opened 2018-11-12
Add ClientConfigXmlGenerator

gurbuzali:


cangencer:
Can we send PR to Hazelcast as well? I think it can be added in a patch to 3.11?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/ClientConfigXmlGenerator.java
```diff
@@ -0,0 +1,636 @@
...
+    private static final int INDENT = 5;
+
+    private static final ILogger LOGGER = Logger.getLogger(ConfigXmlGenerator.class);
+
+    private final boolean formatted;
+
+    /**
+     * Creates a new generator instance with {@link #formatted} {@code false}
+     */
+    public ClientConfigXmlGenerator() {
```
cangencer: looks unused

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/ClientConfigXmlGenerator.java
```diff
@@ -0,0 +1,636 @@
...
+    /**
+     * Transforms the given {@link ClientConfig} to xml string
+     */
+    public String generate(ClientConfig clientConfig) {
+        Preconditions.isNotNull(clientConfig, "ClientConfig");
+
+        StringBuilder xml = new StringBuilder();
+        XmlGenerator gen = new XmlGenerator(xml);
+
+        xml.append("<hazelcast-client ")
```
cangencer: why not use `.open` here?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/ClientConfigXmlGenerator.java
```diff
@@ -0,0 +1,636 @@
...
+        }
+
+        private static void appendNode(StringBuilder xml, String name, Object contents, Object... attributes) {
+            if (contents != null || attributes.length > 0) {
+                xml.append('<').append(name);
+                for (int i = 0; i < attributes.length; ) {
+                    Object key = attributes[i++];
+                    Object val = attributes[i++];
+                    if (val != null) {
+                        xml.append(" ").append(key).append("=\"").append(val).append("\"");
```
cangencer: I don't see any escaping of values here

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/ClientConfigXmlGenerator.java
```diff
@@ -0,0 +1,636 @@
...
+                    }
+                }
+                if (contents != null) {
+                    xml.append('>').append(contents).append("</").append(name).append('>');
```
cangencer: also here it looks like values are not escaped properly

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/ClientConfigXmlGenerator.java
```diff
@@ -0,0 +1,708 @@
...
+        }
+
+        private static void appendCloseNode(StringBuilder xml, String name) {
+            xml.append("</").append(name).append('>');
+        }
+
+        private static void appendAttributes(StringBuilder xml, Object... attributes) {
+            for (int i = 0; i < attributes.length; ) {
+                xml.append(" ");
+                escapeXml(attributes[i++], xml);
```
viliam-durina: We don't need to escape attribute names. Rather, we should check them for sanity, I found [these rules](https://stackoverflow.com/a/44057262/952135). I think we can just ignore that, we don't use user-input for attribute names and hard-coded attribute names should be ok.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/ClientConfigXmlGenerator.java
```diff
@@ -0,0 +1,708 @@
...
+                xml.append("=\"");
+                escapeXml(attributes[i++], xml);
+                xml.append("\"");
+            }
+        }
+
+        private static void appendNode(StringBuilder xml, String name, Object contents, Object... attributes) {
+            if (contents != null || attributes.length > 0) {
+                xml.append('<').append(name);
+                for (int i = 0; i < attributes.length; ) {
```
viliam-durina: We can delegate to `appendAttributes` and other append methods here.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/config/ClientConfigXmlGeneratorTest.java
```diff
@@ -0,0 +1,508 @@
...
+@RunWith(HazelcastParallelClassRunner.class)
+public class ClientConfigXmlGeneratorTest extends JetTestSupport {
+
+    private static final boolean DEBUG = false;
+    private static final Random RANDOM = new Random();
+
+    private ClientConfig clientConfig = new ClientConfig();
+
+    @Test
+    public void escape() {
```
cangencer: missing test for escaping attributes

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/ClientConfigXmlGenerator.java
```diff
@@ -0,0 +1,696 @@
...
+                    case '"':
+                        appendTo.append("&quot;");
+                        break;
+                    case '\'':
+                        appendTo.append("&#39;");
+                        break;
+                    case '&':
+                        appendTo.append("&amp;");
+                        break;
+                    case '<':
```
cangencer: technically you don't need to escape `<>` in attributes I think

viliam-durina: You need to escape `<`, you never need to escape `>`

PR #1152, opened 2018-11-19
Don't merge touching sessions in SessWP

viliam-durina:
If two sessions "touch", we currently merge them. Two sessions are 
touching, if when `window1.end == window2.start`. This PR changes the 
behavior to not merge them. Reason: otherwise we'd depend on the order 
of input in edge case.

Example: timeout=2. If input is: `1, wm(3), 3`, the output will be two
sessions: the event 1 creates a session (1, 3) which can be emitted at
wm(3). But if input is: `1, 3, wm(3)`, output would be 1 session (1, 5),
if we merge "touching" sessions. In neither case item 3 is late. The
wm(3) can come at any time depending on exact order in which other
partitions are processed and when the wm is coalesced.

This PR also fixes the case when `saveToSnapshot` is called while in 
`complete` where we didn't prevent creating snapshot while there could 
have been a half-emitted item.


viliam-durina:
Closed, it's part of #1154

PR #1154, opened 2018-11-20
Job upgrade

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/JetInstance.java
```diff
@@ -134,6 +138,39 @@ default Job getJob(@Nonnull String name) {
...
+    /**
+     * Returns the IMap that contains the data of an exported snapshot with the
+     * given name.
+     * <p>
+     * The contents of the map are unspecified.
+     *
+     * @param name exported snapshot name
+     * @return IMap instance with the data
+     */
+    default IMapJet<Object, Object> getExportedState(@Nonnull String name) {
```
cangencer: it shouldn't just return a map. The underlying map should not be visible to the user. Instead some wrapper object which has some additional info would be more useful. i.e. when the snapshot was taken, which job it originated from, how big is it.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/JetInstance.java
```diff
@@ -134,6 +138,39 @@ default Job getJob(@Nonnull String name) {
...
+        return getMap(Jet.EXPORTED_STATES_PREFIX + name);
+    }
+
+    /**
+     * Returns the collection of names of existing exported states.
+     */
+    default Collection<String> getExportedStateNames() {
```
cangencer: see my comment above. Returning a list of strings without any context would not be useful.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
```diff
@@ -143,17 +144,92 @@ default void join() {
...
+     * <p>
+     * Job status will be {@link JobStatus#COMPLETED} after cancellation, even
+     * though the job didn't really complete. However, {@link Job#join()} will
+     * throw an exception.
+
+     * TODO [viliam] finish javadoc
+     *
+     * @param name name of the snapshot
+     */
+    void cancelAndExportState(String name);
```
cangencer: is there a way to suspend and export state?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/JetInstance.java
```diff
@@ -134,6 +138,39 @@ default Job getJob(@Nonnull String name) {
...
+                                     .map(o -> o.getName().substring(Jet.EXPORTED_STATES_PREFIX.length()))
+                                     .collect(toList());
+    }
+
+    /**
+     * Destroys the map with exported state with the given name.
+     *
+     * @param name exported snapshot name
+     */
+    default void deleteExportedState(@Nonnull String name) {
```
cangencer: doesn't seem like this is needed. We don't have `JetInstance.deleteMap()`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Util.java
```diff
@@ -113,4 +115,20 @@ public static String idToString(long id) {
...
         return new String(buf);
     }
+
+    /**
+     * TODO [viliam] javadoc
+     *
+     * @param instance
+     * @param snapshotMapName
+     */
+    public static void cleanUpSnapshotMap(JetInstance instance, String snapshotMapName) {
```
cangencer: why does this need to be a public API?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -422,4 +423,25 @@ public JobConfig setClassLoaderFactory(@Nullable JobClassLoaderFactory classLoad
...
+
+    /**
+     * TODO [viliam] finish javadoc
+     *
+     * You can restore state also for non-snapshotted jobs.
+     *
+     * @param initialSnapshotName
+     * @return
+     */
+    public JobConfig setInitialSnapshotName(String initialSnapshotName) {
```
cangencer: here we call it "initial snapshot" but in the Job API it's called `exportedState`. It doesn't quite match in my opinion.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
viliam-durina: Yes:
```
job.suspend();
while (job.getStatus() != SUSPENDED) {
    Thread.sleep(500);
}
job.exportState();
```
I don't think this use case is common to have a special support for it. When would you need it?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/JetInstance.java
viliam-durina: fixed, now returns `JobStateSnapshot`

cangencer:
Missing things: 

* `getJobStateSnapshot` in current form is racy because it tries to destroy the map. We need a method like `distributedObjectExists` in `JetInstance` on both member and client
* `SnapshotValidationRecord` should be `IdentifiedDataSerializable`
* `ExportSnapshotTest` should also test everything from client

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
mtopolnik: A logical name would be `importSnapshot`

viliam-durina: The export method is now called `exportSnapshot` and the holder object is `JobStateSnapshot`. I don't like `importSnapshot`, because we don't import anything. The initial snapshot can even be loaded multiple times, if the job restarts before it creates first regular snapshot. It is the snapshot we load initially if we don't have a regular snapshot.

PR #1161, opened 2018-11-26
Upgrade IMDG dependency to 3.12-SNAPSHOT

mtopolnik:


cangencer:
verify

cangencer:
enterprise, code samples and so on also need updating

PR #1166, opened 2018-12-05
Make watermarks source explicit in Pipeline API

viliam-durina:


viliam-durina:
verify
Can't reproduce the last test failure, logs are incomplete...

viliam-durina:
verify

cangencer:
* JMS source should also include default timestamps
* It would be better if pipeline would fail during planning stage when `withDefaultTimestamps()` is used but source doesn't provide default timestamps.
* It should be documented on sources that support it what the `defaultTimestamp` corresponds to.

PR #1170, opened 2018-12-11
Create priority queue with capacity as 1 if queue size is 0

gurbuzali:
fixes https://github.com/hazelcast/hazelcast-jet/issues/1169

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/serialization/PriorityQueueHook.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/serialization/PriorityQueueHook.java
@@ -23,7 +23,6 @@ import com.hazelcast.nio.serialization.SerializerHook;
 import com.hazelcast.nio.serialization.StreamSerializer;
 
 import java.io.IOException;
-import java.util.Comparator;
 import java.util.PriorityQueue;
 
 public final class PriorityQueueHook implements SerializerHook<PriorityQueue> {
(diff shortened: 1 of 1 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/serialization/PriorityQueueHook.java
```diff
@@ -59,7 +60,8 @@ public void write(ObjectDataOutput out, PriorityQueue queue) throws IOException
             @Override
             public PriorityQueue read(ObjectDataInput in) throws IOException {
                 int size = in.readInt();
-                PriorityQueue res = new PriorityQueue(size, in.readObject());
+                Comparator comparator = in.readObject();
+                PriorityQueue res = size < 1 ? new PriorityQueue(comparator) : new PriorityQueue(size, comparator);
```
viliam-durina: It's ok, but shorter would be:
```
PriorityQueue res = new PriorityQueue(Math.min(1, size), in.readObject());
```

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/serialization/PriorityQueueHook.java
```diff
@@ -59,7 +59,7 @@ public void write(ObjectDataOutput out, PriorityQueue queue) throws IOException
             @Override
             public PriorityQueue read(ObjectDataInput in) throws IOException {
                 int size = in.readInt();
-                PriorityQueue res = new PriorityQueue(size, in.readObject());
+                PriorityQueue res = new PriorityQueue(Math.min(1, size), in.readObject());
```
gokhanoner: Isn't that suppose to be `Math.max(1, size)` ?

cangencer: It was fixed here: https://github.com/hazelcast/hazelcast-jet/pull/1172

PR #1173, opened 2018-12-12
Add more metrics for WM debugging

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/InboundEdgeStream.java
```diff
@@ -40,4 +40,7 @@
     int capacities();
 
     int sizes();
+
+    long topObservedWm();
```
cangencer: let's add some javadoc to these for future..

PR #1177, opened 2018-12-13
add workaround for including javadoc from core module into distribution

cangencer:


cangencer:
yes it should be applied to master as well.

PR #1178, opened 2018-12-14
Ensure minimum WM rate to improve WM debugging

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/Planner.java
```diff
@@ -50,6 +50,13 @@
 @SuppressWarnings("unchecked")
 public class Planner {
 
+    /**
+     * Minimum frequency of watermarks. This is not technically necessary, but
+     * improves debugging by avoiding too large gaps between WMs and the user
+     * can better observe if input to WM coalescing is lagging.
+     */
+    private static final int MINIMUM_WATERMARK_RATE = 1_000;
```
cangencer: wouldn't it be better to call it `MAXIMUM_WATERMARK_GAP` or similar?

viliam-durina: agree

viliam-durina:
verify

PR #1183, opened 2018-12-18
Don't scale up to members with no partitions

viliam-durina:


hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/StreamSourceStageTestBase.java
```diff
@@ -103,6 +103,7 @@ protected void test(
             return;
         }
         stageWithTimestamps
+                .peek()
```
gurbuzali: leftover ?

viliam-durina: There are just 2 items in the source logged, this helps debugging.

PR #1184, opened 2018-12-18
Remove max-retain and policies

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/Planner.java
```diff
@@ -190,6 +185,18 @@ static String uniqueName(
         }
     }
 
+    /**
+     * Returns new instance with emit policy replaced with the
```
mtopolnik: Returns _a_ new instance

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/WatermarkCoalescer.java
```diff
@@ -153,34 +143,71 @@ public long coalescedWm() {
         public long topObservedWm() {
             return Long.MIN_VALUE;
         }
+    }
+
+    /**
+     * Special-case implementation for zero inputs (source processors).
+     */
+    private static final class SingleInputImpl extends WatermarkCoalescer {
```
mtopolnik: Is it zero inputs or single input? Either Javadoc or class name is wrong.

PR #1191, opened 2019-01-02
Fix possible race with exportSnapshot

viliam-durina:
We changed the `jobStatus` to SUSPENDED and then we proceeded to copy
the validation key to the cache. In the meantime, the snapshot could
have been deleted.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
@@ -257,9 +257,12 @@ public class MasterContext {
      *     <li>a string with a message why this call did nothing or null, if
      *          this call actually initiated the termination
      * </ol>
+     *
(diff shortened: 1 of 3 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -259,7 +259,7 @@ public JobExecutionRecord jobExecutionRecord() {
      * </ol>
      */
     @Nonnull
-    Tuple2<CompletableFuture<Void>, String> requestTermination(TerminationMode mode) {
+    Tuple2<CompletableFuture<Void>, String> requestTermination(TerminationMode mode, boolean allowExportingSnapshot) {
```
cangencer: what does this parameter mean?

viliam-durina: i clarified it :)

PR #1192, opened 2019-01-02
Introduced ClusterMetadata 

eminn:
and methods and messages to retrieve it from client

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClusterMetadata.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClusterMetadata.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
+ * Copyright (c) 2008-2019, Hazelcast, Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClusterMetadata.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClusterMetadata.java
@@ -18,19 +18,14 @@ package com.hazelcast.jet.impl;
 
 import com.hazelcast.cluster.ClusterState;
 import com.hazelcast.core.Cluster;
-import com.hazelcast.core.Member;
 import com.hazelcast.instance.BuildInfoProvider;
 import com.hazelcast.jet.impl.execution.init.JetInitDataSerializerHook;
-import com.hazelcast.nio.Address;
 import com.hazelcast.nio.ObjectDataInput;
 import com.hazelcast.nio.ObjectDataOutput;
 import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
 
 import javax.annotation.Nonnull;
 import java.io.IOException;
-import java.util.HashSet;
-import java.util.Set;
-import java.util.stream.Collectors;
 
 public class ClusterMetadata implements IdentifiedDataSerializable {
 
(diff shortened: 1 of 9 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/operation/GetClusterMetadataOperation.java
```diff
@@ -0,0 +1,63 @@
...
+import com.hazelcast.jet.impl.ClusterMetadata;
+import com.hazelcast.jet.impl.JetService;
+import com.hazelcast.jet.impl.execution.init.JetInitDataSerializerHook;
+import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
+import com.hazelcast.spi.Operation;
+import com.hazelcast.spi.ReadonlyOperation;
+
+public class GetClusterMetadataOperation extends Operation implements
+        IdentifiedDataSerializable,
+        ReadonlyOperation {
```
cangencer: I was looking at `JobSummaryListOperation` and it's marked as `AllowedDuringPassiveState` while this is marked as read-only. What is the difference?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/client/JetGetClusterMetadataMessageTask.java
```diff
@@ -0,0 +1,58 @@
...
+import com.hazelcast.instance.Node;
+import com.hazelcast.jet.impl.operation.GetClusterMetadataOperation;
+import com.hazelcast.nio.Connection;
+import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.spi.Operation;
+import com.hazelcast.spi.serialization.SerializationService;
+
+public class JetGetClusterMetadataMessageTask extends
+        AbstractJetMessageTask<JetGetClusterMetadataCodec.RequestParameters> {
+    protected JetGetClusterMetadataMessageTask(ClientMessage clientMessage,
```
cangencer: indentation should be fixed

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetClientInstanceImpl.java
```diff
@@ -138,9 +139,23 @@ public Job getJob(long jobId) {
...
+                ClientMessage response = invocation.invoke().get();
+                return serializationService.toObject(JetGetJobSummaryListCodec.decodeResponse(response).response);
+        });
+    }
+
+    /**
+     * Returns summary of cluster details.
+     */
+    @Nonnull
+    public ClusterMetadata getClusterMetadata() {
```
cangencer: seems some duplication of code with above method.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/operation/GetClusterMetadataOperation.java
eminn: From our point of view there is no difference, so I will align them on `ReadonlyOperation`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClusterMetadata.java
```diff
@@ -0,0 +1,141 @@
...
+        out.writeBoolean(sslEnabled);
+    }
+
+    @Override
+    public void readData(ObjectDataInput in) throws IOException {
+        name = in.readUTF();
+        version = in.readUTF();
+        state = in.readObject();
+        int memberSize = in.readInt();
+        if (memberSize > 0) {
```
cangencer: if size = 0 this will cause members to be null.. I'm not sure if this intended? Also is list of members needed here, as it's already available on the client?

eminn: yes, we can remove list of members from here and rely on the client

eminn:
verify

Issue #1176, opened 2018-12-13
Registering/deregistering migration listener can hang


viliam-durina:
`PartitionService.addMigrationListener()` and `removeMigrationListener()` do a call that gets executed on an `async` thread. We call these methods also from an async thread (from `Processor.init()`) too. If there are enough parallel calls, we can block out all async threads by operations that cannot complete, because they wait for an async thread -> a deadlock.

Stack trace of all async threads during the deadlock is:
```
"hz._hzInstance_1_jet.async.thread-8" #131 prio=5 os_prio=0 tid=0x0000000019bb0000 nid=0x4b6c in Object.wait() [0x00000000269ee000]
   java.lang.Thread.State: TIMED_WAITING (on object monitor)
    at java.lang.Object.wait(Native Method)
    at com.hazelcast.spi.impl.AbstractCompletableFuture.get(AbstractCompletableFuture.java:229)
    - locked <0x00000000f7996c48> (a com.hazelcast.internal.util.futures.ChainingFuture)
    at com.hazelcast.spi.impl.AbstractCompletableFuture.get(AbstractCompletableFuture.java:191)
    at com.hazelcast.spi.impl.eventservice.impl.EventServiceImpl.invokeOnAllMembers(EventServiceImpl.java:320)
    at com.hazelcast.spi.impl.eventservice.impl.EventServiceImpl.deregisterListener(EventServiceImpl.java:312)
    at com.hazelcast.internal.partition.impl.PartitionEventManager.removeMigrationListener(PartitionEventManager.java:89)
    at com.hazelcast.internal.partition.impl.InternalPartitionServiceImpl.removeMigrationListener(InternalPartitionServiceImpl.java:1072)
    at com.hazelcast.internal.partition.PartitionServiceProxy.removeMigrationListener(PartitionServiceProxy.java:104)
    at com.hazelcast.jet.impl.connector.ReadWithPartitionIteratorP$MigrationWatcher.deregister(ReadWithPartitionIteratorP.java:435)
    ...
```

We can resolve this by:
- offloading the registration to another thread
- using one global registration in `JetService`

PR #1197, opened 2019-01-10
Partially revert "Add partition migration detection (#1076)"

viliam-durina:
This feature currently causes deadlock in soak tests, needs to be fixed
as per #1176. This is a cherry-pick from 0.7-maintenance.


viliam-durina:
verify

PR #1202, opened 2019-01-16
Jet CLI

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
```diff
@@ -201,7 +201,7 @@ default void join() {
     JobStateSnapshot cancelAndExportSnapshot(String name);
 
     /**
-     * Initiates an export of a state snapshot and saves it under the given
+     * Exports a state snapshot and saves it under the given
```
viliam-durina: `cancelAndExportSnapshot` also needs this fix.

PR #1218, opened 2019-01-28
Improvements to JetCommandLine

cangencer:


hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetCommandLine.java
```diff
@@ -67,9 +68,16 @@
 @Command(
         name = "jet",
         header = "Hazelcast Jet",
```
eminn: should we also include the product version in header ?

cangencer:
verify

Holmistr:
verify

Holmistr:
verify

Holmistr:
verify

PR #1223, opened 2019-01-31
Fix java9+ build

viliam-durina:
Credits go to @gokhanoner

### DIFF
--- a/hazelcast-jet-spring/pom.xml
+++ b/hazelcast-jet-spring/pom.xml
@@ -135,6 +135,7 @@
             <groupId>javax.annotation</groupId>
             <artifactId>javax.annotation-api</artifactId>
             <version>1.3.2</version>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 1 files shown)


hazelcast-jet-spring/pom.xml
```diff
@@ -131,5 +131,10 @@
             <version>${spring.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>javax.annotation</groupId>
+            <artifactId>javax.annotation-api</artifactId>
+            <version>1.3.2</version>
+        </dependency>
```
gokhanoner: I believe this is only needed for test scope, since `hazelcast-jet-spring` module itself doesn't require this dependency?

viliam-durina: If the annotations go unnoticed in test scope without this dep, why would't they at runtime?

gokhanoner: You’re not using any annotation in hazelcast-jet-spring module, how could it cause any problem at runtime?

PR #1225, opened 2019-01-31
Shutdown Jet client automatically when using JetBootstrap

cangencer:


hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetBootstrap.java
```diff
@@ -153,7 +157,8 @@ private static void error(String msg) {
     }
 
     /**
-     * Returns the bootstrapped {@code JetInstance}.
+     * Returns the bootstrapped {@code JetInstance}. The instance will automatically
+     * be shutdown once the {@code main()} method of the JAR returns.
```
viliam-durina: "will be automatically shut down"

PR #1241, opened 2019-02-06
Fix an initialization race

mtopolnik:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/NodeExtensionCommon.java
```diff
@@ -41,17 +40,16 @@
...
+    private volatile ILogger logger;
 
     NodeExtensionCommon(Node node) {
         this.node = node;
     }
 
     void afterStart() {
-        jetService = node.nodeEngine.getService(JetService.SERVICE_NAME);
-        jetService.getJobCoordinationService().startScanningForJobs();
         logger = node.getLogger(getClass().getName());
```
viliam-durina: We should check the `logger` is not null when using it. It can happen.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/NodeExtensionCommon.java
```diff
@@ -82,6 +82,14 @@ void printNodeInfo(ILogger log, String addToProductName) {
         log.info(COPYRIGHT_LINE);
     }
 
+    private JetService jetService() {
+        JetService local = jetService;
+        if (local == null) {
+            local = jetService = node.nodeEngine.getService(JetService.SERVICE_NAME);
```
viliam-durina: I'd throw specific exception here if `jetService` is still null after this.

PR #1242, opened 2019-02-06
Improve constantKey perf, fix global aggregation

### LINKED ISSUE #1240: Don't use same constant key for all global aggregations
Currently all global aggregations use the same key [(here)](https://github.com/hazelcast/hazelcast-jet/blob/5dd0ac78d0c2caceb3102640588e68bff537db64/hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/DistributedFunctions.java#L31). This causes that all aggregations run on the same member and snapshots are stored in the same partition.  We should use constant, but randomly generated key for each aggregation.

viliam-durina:
Each invocation of constantKey() will return a different function
instance, that constantly returns a random key.

Fixes #1240

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/DistributedFunctions.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/DistributedFunctions.java
@@ -16,6 +16,8 @@
 
 package com.hazelcast.jet.function;
 
+import com.hazelcast.jet.impl.util.ConstantFunction;
+
 import javax.annotation.Nonnull;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -24,11 +26,6 @@ import java.util.Map.Entry;
  * Factory methods for several common distributed functions.
  */
 public final class DistributedFunctions {
-
-    /**
-     * The string key returned by {@link #constantKey()}.
-     */
-    public static final String CONSTANT_KEY = "ALL";
 
     private DistributedFunctions() {
     }
@@ -63,15 +60,14 @@ public final class DistributedFunctions {
     }
 
     /**
-     * Returns a function that always evaluates to {@value #CONSTANT_KEY}. This
+     * Returns a function that always evaluates to the given {@code key}. This
      * is useful as a key extractor in group-by operations where no
-     * classification by key is desired. Note, however, that all such
-     * aggregations will run on the same member because they are routed using
-     * the same key. To likely avoid that, use different key for each such step
-     * using {@code t -> "foo"}.
+     * classification by key is desired; in other words, a global group is
+     * used. Choose a random {@code key} so that each global group is handled
+     * by a random member.
      */
     @Nonnull
-    public static <T> DistributedFunction<T, String> constantKey() {
-        return t -> CONSTANT_KEY;
+    public static <T, K> DistributedFunction<T, K> constantKey(K key) {
+        return new ConstantFunction<>(key);
     }
 }
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/DistributedFunctions.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/DistributedFunctions.java
@@ -16,8 +16,6 @@
 
 package com.hazelcast.jet.function;
 
-import com.hazelcast.jet.impl.util.ConstantFunction;
-
 import javax.annotation.Nonnull;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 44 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/PipelineImpl.java
```diff
@@ -158,6 +154,16 @@ public String toDotString() {
 
     void register(Transform stage, List<Transform> downstream) {
         List<Transform> prev = adjacencyMap.put(stage, downstream);
-        assert prev == null : "Double registering of a Stage with this Pipeline: " + stage;
+        assert prev == null : "Double registration of a Stage with this Pipeline: " + stage;
+    }
+
+    void makeNamesUnique() {
+        Set<String> usedNames = new HashSet<>();
+        for (Transform transform : adjacencyMap.keySet()) {
```
cangencer: is this done in topological way? so we don't end up with stages like map-3 -> map-2 -> map

viliam-durina: Yes, pipeline transforms are always added in topological order. You can't add sink before the source. It's also checked [here](https://github.com/hazelcast/hazelcast-jet/pull/1242/files#diff-f1c767808988b2fb6d4adb9a4971023bR107) before creating the DAG.

mtopolnik: Once we start modifying the transform DAG, it will disturb the assumed ordering.

viliam-durina: We don't do it now. That's why there's the check in `createDag`: if we ever unintentionally break the order, it will fail.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/DistributedFunctions.java
```diff
@@ -63,12 +60,14 @@ private DistributedFunctions() {
...
+     * Returns a function that always evaluates to the given {@code key}. This
+     * is useful as a key extractor in group-by operations where no
+     * classification by key is desired; in other words, a global group is
+     * used. Choose a random {@code key} so that each global group is handled
+     * by a random member.
      */
     @Nonnull
-    public static <T> DistributedFunction<T, String> constantKey() {
-        return t -> CONSTANT_KEY;
+    public static <T, K> DistributedFunction<T, K> constantKey(K key) {
```
cangencer: does this need to be part of public API? I think it would be better to move it to `ConstantFunction`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/StreamStageImpl.java
```diff
@@ -131,7 +131,7 @@ public StreamStageImpl(
 
     @Nonnull @Override
     public <R> StreamStage<R> rollingAggregate(@Nonnull AggregateOperation1<? super T, ?, ? extends R> aggrOp) {
-        return groupingKey(constantKey()).rollingAggregate(aggrOp, (k, v) -> v);
+        return groupingKey(constantKey(name().hashCode())).rollingAggregate(aggrOp, (k, v) -> v);
```
cangencer: is it a good idea to use `name`? it's always generated to the same value if not explicitly set (i.e. rolling-aggregate)

viliam-durina: Hmm, true, in this place it's not unique, it's made unique later when calling `createDag`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/DistributedFunctions.java
viliam-durina: Yes, it doesn't. Pipeline API uses it internally and it's not needed for core api - the user can use `allToOne` and appropriate key extractor in the downstream vertex.

cangencer:
verify

viliam-durina:
Breaking changes in the PR: 
- `DistributedFunction.constantKey()` is removed
- `Edge.allToOne()` has one parameter added: the constant key. It used to 
generate random key, which broke job upgradability and also broke joins which
the user intuitively guessed will go to the same processor: all-to-one didn't mean
all-to-the-same-one.

PR #1248, opened 2019-02-08
Added license key property to be injected by enterprise docker image

eminn:


hazelcast-jet-distribution/src/main/resources/bin/common.sh
```diff
@@ -38,6 +38,14 @@ if [ "x$JAVA_OPTS_DEFAULT" != "x" ]; then
     JAVA_OPTS="${JAVA_OPTS_DEFAULT} ${JAVA_OPTS}"
 fi
 
+if [ "x$JET_LICENSE_KEY" != "x" ]; then
+	JAVA_OPTS="$JAVA_OPTS -Dhazelcast.enterprise.license.key=${JET_LICENSE_KEY}"
+fi
+
+if [ "x$JET_LICENCE_KEY" != "x" ]; then
```
cangencer: the if is duplicated, also the indentantion on the various options should be aligned. I would also group the statements effecting JAVA_OPTS and CLASSPATH so it looks bit more organized

eminn: It is not duplicated but covering the spelling options of `licen(c|s)e`

cangencer: why do we need two different properties though?

eminn: The outcome of this https://github.com/hazelcast/hazelcast-docker/issues/68 was to accept both spellings for user convenience.

cangencer: ok, at least the indentation of the file should be fixed / cleaned up. I'm also not sure why we use `"x$JET_LICENSE_KEY" != "x"` rather than `if [ -n "$VAR" ]`

eminn: I agree with the indentation, I'll fix it. For the other concern see https://stackoverflow.com/a/174288 for more detailed answer.

cangencer:
verify

PR #1251, opened 2019-02-11
Fix `customTransform`

viliam-durina:
* add overrides returning `StreamStage` and `BatchStage`
* add 3 overloads taking each form of processor supplier
* add missing tests


cangencer:
Marked it as breaking-change, needs to document what's changed

PR #1258, opened 2019-02-12
Improve `limitingLag` javadoc

### LINKED ISSUE #712: Clarify JavaDoc in the WatermarkPolicies
Please see [this](https://github.com/jerrinot/hazelcast-jet/blob/cf6a3297d09ae56931414450c95e73de4634eaed/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/WatermarkPolicies.java#L38-L41):
```
    /**
     * Maintains a watermark that lags behind the top observed timestamp by the
     * given amount. In the case of a stream lull the watermark does not
     * advance towards the top observed timestamp and remains behind it
     * indefinitely.
     *
     * @param lag the desired difference between the top observed timestamp
     *            and the watermark
     */
```

I have no idea what does it mean. I would like to see some guidance here. e.g. when this is a good choice, when it's not, etc.. 

Also I am not sure if "lull" is a typo or a real word:) We should use simple English.

viliam-durina:
Fixes #712

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/WatermarkPolicy.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/WatermarkPolicy.java
@@ -63,9 +63,13 @@ public interface WatermarkPolicy {
 
     /**
      * Maintains a watermark that lags behind the top observed timestamp by the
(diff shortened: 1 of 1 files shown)


mtopolnik:
verify

PR #1277, opened 2019-02-19
Check cluster-state to be same for all members before submitting the jobs

gurbuzali:
PR adds a cluster-wide check for the cluster state before submitting the jobs.

Before this change, when cluster is at `FROZEN` state we were not starting the jobs. With the change, when cluster is at `PASSIVE` or `IN_TRANSITION` we are not starting the jobs.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
@@ -473,7 +473,7 @@ public class JobCoordinationService {
                 || clusterState == PASSIVE || clusterState == IN_TRANSITION) {
             return false;
         }
-        if (!checkAllMembersHaveTheSameState(clusterState)) {
+        if (!allMembersHaveSameState(clusterState)) {
             LoggingUtil.logFine(logger, "Not starting jobs because not all members have the same state: %s",
                     clusterState);
             return false;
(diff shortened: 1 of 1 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -479,6 +491,21 @@ boolean shouldStartJobs() {
...
     }
 
+    private boolean checkAllMembersHaveTheSameState(ClusterState clusterState) {
+        Set<Member> members = nodeEngine.getClusterService().getMembers();
+        InternalOperationService operationService = nodeEngine.getOperationService();
+        List<Future<ClusterMetadata>> futures =
+                members.stream()
+                       .filter(m -> !m.localMember())
+                       .map(m -> {
+                           Future<ClusterMetadata> future = operationService.invokeOnTarget(JetService.SERVICE_NAME,
```
cangencer: variable can be inlined

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -479,6 +491,21 @@ boolean shouldStartJobs() {
...
+                                   new GetClusterMetadataOperation(), m.getAddress());
+                           return future;
+                       })
+                       .collect(toList());
+        return futures.stream().map(f -> uncheckCall(f::get)).allMatch(metaData -> metaData.getState() == clusterState);
```
viliam-durina: I think `scheduleWithRepetition` cancels the repetition if some execution fails. This is likely to fail on cluster changes, we should catch the exceptions.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -479,6 +490,24 @@ boolean shouldStartJobs() {
                 && !partitionService.hasOnGoingMigrationLocal();
     }
 
+    private boolean allMembersHaveSameState(ClusterState clusterState) {
+        // TODO remove once the issue is fixed on the imdg side
```
cangencer: is there an issue created on HZ? if not, could we create and link to it?

gurbuzali:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
gurbuzali: I've set the milestone to 3.12 since 3.12.1 is not created yet
https://github.com/hazelcast/hazelcast-enterprise/issues/2747

gurbuzali:
verify

PR #1278, opened 2019-02-19
JobConfig made IdentifiedDataSerializable and DataSerializerHooks are…

eminn:
… refactored

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
@@ -16,7 +16,6 @@
 
 package com.hazelcast.jet.config;
 
-import com.hazelcast.internal.serialization.impl.SerializationUtil;
 import com.hazelcast.jet.JetInstance;
 import com.hazelcast.jet.Job;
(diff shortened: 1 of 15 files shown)


eminn:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -412,4 +417,86 @@ public JobConfig setInitialSnapshotName(@Nullable String initialSnapshotName) {
...
+    }
+
+    @Override
+    public void writeData(ObjectDataOutput out) throws IOException {
+        out.writeUTF(name);
+        out.writeObject(processingGuarantee);
+        out.writeLong(snapshotIntervalMillis);
+        out.writeBoolean(autoScaling);
+        out.writeBoolean(splitBrainProtectionEnabled);
+        SerializationUtil.writeNullableList(resourceConfigs, out);
```
cangencer: we have built in serializer for list, why not just use writeObject?

eminn: I've forgot that we have them in place

PR #1279, opened 2019-02-19
Simplify pipeline tests

mtopolnik:
Main difference is using a simple custom source instead of the IMap event journal.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/ComputeStageImplBase.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/ComputeStageImplBase.java
@@ -354,13 +354,15 @@ public abstract class ComputeStageImplBase<T> extends AbstractStage {
     }
 
     @Nonnull
+    @SuppressWarnings("unchecked")
     <K, RET> RET attachPartitionedCustomTransform(
             @Nonnull String stageName,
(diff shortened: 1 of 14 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/ComputeStageImplBase.java
```diff
@@ -354,22 +353,23 @@
...
     @Nonnull
+    @SuppressWarnings("unchecked")
     <K, RET> RET attachPartitionedCustomTransform(
             @Nonnull String stageName,
             @Nonnull ProcessorMetaSupplier procSupplier,
             @Nonnull DistributedFunction<? super T, ? extends K> partitionKeyFn
-
     ) {
-        return attach(partitionedCustomProcessorTransform(stageName, transform, procSupplier, partitionKeyFn),
+        DistributedFunction adaptedKeyFn = fnAdapter.adaptKeyFn(partitionKeyFn);
```
viliam-durina: There are few more instances where the keyFn is not adapted, for example in `BatchStageWithKeyImpl.aggregate*`

mtopolnik: Batch stages don't need adaptation, it's only for timestamped items

PR #1287, opened 2019-02-23
Simplify unbounded stream tests

mtopolnik:
Use a bounded streaming source so we can join() the job and 
deterministically assert the output.


mtopolnik:
verify

cangencer:
verify

PR #1291, opened 2019-02-26
Allow zero arguments

viliam-durina:
If no arguments were given to the submitted jar, an array with length
one, containing an empty string was passed to the `main` method, which
could confuse it.


viliam-durina:
verify

PR #1292, opened 2019-02-26
Simplify mapUsingIMap and mapUsingReplicatedMap

cangencer:


cangencer:
Breaking changes: 

* `GeneralStage.mapUsingIMap()` and `GeneralStage.mapUsingRepliactedMap` now take a `lookupKeyFn` along with a `mapFn` and the value for the given key is automatically looked up.

PR #1293, opened 2019-02-26
Renaming of functional interfaces

cangencer:
* Remove all unused functional interfaces
* Rename `DistributedFunction` and similar to `FunctionEx` to help with readability


cangencer:
verify

PR #1298, opened 2019-02-27
WindowResultFunction -> Function<WindowResult>

mtopolnik:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/datamodel/DataModelSerializerHooks.java
```diff
@@ -164,6 +164,49 @@ public void destroy() {
...
+
+        @Override
+        public Serializer createSerializer() {
+            return new StreamSerializer<KeyedWindowResult>() {
+                @Override
+                public void write(ObjectDataOutput out, KeyedWindowResult kwr) throws IOException {
+                    out.writeLong(kwr.start());
+                    out.writeLong(kwr.end());
+                    out.writeObject(kwr.key());
+                    out.writeObject(kwr.result());
```
cangencer: final flag isn't serialized, also add a test to `DataModelSerializerHooksTest`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/datamodel/DataModelSerializerHooks.java
```diff
@@ -164,6 +164,49 @@ public void destroy() {
...
+                }
+
+                @Override
+                public KeyedWindowResult read(ObjectDataInput in) throws IOException {
+                    long start = in.readLong();
+                    long end = in.readLong();
+                    Object key = in.readObject();
+                    Object result = in.readObject();
+                    return new KeyedWindowResult<>(start, end, key, result, true);
```
cangencer: seem flag is hardcoded here as well

hazelcast-jet-core/src/main/java/com/hazelcast/jet/datamodel/KeyedWindowResult.java
```diff
@@ -0,0 +1,76 @@
...
+import static com.hazelcast.jet.impl.util.Util.toLocalTime;
+
+/**
+ * Holds the result of a group-and-aggregate operation performed over a
+ * time window.
+ *
+ * @param <K> type of the grouping key
+ * @param <R> type of the aggregated result
+ */
+public class KeyedWindowResult<K, R> extends WindowResult<R> {
```
cangencer: I think it would be better to implement Map.Entry here, or extend `TimestampedEntry` so that it can be directly written to map sink. WindowResult can be perhaps an interface?

cangencer:
verify

PR #1305, opened 2019-03-01
Moves tasks waiting for remote response

sancar:
Threads that handles client tasks are refactored on the server side.
Generic executor was opening `coreSize * 20` threads when needed
before. This was not desirable. We have moved the blocking ones to
a second new executor with high max thread count and reduced the number
of threads that generic client executor opens to `coreSize`

This pr marks JetMessageTasks waiting for remote response as
BlockingMessageTask. Doing so, in terms of Jet perspective the old
behaviour is preserved.

related to https://github.com/hazelcast/hazelcast/issues/4641


cangencer:
How can we make them non-blocking? I don't think they need to blocking, most of them don't do anything other than fire off an operation. Is there an example?

sancar:
We have worked on analyzing this together with @eminn. Many thanks to him.
It seems that your operations are blocking because they are making use of IMap internally. In hazelcast core side, we don't use distributed data structures as you know. You need to use async API when you touch IMap and handle the response with callbacks.  Or remove the need for using Distributed data structures completely. With your current codebase, it does not seem like an easy task.

PR #1307, opened 2019-03-01
Remove TimestampedItem from API

cangencer:


mtopolnik:
We should remove `TimestampedItem` entirely and use `JetEvent` in the convenient source.

cangencer:
We can do that in a separate PR

PR #1312, opened 2019-03-05
Delete TimestampedItem

mtopolnik:


cangencer:
we could also just emit JetEvent as is to avoid litter? and have the flat mapper check for it?

PR #1315, opened 2019-03-07
Shut down JetClassLoader to avoid leaks

viliam-durina:
When we remove JetClassLoader after the execution is over, we want to
prevent someone still holding reference to it to try to load any class -
it would fail (because resources map was destroyed) and it would
recreate the map.


cangencer:
verify

viliam-durina:
@cangencer Test was added

PR #1320, opened 2019-03-07
Simplify and fix race in job cleanup

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -321,65 +293,64 @@ void completeJob(long jobId, String coordinator, long completionTime, Throwable
...
+    void cleanup(NodeEngine nodeEngine) {
+        Collection<DistributedObject> maps =
+                nodeEngine.getProxyService().getDistributedObjects(MapService.SERVICE_NAME);
+
+        maps.stream()
+                .filter(map -> map.getName().startsWith(SNAPSHOT_DATA_MAP_PREFIX))
+                .forEach(map -> {
+                    long id = jobIdFromMapName(map.getName(), SNAPSHOT_DATA_MAP_PREFIX);
+                    // job is not active
+                    if (!jobRecords.containsKey(id)) {
```
viliam-durina: Might be better to get `jobRecords.keySet()` in advance and then query a local map. We're going to query every key in `jobRecords` three times: each active job has 2 snapshot maps and 1 resource map.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -321,65 +293,64 @@ void completeJob(long jobId, String coordinator, long completionTime, Throwable
...
+                        map.destroy();
+                    }
+                });
+
+        maps.stream()
+                .filter(map -> map.getName().startsWith(RESOURCES_MAP_NAME_PREFIX))
+                .forEach(map -> {
+                    long id = jobIdFromMapName(map.getName(), RESOURCES_MAP_NAME_PREFIX);
+                    // if job is finished, we can safely delete the map
+                    if (jobResults.containsKey(id)) {
```
viliam-durina: Again, pre-loading keys might be better. Except if there are few active jobs and lots of completed jobs. Maybe:
```java
Set<Long> jobResultsKeysLocal = jobResults.size() > 50 * jobRecordKeysLocal.size()
    ? Collections.newSetFromMap(jobResults) : jobResults.keySet()
```

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
cangencer: Yes I thought about this too, it's only an issue if there's a large number of running jobs though.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
cangencer: I don't think it's necessary, in the "happy path" the job is likely to be already running and the map won't be queried.

PR #1322, opened 2019-03-08
Make configureJetService synchronized

cangencer:


viliam-durina:
Is this a fix to the ConcurrentModificationEx because we mutate the config and config is shared?

cangencer:
Yes, it's because of the concurrent modification of Properties

PR #1325, opened 2019-03-11
Subsequent transform fusing

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/Planner.java
```diff
@@ -101,13 +109,109 @@ DAG createDag() {
...
+                    flatMapFn = flatMapFn == null
+                            ? function::apply
+                            : flatMapFn.andThen(r -> r.flatMap(function));
+                }
+                lastFlatMap = i + 1;
+            }
+        }
+
+        FunctionEx trailingMapFn = mergeMapFunctions(chain.subList(lastFlatMap, chain.size()));
+        String name = chain.stream().map(Transform::name).collect(Collectors.joining(", ", "fused(", ")"));
```
cangencer: this could create quite long names, how will look in the Management Center?

viliam-durina: It's an issue of MC imo, it should trim it somehow maybe using `text-overflow:ellipsis` :)
Or the user, seeing it, can rename the the stages. The names can be user-assigned, i didn't want to just replace them or not display all of them...

PR #1332, opened 2019-03-18
Support multigraphs

### LINKED ISSUE #844: Incorrect validation when adding Edge to DAG
Hi,

I see that method **DAG.edge(Edge)** method throws an **IllegalArgumentException** if there already is an edge between the same vertices of the Edge being added.

I have this case however, that makes me think that this kind of edge validation is incorrect:

- a **source** that emits objects of type **A** and **B**
- a **custom router processor**, that routes input objects to **edge 0** if they are of **type A**, to **edge 1** if they are of **type B**
- a **custom calculator processor** that **expects** inputs of **type A on edge 0** and inputs of **type B on edge 1**.
- a **sink**

My DAG therefore has these edges:
- source -> router
- router(0) -> calculator(0)
- router(1) -> calculator(1)
- calculator -> sink

As you can see, there are 2 edges between router and calculator, but they are not an error...they are meant to be like that.

viliam-durina:
Fixes #844


vladoschreiner:
Vilo proč prosímtě implementujeme multigrafy?

cangencer:
This will require changes in Management Center as well. I'll put it to 3.2 for now, but we also need a PRD for it.

cangencer:
These can also be used to implement self-joins, i.e. given a list of people with ids and also a parentId, you can do an aggregation on each parent/child pair.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/DAG.java
```diff
@@ -141,10 +141,6 @@ public DAG edge(@Nonnull Edge edge) {
         if (edge.getDestination() == null) {
```
cangencer: Javadoc says: "It is an error to add an edge that connects the same two vertices as another existing edge." which is a bit confusing. This should be clarified and ref manual also updated if necessary.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/MultigraphTest.java
```diff
@@ -0,0 +1,54 @@
...
+import static java.util.stream.Collectors.toMap;
+import static org.junit.Assert.assertEquals;
+
+public class MultigraphTest extends JetTestSupport {
+
+    @Test
+    public void test() {
+        DAG dag = new DAG();
+        List<Integer> input = IntStream.range(0, 200_000).boxed().collect(Collectors.toList());
+        Vertex source = dag.newVertex("source", ListSource.supplier(input));
```
cangencer: test could rewritten to use the new test sources

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
```diff
@@ -335,10 +335,10 @@ private void initProcSuppliers(long jobId, long executionId) {
 
     private void initDag() {
         final Map<Integer, VertexDef> vMap = vertices.stream().collect(toMap(VertexDef::vertexId, v -> v));
-        vertices.forEach(v -> {
+        for (VertexDef v : vertices) {
```
cangencer: why change the forEach here? technically it provides more efficient iteration then `for ()` since it doesn't create an iterator

viliam-durina: OTOH it doesn't create a lambda, so in terms of allocations it's the same. I like the old school way, it's easier to debug, stack trace is shorter.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/MultigraphTest.java
viliam-durina: The new test sources are for pipeline API. I would have to use:
```java
((BatchSourceTransform) TestSources.items(input)).metaSupplier
```

cangencer: I think we can create processor version of those sources and get rid of ListSource

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/DAG.java
viliam-durina: I updated this. I didn't find a good place to update the refman, I don't think it's important to mention there.

PR #1337, opened 2019-03-20
add JetConfigXmlGenerator

gurbuzali:


cangencer:
could we add some tests?

PR #1342, opened 2019-03-27
Auto-enable hot restart when lossless restart is enabled

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JetConfig.java
```diff
@@ -46,6 +48,13 @@
 
     private static final ILogger LOGGER = Logger.getLogger(JetConfig.class);
 
+    static {
+        // set jet.home to CWD if not already configured
+        if (System.getProperty(JetProperties.JET_HOME.getName()) == null) {
```
mtopolnik: You could extract the constant here

PR #1345, opened 2019-03-31
Two optimizations

mtopolnik:
1. Cache the result of `logger.isFinestEnabled()` in `TaskletExecutionService`
2. Eagerly release the data from `GroupP` while completing

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
@@ -149,9 +149,9 @@ public class TaskletExecutionService {
                 .map(blockingTaskletExecutor::submit)
                 .collect(toList());
 
-        // do not return from this method until all workers have started. Otherwise
+        // Do not return from this method until all workers have started. Otherwise
         // on cancellation there is a race where the executor might not have started
-        // the worker yet. This would results in taskletDone() never being called for
+        // the worker yet. This would result in taskletDone() never being called for
         // a worker.
         uncheckRun(startedLatch::await);
     }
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
```diff
@@ -274,12 +274,13 @@ public void run() {
         public void run() {
             IdleStrategy idlerLocal = idlerCooperative;
             long idleCount = 0;
+            boolean shouldLogFinest = logger.isFinestEnabled();
```
viliam-durina: Logging level can be changed dynamically. As a compromise, we could call `isFinestEnabled` once per iteration of tasklets. Also the name can be `logFinestEnabled`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/GroupP.java
```diff
@@ -82,4 +81,26 @@ protected boolean tryProcess(int ordinal, @Nonnull Object item) {
...
+        public OUT next() {
+            if (iter == null) {
+                iter = keyToAcc.entrySet().iterator();
+            }
+            if (!iter.hasNext()) {
+                return null;
+            }
+            Entry<K, A> e = iter.next();
+            iter.remove();
+            return mapToOutputFn.apply(e.getKey(), aggrOp.finishFn().apply(e.getValue()));
```
viliam-durina: We should handle map-to-null case here, it was handled before in `Traverser.map`. We should also add test for that.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/GroupP.java
```diff
@@ -82,4 +81,26 @@ protected boolean tryProcess(int ordinal, @Nonnull Object item) {
...
+        private final BiFunction<? super K, ? super R, OUT> mapToOutputFn;
+        private Iterator<Entry<K, A>> iter;
+
+        ResultTraverser(BiFunction<? super K, ? super R, OUT> mapToOutputFn) {
+            this.mapToOutputFn = mapToOutputFn;
+        }
+
+        @Override
+        public OUT next() {
+            if (iter == null) {
```
viliam-durina: We can move this to the constructor and construct the `resultTraverser` in `complete`, if it's null. This would be less checking.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
viliam-durina: What about this? I had issues in the past when a production system got into a bad state and I wanted to enable logging at runtime, but didn't work because the code extracted the value in similar way...

PR #1350, opened 2019-04-03
Fixed nightly profile

olukas:
It fixes or adjusts following:
* categories `NightlyTest` and `SlowTest` were not executed in `nightly` profile
* run only categories `NightlyTest` and `SlowTest` in nightly profile (i.e. not run common tests in this profile)
* `test-coverage` used for Sonar should include also `NightlyTest` and `SlowTest` categories since they are also related to test coverage
* `redirectTestOutputToFile` is moved to property to have it configured automatically for all test-related profiles
* removed `failIfNoTests=false` from surefire plugin configuration - it can cause that no tests were run and it looks like everything finished ok. In case when anybody needs it for locally testing etc., then `-DfailIfNoTests=false` can be used in maven command 
* moved tests `MulticastDiscoveryTest`, `BackpressureTest` and `WordCountTest` from nightly profile to common tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -124,6 +124,8 @@
         <sonar.java.coveragePlugin>jacoco</sonar.java.coveragePlugin>
         <sonar.language>java</sonar.language>
         <sonar.verbose>true</sonar.verbose>
+
+        <maven.test.redirectTestOutputToFile>true</maven.test.redirectTestOutputToFile>
     </properties>
 
     <licenses>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -458,9 +458,20 @@
...
-                        <configuration>
+                        <configuration combine.self="override">
+                            <runOrder>failedfirst</runOrder>
                             <forkCount>1</forkCount>
-                            <excludedGroups></excludedGroups>
+                            <trimStackTrace>false</trimStackTrace>
+                            <argLine>
+                                ${argLine}
+                            </argLine>
+                            <includes>
```
cangencer: this seems redundant, it would be inherited from parent

pom.xml
```diff
@@ -485,12 +496,17 @@
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-surefire-plugin</artifactId>
-                        <configuration>
-                            <testFailureIgnore>true</testFailureIgnore>
-                            <argLine>@{argLine}</argLine>
+                        <configuration combine.self="override">
```
cangencer: likewise we shouldn't override but inherit from parent here

olukas:
@cangencer Concerns were fixed. Thanks for the review.

olukas:
Previous fail was caused by https://github.com/hazelcast/hazelcast-jet/issues/1357 - I added `Ignore` for that.

PR #1369, opened 2019-04-16
Add `filtering` and `flatMapping` AggrOp wrappers

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
```diff
@@ -376,18 +378,35 @@ private AggregateOperations() {
      * U} to one accepting items of type {@code T} by applying a mapping
      * function to each item before accumulation.
      * <p>
+     * Operation is es
```
cangencer: typo?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
```diff
@@ -379,15 +381,30 @@ private AggregateOperations() {
      * If the {@code mapFn} returns {@code null}, the item won't be aggregated
      * at all. This allows applying a filter at the same time.
      *
+     * <h4>Example</h4>
+     *
+     * <pre>{@code
+     *     // calculate set of surnames in each city
+     *     p.drawFrom(...)
```
cangencer: I think having the pipeline here creates more noise, we could just have the AggregateOperations part, what do you think?

PR #1371, opened 2019-04-16
Yaml Configuration Support

eminn:


pom.xml
```diff
@@ -78,7 +78,7 @@
         <target.dir>target</target.dir>
         <maven.build.timestamp.format>yyyyMMdd</maven.build.timestamp.format>
         <timestamp>${maven.build.timestamp}</timestamp>
-        <hazelcast.version>3.12</hazelcast.version>
+        <hazelcast.version>3.13-SNAPSHOT</hazelcast.version>
```
cangencer: this needs to be 3.12, since 3.1 will be released with 3.12 as a dependency

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/ClasspathXmlJetConfig.java
```diff
@@ -0,0 +1,103 @@
...
+import static com.hazelcast.util.Preconditions.checkTrue;
+
+/**
+ * A {@link JetConfig} which is initialized by loading an XML configuration file from the classpath.
+ *
+ * @see FileSystemXmlJetConfig
+ * @see UrlYamlJetConfig
+ * @see InMemoryYamlJetConfig
+ */
+public class ClasspathXmlJetConfig extends JetConfig {
```
cangencer: I'm not sure if we need this class? I prefer the static methods we have like in `JetConfig.loadFromClasspath`. It's easier to discover them. 
I don't like the idea of extending `JetConfig`. The same applies for the other versions of this class. We deliberately didn't offer these for Jet, and instead only mainly offer `loadFromStream()` which covers more cases.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/config/XmlJetMemberConfigLocator.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    protected boolean locateInWorkDir() {
+        return loadFromWorkingDirectory(HAZELCAST_MEMBER_XML);
+    }
+
+    @Override
+    protected boolean locateOnClasspath() {
+        return loadConfigurationFromClasspath(HAZELCAST_MEMBER_XML)
```
cangencer: enterprise should have higher priority since it will have both in classpath

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/ClasspathXmlJetConfig.java
eminn: The approach that you are suggesting makes implementation more complicated around the `JetConfig` class since each option has to work it's way around whether it is loading XML or YAML. For example, in the `loadFromStream()`  we don't know whether it is a XML or YAML stream.

cangencer: we could have loadXmlFromStream and so on, and deprecate the current methods?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/config/XmlJetMemberConfigLocator.java
eminn: I've just double-checked with the current master, this just preserves the current behavior.

See -> https://github.com/hazelcast/hazelcast-jet/blob/master/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/config/XmlJetConfigLocator.java#L64

cangencer: yes, you are right I thought this was the default one but it's not.

eminn:
> It seems to me there is way too much duplication in the tests. It will be a real pain to update all these tests when there is a new property. For example I can find 26 different instances where we validate MetricsConfig differently.

 The reason is that we have way too many different methods to load configuration. We don't need to verify newly introduced properties on each loading mechanism.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JetConfigBuilder.java
```diff
@@ -0,0 +1,33 @@
...
+
+package com.hazelcast.jet.config;
+
+/**
+ * Interface for all config builders.
+ *
+ * @see com.hazelcast.jet.impl.config.XmlJetConfigBuilder
+ * @see com.hazelcast.jet.impl.config.YamlJetConfigBuilder
+ */
+public interface JetConfigBuilder {
```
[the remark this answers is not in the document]
eminn: Since we are using concrete implementations directly, we can remove it.

cangencer:
> > It seems to me there is way too much duplication in the tests. It will be a real pain to update all these tests when there is a new property. For example I can find 26 different instances where we validate MetricsConfig differently.
> 
> The reason is that we have way too many different methods to load configuration. We don't need to verify newly introduced properties on each loading mechanism.

I think you can still consolidate the tests, for example there's a lot of tests like this:
```
 @Override
    @Test(expected = IllegalArgumentException.class)
    public void when_fileSystemNullProperties_thenThrowsException() throws Exception {
        JetConfig.loadFromFileSystem(new File("test"), null);
    }

```
but you can just make the function to get JetConfig a constructor parameter for the abstract class

eminn:
> > > It seems to me there is way too much duplication in the tests. It will be a real pain to update all these tests when there is a new property. For example I can find 26 different instances where we validate MetricsConfig differently.
> > 
> > 
> > The reason is that we have way too many different methods to load configuration. We don't need to verify newly introduced properties on each loading mechanism.
> 
> I think you can still consolidate the tests, for example there's a lot of tests like this:
> 
> ```
>  @Override
>     @Test(expected = IllegalArgumentException.class)
>     public void when_fileSystemNullProperties_thenThrowsException() throws Exception {
>         JetConfig.loadFromFileSystem(new File("test"), null);
>     }
> ```
> but you can just make the function to get JetConfig a constructor parameter for the abstract class

Extracted a new parameterized test to cover those cases. See `JetConfigIllegalArgumentsTest`

PR #1383, opened 2019-04-29
Fix regression in declarative config file resolution

### LINKED ISSUE #1380: hazelcast.jet.config is ignored if given file does not have file extension "xml", "yaml" or "yml"
Jet counterpart of https://github.com/hazelcast/hazelcast/issues/14924

eminn:
Changed behavior to treat an XML configuration file which has a non .xml file extension as XML files to satisfy backwards compatibility.

YAML configuration has a stricter requirement on file extensions
which only accepts .yaml or .yml extension.

Fixes #1380

Depends on https://github.com/hazelcast/hazelcast/pull/14945


eminn:
verify

PR #1390, opened 2019-05-07
Fixed test-coverage profile

olukas:
`argLine` property was not correctly passed to jacoco plugin which caused that Sonar build was not able to show code coverage.


olukas:
@cangencer yes, it was  regression which was caused by https://github.com/hazelcast/hazelcast-jet/commit/8a61eedb23930b5d4b902ec60522bd79a9ea3639. I did not know that `${argLine}` will not be passed correctly to jacoco plugin.

PR #1406, opened 2019-05-22
Add samples to javadoc in AggregateOperations

### LINKED ISSUE #1363: Add code samples in javadoc for AggregateOperations
Similar for pipeline operations, there should be code samples for usage of AggregateOperations.

viliam-durina:
Fixes #1363

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
@@ -36,8 +36,9 @@ import com.hazelcast.jet.function.SupplierEx;
 import com.hazelcast.jet.function.ToDoubleFunctionEx;
 import com.hazelcast.jet.function.ToLongFunctionEx;
 import com.hazelcast.jet.function.TriFunction;
+import com.hazelcast.jet.pipeline.BatchStage;
+import com.hazelcast.jet.pipeline.BatchStageWithKey;
 import com.hazelcast.jet.pipeline.GeneralStage;
-import com.hazelcast.jet.pipeline.StageWithKeyAndWindow;
 import com.hazelcast.jet.pipeline.StageWithWindow;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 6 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
```diff
@@ -251,7 +298,14 @@ private AggregateOperations() {
      * Returns an aggregate operation that computes the arithmetic mean of the
      * {@code long} values it obtains by applying {@code getLongValueFn} to
      * each item.
+     * <p>
```
vladoschreiner: Please add:
- code sample
- SeeAlso averagingDouble, averagingInt and when to use them (it's the input value that should be long or it;'s the output value of the aggregation?)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
```diff
@@ -323,6 +382,13 @@ private AggregateOperations() {
      * approximates the rate of change of {@code y} as a function of {@code x},
      * where {@code x} and {@code y} are {@code long} quantities obtained
      * by applying the two provided functions to each item.
+     * <p>
```
vladoschreiner: Code sample would be lovely.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
```diff
@@ -416,12 +482,12 @@ private AggregateOperations() {
      * See also {@link #filtering filtering()} and {@link #flatMapping
      * flatMapping()}.
      *
+     * @param mapFn the function to apply to input items
```
vladoschreiner: Extend the comment in the code sample:
// calculate the set of surnames from the stream of Persons

hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
```diff
@@ -705,15 +821,28 @@ private AggregateOperations() {
      * {@code HashMap} where the key is the result of applying {@code keyFn}
      * and the value is a list of the items with that key.
      * <p>
-     * This operation achieves the effect of a cascaded group-by where the
+     * Sample usage:
+     * <pre>{@code
+     *     // the result will be a single instance of HashMap<SensorId, List<Measurement>>
+     *     AggregateOperations.groupingBy(Measurement::getSensorId)
```
vladoschreiner: The groupingBy aggregate operation seems quite confusing as one might use it instead of .groupingKey and lose the advantages of parallel aggregation. 
Could we make it explicit in the JavaDoc that .groupingKey should be always used before .groupingBy, including the reasoning?

viliam-durina:
verify

Issue #1348, opened 2019-04-02
Exception for invalid snapshot name is too verbose


cangencer:
```
com.hazelcast.jet.JetException: State for job 'Trade Analyser', execution ae85-41b5-8dff-42ff
 was supposed to be restored from '__jet.exportedSnapshot.snapshot-4', but 
that map doesn't contain the validation record: not an IMap with Jet snapshot or corrupted
```

We should just say specified snapshot name "snapshot-4" doesn't exist.

PR #1408, opened 2019-05-23
Improve error messages in SnapshotValidator

mtopolnik:
Fixes https://github.com/hazelcast/hazelcast-jet/issues/1348.

Includes refactoring of `MasterJobContext.tryStartJob()`.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/SnapshotValidator.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/SnapshotValidator.java
@@ -31,33 +31,41 @@ final class SnapshotValidator {
     /**
      * Validates a snapshot with the given id.
      *
-     * @param snapshotId snapshot ID or {@link JobExecutionRecord#NO_SNAPSHOT}, if snapshot id is
-     *                  not known
-     * @param jobIdString name and id of the job (for debug output)
-     * @param map snapshot map to validate
-     * @return the snapshot id of the snapshot being validated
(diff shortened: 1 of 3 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/SnapshotValidator.java
```diff
@@ -31,30 +31,33 @@ private SnapshotValidator() {
...
      * @param map snapshot map to validate
      * @return the snapshot id of the snapshot being validated
      */
     static long validateSnapshot(long snapshotId, String jobIdString, IMap<Object, Object> map) {
         SnapshotValidationRecord validationRecord = (SnapshotValidationRecord) map.get(SnapshotValidationRecord.KEY);
         if (validationRecord == null) {
-            throw new JetException("State for " + jobIdString + " was supposed to be restored from '" + map.getName()
-                    + "', but that map doesn't contain the validation record: not an IMap with Jet snapshot or corrupted");
+            throw new JetException(String.format(
+                    "snapshot-%d doesn't exist or is damaged. Unable to restore the state for %s.",
```
cangencer: this doesn't actually include the name of the snapshot the user requested

viliam-durina: We don't have snapshot name here, only the map name. The map name can also be the name of one of the runtime maps. I prefer keeping the old message: it says what we know happend, not what might be the case.

PR #1412, opened 2019-05-24
Add check for matching expected Hazelcast version with actual version in classpath

### LINKED ISSUE #1405: Fail fast if another version of HZ than expected is in classpath
When you start a Jet node and there is another version of Hazelcast in the classpath than the expected the node should refuse to start with an exception message. This frequently happens due to some dependency configuration error.

cangencer:
It can be that different version of Hazelcast is included as a dependency due to misconfiguration. This can cause unexpected behaviour or confusing errors. Jet should fail fast in this scenario and refuse to start with a clear exception message.

Fixes #1405

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Jet.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Jet.java
@@ -137,7 +137,7 @@ public final class Jet {
      */
     private static void assertHazelcastVersion() {
         String hzVersion = BuildInfoProvider.getBuildInfo().getVersion();
-        try (InputStream resource = Jet.class.getClassLoader().getResourceAsStream("jet-runtime.properties")){
+        try (InputStream resource = Jet.class.getClassLoader().getResourceAsStream("jet-runtime.properties")) {
             Properties p = new Properties();
             p.load(resource);
             String jetHzVersion = p.getProperty("jet.hazelcast.version");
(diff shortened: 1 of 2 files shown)


cangencer:
I'm not sure how to write a unit test for this, but I tested it with a project that included a dependency to hazelcast 3.11. Would be great if someone else could test it before approving.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Jet.java
```diff
@@ -125,6 +132,27 @@ static JetClientInstanceImpl getJetClientInstance(HazelcastInstance client) {
...
+     * Makes sure that the Jet-expected Hazelcast version and the one in classpath match
+     */
+    private static void assertHazelcastVersion() {
+        String hzVersion = BuildInfoProvider.getBuildInfo().getVersion();
+        try (InputStream resource = Jet.class.getClassLoader().getResourceAsStream("jet-runtime.properties")) {
+            Properties p = new Properties();
+            p.load(resource);
+            String jetHzVersion = p.getProperty("jet.hazelcast.version");
+            if (!hzVersion.equals(jetHzVersion)) {
+                throw new JetException("Jet uses Hazelcast version " + jetHzVersion + " however " +
```
eminn: I think we should replace Hazelcast and hazelcast occurences with Hazelcast IMDG for better separation

cangencer: I'm not sure, as the dependency is called "hazelcast.jar"

PR #1421, opened 2019-05-30
Development sources

jerrinot:
Easy-to-use sources intended to be used during pipeline development.
It's often convenient to have easy-to-use sources with well defined
behaviour when developing a streaming application.

I am not sure if this belongs here or rather to the https://github.com/hazelcast/hazelcast-jet-contrib repo. 
I suspect you almost always need some easy-to-use source when starting new development with Jet.


cangencer:
how about making these composable? so you can decorate a source to limit rates and what not? We used something similar in tests, see `ThrottleWrappedP`

jerrinot:
that's something I was considering. but then it's harder to use and simplicity is THE reason to have this. there is something nice when you can do:
```
    @Test
    public void fixedRate() {
        Pipeline pipeline = Pipeline.create();

        pipeline.drawFrom(DevelopmentSources.fixedRate(1, MICROSECONDS))
                .withIngestionTimestamps()
                .window(tumbling(1000))
                .aggregate(counting())
                .drainTo(logger());

        jet().newJob(pipeline).join();
    }
```
and it works. no dealing with any builder, etc. But perhaps as with https://github.com/hazelcast/hazelcast-jet-contrib/pull/11 I could extract the components, make them reusable and still have some super-simple convenience.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/DevSources.java
```diff
@@ -0,0 +1,187 @@
...
+    private static final AtomicLong COUNTER = new AtomicLong();
+    private static final String NAME_PREFIX = "dev-source-";
+
+    private DevSources() {
+
+    }
+
+    /**
+     * Streaming source which emit with fixed rate. It attempts to compensate for various system hiccups so the rate
+     * over a period of time is constant.
```
mtopolnik: The doc doesn't state what the source is actually emitting.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/DevSources.java
```diff
@@ -0,0 +1,187 @@
...
+     * @param items values to emit
+     * @param <T> type of the element
+     * @return source emitting all specified items
+     */
+    public static <T> BatchSource<T> of(T... items) {
+        return fromIterator(() -> Arrays.asList(items).iterator());
+    }
+
+    private static <T> StreamSource<T> triggerDrivenStreamSource(SupplierEx<Trigger> triggerSupplier,
+                                                                 SupplierEx<T> itemSupplier) {
```
mtopolnik: Our convention is to put the parameter list on its own line when it doesn't fit all into one line.

```
private static <T> StreamSource<T> triggerDrivenStreamSource(
        SupplierEx<Trigger> triggerSupplier, SupplierEx<T> itemSupplier
) {
```

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/DevSources.java
```diff
@@ -0,0 +1,187 @@
...
+     * You have to provide your own item supplier.
+     * The source is not distributed it means only a single Jet instance will emit items.
+     *
+     * @param period period which to emit
+     * @param timeUnit units for period
+     * @param itemSupplier supplier of items to be emitted. It's called whenever a new item is about to be emitted
+     * @param <T> type of emitted items
+     * @return source emitting with a fixed rate
+     */
+    public static <T> StreamSource<T> fixedRate(long period, TimeUnit timeUnit, SupplierEx<T> itemSupplier) {
```
mtopolnik: All functional interfaces must specify variance, in a boilerplate fashion (PECS). So, `SupplierEx<? extends T>` in this case.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/DevSources.java
mtopolnik: Also, `@Nonnull` must be everywhere, on the return values as well.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/DevSources.java
```diff
@@ -0,0 +1,187 @@
...
+public final class DevSources {
+    private static final AtomicLong COUNTER = new AtomicLong();
+    private static final String NAME_PREFIX = "dev-source-";
+
+    private DevSources() {
+
+    }
+
+    /**
+     * Streaming source which emit with fixed rate. It attempts to compensate for various system hiccups so the rate
```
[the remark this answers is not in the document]
jerrinot: is there a way to teach Intellij to do this for you?

jerrinot:
@cangencer I pushed a new version. Arguably it composes better. but I am not sure if it's an improvement:) See the [commit comments](https://github.com/hazelcast/hazelcast-jet/pull/1421/commits/5f08654ab352785ed776041581f6240d642a475a)

The other option is to take the [previous revision](https://github.com/hazelcast/hazelcast-jet/blob/82a490ecbb044e1eb1ffc3fd8f6dd04560eb8e9a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/DevSources.java) and just just make the `Trigger`, `fromIterator()` and `triggerDriven()` public. (+ document them and perhaps give them better names).

WDYT?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/DevSources.java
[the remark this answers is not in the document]
viliam-durina: I use [Wrap to Column](https://plugins.jetbrains.com/plugin/7234-wrap-to-column) plugin, i have it configured to wrap at 79 characters. This works for method javadocs, class javadocs will be 4 characters longer, because the configured value includes the leading whitespaces.

For deeper indentations, such as when describing method params, I let marko reformat it when reviewing ;-)

hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/DevSourcesTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.junit.Test;
+
+import java.util.UUID;
+
+import static com.hazelcast.jet.aggregate.AggregateOperations.counting;
+import static com.hazelcast.jet.pipeline.Sinks.logger;
+import static com.hazelcast.jet.pipeline.WindowDefinition.tumbling;
+import static java.util.concurrent.TimeUnit.MICROSECONDS;
+
+public class DevSourcesTest extends PipelineStreamTestSupport {
```
jerrinot: this is WIP, something I used while working on it. obviously not a real test:)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/DevSources.java
```diff
@@ -0,0 +1,244 @@
...
+     * The supplier can return <code>null</code> to indicate no item is
+     * available at the moment. Jet will call the supplier again at some
+     * later time.
+     *
+     * @param supplier supplier providing items to be emitted by the source
+     * @param <T> type of emitted items
+     * @return source emitting items provided by the supplier
+     */
+    @Nonnull
+    public static <T> StreamSource<T> asStreamSource(@Nonnull SupplierEx<? extends T> supplier) {
```
cangencer: how about `generatedStream` ?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/DevSources.java
```diff
@@ -0,0 +1,244 @@
...
+import static com.hazelcast.jet.function.Suppliers.minDelayThrottle;
+
+/**
+ * Easy-to-use sources intended to be used during pipeline development.
+ * It's often convenient to have easy-to-use sources with well defined
+ * behaviour when developing a streaming application.
+ *
+ * @since 3.2
+ */
+public final class DevSources {
```
cangencer: I think `TestSources` is a better name

hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/Suppliers.java
```diff
@@ -0,0 +1,149 @@
...
+     * {@see #minDelayThrottle}
+     *
+     * @param period throttling period
+     * @param timeUnit units for the throttling period
+     * @param delegate supplier to delegate to
+     * @param <T> type of items provided by the supplier
+     * @return throttling supplier
+     */
+    @Nonnull
+    public static <T> SupplierEx<T> fixedRateThrottle(
```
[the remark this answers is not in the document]
jerrinot: would you prefer a separate class in the same package?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/DevSources.java
jerrinot: :+1:

hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/Suppliers.java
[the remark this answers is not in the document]
cangencer: yes, that would be fine.

PR #1438, opened 2019-06-11
Two-phase commit for snapshots

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
```diff
@@ -201,58 +204,174 @@ default boolean completeEdge(int ordinal) {
...
      * <p>
+     * If this processor communicates with an external transactional store, it
+     * should do few more things:
+     * <ul>
+     *     <li>mark the current active transaction with the external system as
+     *     <em>pending</em> and stop using it. The pending transaction will be
+     *     completed when {@link #onSnapshotCompleted} with {@code
+     *     commitTransactions == true} is called
+     *
+     *     <li>store IDs of the pending transaction(s) to she snapshot. Note
```
eminn: she -> the

cangencer:
would it not make sense to have a option to turn on/off 2PC in JobConfig?

cangencer:
> > would it not make sense to have an option to turn on/off 2PC in JobConfig?
> > I don't think so. If we turned it off and there are transactional vertices, they will break. If there are not, the phase-2 is very little CPU will be saved, it's not worth the complexity.

I was thinking mainly in some sinks it could cause additional latency, because you would need to wait for a snapshot to commit.

viliam-durina:
>I was thinking mainly in some sinks it could cause additional latency
The sinks have an option to reduce the guarantee, see [kafka](https://github.com/viliam-durina/hazelcast-jet/blob/5df7e94ce569a5d9ac97f0dbc15a671641d2504e/hazelcast-jet-kafka/src/main/java/com/hazelcast/jet/kafka/KafkaSinks.java#L285-L307) or [file](https://github.com/viliam-durina/hazelcast-jet/blob/14a5a7636c4708bb14425816246e62cd0f4bda43/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/FileSinkBuilder.java#L98-L120) sink.

viliam-durina:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorState.java
```diff
@@ -30,9 +31,13 @@
     PROCESS_WATERMARK,
 
     /**
-     * Making calls to {@link Processor#tryProcess()} and {@link
-     * Processor#process(int, com.hazelcast.jet.core.Inbox)} until the inbox
-     * is empty.
+     * Making calls to {@link Processor#tryProcess()} until it returns true.
+     */
+    NULLARY_TRY_PROCESS,
```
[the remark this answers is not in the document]
viliam-durina: People who don't understand it should ignore it  :-)
We use this term for some time already to distinguish the nullary `tryProcess` from the non-nullary one in `AbstractProcessor`. The term is internal, so no worry about confusing the users.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterSnapshotContext.java
```diff
@@ -111,66 +114,73 @@ void startScheduledSnapshot(long executionId) {
         } finally {
             mc.unlock();
         }
-        tryBeginSnapshot();
+        tryBeginSnapshotPhase1();
```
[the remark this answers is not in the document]
viliam-durina: In similar logic we could rename `onSnapshotPhase2Completed` to `onSnapshotCompleted`, because it's the last phase, but I think it's clearer the current way and not too long.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/SnapshotContext.java
```diff
@@ -46,21 +48,30 @@
      * decremented as the tasklets complete (this is when they receive
      * DONE_ITEM and after all pending async ops completed).
      */
-    private int numTasklets = Integer.MIN_VALUE;
+    private int numSsTasklets = Integer.MIN_VALUE;
```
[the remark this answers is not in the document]
viliam-durina: Also `numPrioritySsTasklets` to `numPriorityStoreSnapshotTasklets`? These are used a lot and documented, I'd rather keep them as they are.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/SnapshotContext.java
```diff
@@ -134,66 +173,120 @@ boolean isTerminalSnapshot() {
...
+                        + snapshotId;
+        assert numPTasklets >= 0 : "numPTasklets=" + numPTasklets;
+        if (isCancelled) {
+            throw new CancellationException("execution cancelled");
+        }
+        this.lastPhase1Successful = success;
+        assert numPrioritySsTasklets == 0 : "numPrioritySsTasklets=" + numPrioritySsTasklets;
+
+        boolean casSuccess = numRemainingTasklets.compareAndSet(0, numPTasklets);
+        assert casSuccess : "numRemainingTasklets wasn't 0, but " + numRemainingTasklets.get();
```
cangencer: this should be a failure, not just assertion

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
```diff
@@ -543,12 +543,18 @@ public long lastSnapshotId() {
         return lastSnapshotId;
     }
 
-    public int getStoreSnapshotTaskletCount() {
+    public int getSsTaskletCount() {
```
cangencer: I would prefer this to be written out as before

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
```diff
@@ -543,12 +543,18 @@ public long lastSnapshotId() {
...
         return (int) tasklets.stream()
                              .filter(t -> t instanceof StoreSnapshotTasklet)
                              .count();
     }
 
+    public int getPTaskletCount() {
```
cangencer: `getProcessorTaskletCount`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/SnapshotContext.java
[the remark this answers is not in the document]
cangencer: It's ok to use it like this here, but not in `ExecutionPlan` IMHO

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/operation/SnapshotCompleteOperation.java
```diff
@@ -0,0 +1,88 @@
...
+import com.hazelcast.jet.impl.execution.init.JetInitDataSerializerHook;
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
+
+import static com.hazelcast.jet.impl.util.LoggingUtil.logFine;
+
+public class SnapshotCompleteOperation extends AsyncJobOperation {
```
cangencer: would it better to call this SnapshotPhase1CompleteOperation? or at least document it as such.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/TwoPhaseSnapshotCommitUtility.java
```diff
@@ -0,0 +1,427 @@
...
+     * Call from {@link Processor#close()}.
+     *
+     * <p>The implementation must not commit or rollback any pending transactions
+     * - the job might have failed between after snapshot phase 1 and 2. The
+     * pending transactions might be recovered after the restart.
+     */
+    public abstract void close() throws Exception;
+
+    protected boolean usesTransactionLifecycle() {
+        return externalGuarantee == EXACTLY_ONCE ||
```
cangencer: I see these two variables only used here, but this method is not used. Are these needed?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/TwoPhaseSnapshotCommitUtility.java
```diff
@@ -0,0 +1,427 @@
...
+import javax.annotation.Nullable;
+import java.util.Map.Entry;
+import java.util.function.Consumer;
+
+import static com.hazelcast.jet.config.ProcessingGuarantee.AT_LEAST_ONCE;
+import static com.hazelcast.jet.config.ProcessingGuarantee.EXACTLY_ONCE;
+import static com.hazelcast.jet.impl.util.ExceptionUtil.sneakyThrow;
+
+/**
+ * A base class for transaction utilities implementing different transaction
```
cangencer: documentation could be improved here.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/TwoPhaseSnapshotCommitUtility.java
```diff
@@ -0,0 +1,427 @@
...
+    private final FunctionEx<TXN_ID, LoggingNonThrowingResource<TXN_ID, RES>> createTxnFn;
+    private final Consumer<TXN_ID> recoverAndCommitFn;
+    private final ConsumerEx<Integer> recoverAndAbortFn;
+
+    /**
+     * @param outbox the outbox passed to the {@link Processor#init} method
+     * @param procContext the context passed to the {@link Processor#init} method
+     * @param isSource true, if the processor is a source (a reader), false for
+     *      a sink (a writer)
+     * @param externalGuarantee guarantee required for the source/sink. Must
```
cangencer: what does this mean? is this the guarantee provided by the source/sink implemenation?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/SnapshotContext.java
viliam-durina: There's no way the user can cause it. I hit it during development. It's a check for further changes that we expect will be met.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/TwoPhaseSnapshotCommitUtility.java
viliam-durina: The whole utility is unused here. Also it's abstract. It is used in `2pc-file` and `2pc-kafka` branches, in both of them.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/TwoPhaseSnapshotCommitUtility.java
viliam-durina: Yes. It can be lower than job's, but not higher.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
```diff
@@ -246,63 +251,34 @@ private void stateMachineStep() {
...
 
             case COMPLETE_EDGE:
-                progTracker.notDone();
                 if (isSnapshotInbox()
                         ? processor.finishSnapshotRestore() : processor.completeEdge(currInstream.ordinal())) {
                     assert !outbox.hasUnfinishedItem() || !isSnapshotInbox() :
                             "outbox has an unfinished item after successful finishSnapshotRestore()";
                     progTracker.madeProgress();
-                    state = initialProcessingState();
+                    state = processingState();
```
jbartok: I think it would help readability if the insides of this `processingState` method would be exposed. I'm thinking something like `state = inboundOrdinalsLeft() ? PROCESS_WATERMARK : COMPLETE`. That's something you can immediately understand without needing to look into a method body first.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
```diff
@@ -233,11 +239,10 @@ public MetricsContext getMetricsContext() {
     private void stateMachineStep() {
         switch (state) {
```
jbartok: This switch statement is a bit of a monster... Extracting some methods comes to mind, but I can imagine that actually hurting readability. Give it a try, if you feel like it?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
```diff
@@ -314,47 +290,46 @@ private void stateMachineStep() {
...
+                if (processor.onSnapshotCompleted(ssContext.isLastPhase1Successful())) {
+                    pendingSnapshotId2++;
+                    ssContext.phase2DoneForTasklet();
                     progTracker.madeProgress();
-                    state = EMIT_DONE_ITEM;
+                    state = state == ON_SNAPSHOT_COMPLETED_BEFORE_END
+                            ? EMIT_DONE_ITEM
+                            : processingState();
+                }
+                return;
```
jbartok: This is very confusing code. I understand it now and it's probably perfectly correct but it's mad scientist code. It took me a while to understand that the handling of the two states actually differs and that `state = state ==` is also confusing at a first glance. (Btw. do my comments at least give you insight into how less intelligent people perceive the world?)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
```diff
@@ -314,47 +290,46 @@ private void stateMachineStep() {
...
+            case WAITING_FOR_SNAPSHOT_COMPLETED:
+                long currSnapshotId2 = ssContext.activeSnapshotIdPhase2();
+                if (currSnapshotId2 >= pendingSnapshotId2) {
+                    state = ON_SNAPSHOT_COMPLETED_BEFORE_END;
+                    stateMachineStep(); // recursion
                 }
                 return;
 
+            case COMPLETE:
+                complete();
```
jbartok: We should probably have each clause in the switch statement at the same semantic level. Either inline everything in each, or methods in each. Basically same as my previous comment about refactoring the switch.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/TwoPhaseSnapshotCommitUtility.java
```diff
@@ -0,0 +1,427 @@
...
+import static com.hazelcast.jet.impl.util.ExceptionUtil.sneakyThrow;
+
+/**
+ * A base class for transaction utilities implementing different transaction
+ * strategies.
+ *
+ * @param <TXN_ID> type fo the transaction ID
+ * @param <RES> type of the transactional resource
+ */
+public abstract class TwoPhaseSnapshotCommitUtility<TXN_ID extends TransactionId,
```
jbartok: No usages anywhere. If it has some on other branches, then maybe move the code there too?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/operation/SnapshotOperation.java
```diff
@@ -60,16 +60,16 @@ public SnapshotOperation(long jobId, long executionId, long snapshotId, String m
         ExecutionContext ctx = service.getJobExecutionService().assertExecutionContext(
                 getCallerAddress(), jobId(), executionId, getClass().getSimpleName()
         );
-        CompletableFuture<SnapshotOperationResult> future = ctx.beginSnapshot(snapshotId, mapName, isTerminal)
+        CompletableFuture<SnapshotOperationResult> future = ctx.beginSnapshotPhase1(snapshotId, mapName, isTerminal)
```
jbartok: Wouldn't make more sense to rename `SnapshotOperation` to `SnapshotPhase1Operation` and `SnapshotCompleteOperation` to `SnapshotPhase2Operation`?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorState.java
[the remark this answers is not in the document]
viliam-durina: Well readability is subjective. We added "nullary" because without it we confused it for the non-nullary `tryProcess` in  `AbstractProcessor`.

The "try" part means the method returns boolean if it cannot finish now. I'd rather remove that part: `NULLARY_PROCESS`. It's the only "try" in `ProcessorState`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobExecutionRecord.java
```diff
@@ -320,6 +320,11 @@ public long endTime() {
             return endTime;
         }
 
+        /**
+         * The time elapsed between phase-1 start and phase-1 success. Doesn't
+         * include the phase-2 - the phase-2 isn't required for the snapshot to
```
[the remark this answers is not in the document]
viliam-durina: I'm not sure, i think it should be there. @mtopolnik, wdyt?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJobProxy.java
```diff
@@ -147,6 +147,10 @@ public void restart() {
         terminate(TerminationMode.RESTART_GRACEFUL);
     }
 
+    public void restart(boolean graceful) {
```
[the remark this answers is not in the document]
viliam-durina: Yes, it's used in the other PRs. 
We also have other tests that terminate members in order to simulate abrupt failures, this method might do the same faster.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
```diff
@@ -252,7 +257,19 @@ default void join() {
      * won't automatically delete the IMap it is exported into. You must
      * manually call {@linkplain JobStateSnapshot#destroy() snapshot.destroy()}
      * to delete it. If your state is large, make sure you have enough memory
-     * to store it.
+     * to store it. The snapshot created using this method will also not be
+     * used for automatic restart - should the job fail, the previous
+     * automatically started snapshot will be used.
```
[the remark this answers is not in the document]
viliam-durina: I think both _saved_ and _started_ work here. This is from javadoc of `exportSnapshot`, which _manually starts_ a snapshot, contrary to an _automatic start_, so using _start_ seems better to me.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
```diff
@@ -252,7 +257,19 @@ default void join() {
...
+     * <p>
+     * For transactional sources or sinks (that is those which use transactions
+     * to confirm reads or to commit writes), Jet will not commit the
+     * transactions when creating a snapshot using this method. The reason for
```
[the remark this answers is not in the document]
viliam-durina: It indeed does. That's why I proposed to remove this method entirely, but we need to discuss it. 

Copying the last snapshot (if there is one) would not solve it. Correctness of 2pc requires that you restore from the latest snapshot, not some older one. Two things can happen:
- before you manage to restore from the exported snapshot, another snapshot could be done and the exported one is stale
- job fails after snapshot was exported, but before it does another automatic snapshot. It will restart from the automatic snapshot done before the exported snapshot, which is again wrong.

This method just doesn't fit and is not useful anyway, even without transactional processors.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterSnapshotContext.java
```diff
@@ -199,7 +208,12 @@ private void onSnapshotCompleted(
                         mergedResult.getNumChunks(), mergedResult.getNumBytes(),
                         mc.jobExecutionRecord().ongoingSnapshotStartTime(), mc.jobId(), mc.jobName(),
                         mc.jobRecord().getDagJson());
+
+                // The decision moment for exported snapshots: after this the snapshot is valid to be restored
+                // from, however it will be not listed by JetInstance.getJobStateSnapshots unless the validation
```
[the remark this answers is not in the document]
viliam-durina: "to be restored from" similar to "the place will be good to go to" or "a problem is ready to work on"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
```diff
@@ -197,58 +200,177 @@ default boolean completeEdge(int ordinal) {
...
 
     /**
-     * Stores its snapshotted state by adding items to the outbox's {@linkplain
-     * Outbox#offerToSnapshot(Object, Object) snapshot bucket}. If this method
-     * returns {@code false}, it will be called again before proceeding to call
-     * any other method.
+     * Stores the processor's state to a state snapshot by adding items to the
+     * outbox's {@linkplain Outbox#offerToSnapshot(Object, Object) snapshot
+     * bucket}. If this method returns {@code false}, it will be called again
+     * before proceeding to call any other method.
```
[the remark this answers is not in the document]
viliam-durina: You "proceed to call", not "proceed to calling".

And "any methods"? Could be "other methods" (without any), but the meaning is same.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
viliam-durina: It's used 3 times, i think it makes sense to have it.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
viliam-durina: ok, will refactor it to `if` :)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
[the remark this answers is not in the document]
cangencer: It's still useful for log based sources i.e. Kafka, but yes it's a fair point. It seems better to always use exportAndCancel at the same time.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
```diff
@@ -197,58 +200,177 @@ default boolean completeEdge(int ordinal) {
...
+     * transactional processors.
+     * <p>
+     * The default implementation takes no action and returns {@code true}.
+     *
+     * @param commitTransactions true, if all members were successful in {@link
+     *      #saveToSnapshot()} and we're not doing {@link Job#exportSnapshot}
+     * @return {@code true} if this step is done, {@code false} to call this
+     *      method again
+     */
+    default boolean onSnapshotCompleted(@SuppressWarnings("unused") boolean commitTransactions) {
```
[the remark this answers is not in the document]
cangencer: we can clarify that this should only need to be implemented for sources/sink which need to coordinate with external entities. And also link to the utility.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorState.java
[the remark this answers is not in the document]
mtopolnik: If i saw the name `TRY_PROCESS`, I'd automatically think it refers to the one taking parameters, because that is the "main" `tryProcess`, while the nullary one is needed only for special purposes that aren't data-driven. This name does use a less-frequent term, but it's still well in the "easily googlable" category (typing "nullary" gives me "nullary function" as the first suggestion).

I would also not like to see `NULLARY_PROCESS`, I'd think "process" is a noun here. So at least for me the current name works the best.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobExecutionRecord.java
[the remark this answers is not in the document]
mtopolnik: I think you can remove them as well because "phase-2" works like a name. If you wrote "the 2nd phase" then I'd think it would have to stay.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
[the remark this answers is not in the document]
mtopolnik: It confuses me a bit to have "started" because it hints at the possibility of it not having yet completed.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
mtopolnik: The shape of this code is the unfortunate consequence of Java not having coroutines. In the processors, where state changes are data driven, we can usually survive using FP idioms, but this logic is too complex and needs a full state machine.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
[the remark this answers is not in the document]
jbartok: First part: I still think w/ "ing" would be better (https://english.stackexchange.com/questions/76004/proceed-to-doing-something-or-proceed-to-do-something)

Second part: I commented about the "s" missing, not the "any", but this indeed makes sense only in the "ing" version.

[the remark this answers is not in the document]
mtopolnik: "Proceed to calling" sounds ok and "proceeding to call" sounds ok. But "proceeding to calling" sounds weird to me.

PR #1394, opened 2019-05-14
Add fault-tolerant custom source

viliam-durina:

PR #1450, opened 2019-06-18
Add release notes and changelog

cangencer:


olukas:
Please see https://github.com/hazelcast/hazelcast-jet/pull/1394#issuecomment-493958041 - based on that PR `buffer.close()` is not allowed for streaming custom source from 3.1. It was possible in 3.0. In case when somebody has application which calls `buffer.close()` it will stop to work in 3.1. I believe it should be explicitly mentioned in release notes.

Also `convenientSourceP` and `convenientTimestampedSourceP` methods of `SourceProcessors` have changed signature. AFAIK we can break backward compatibility in Core API but we should mentioned it somewhere - I think release notes should be that place, shouldn't it?

cangencer:
@olukas Good point, I've added the list of breaking changes now.

hazelcast-jet-distribution/src/main/resources/release_notes.txt
```diff
@@ -5,6 +5,47 @@ issues at https://github.com/hazelcast/hazelcast-jet/issues/<issue number>.
...
 
+** Hazelcast Jet 3.1 **
+
+1. New Features
+
+[pipeline-api] Enable composite transforms using `apply` operator [#1352]
+[pipeline-api] Add support for fault tolerant sources to SourceBuilder API [#1394]
+[core] YAML configuration support [#1371]
+[integration] Source and sink for Elasticsearch [https://github.com/hazelcast/hazelcast-jet-contrib/tree/master/influxdb]
+[integration] Source and sink for InfluxDB [https://github.com/hazelcast/hazelcast-jet-contrib/tree/master/elasticsearch]
```
Serdaro: I think the links need to be fixed for the above two entries.

hazelcast-jet-distribution/src/main/resources/release_notes.txt
```diff
@@ -5,6 +5,47 @@ issues at https://github.com/hazelcast/hazelcast-jet/issues/<issue number>.
...
+[kafka] Kafka client updated to version 2.2
+
+3. Fixes
+
+[jet-cli] Improve parsing of command line arguments for jet.sh [#1409, #1411, #1449]
+[core] Improve handling of errors if operation packets are lost [#1377, #1435]
+[core] Configured classloader in Hazelcast Config should be set as the parent of job classloader [#1395]
+[core] Job classloader will now print a warning if it's used after a job is completed [#1315, #1446]
+[core] Snapshot creation should be handled gracefully even if execution is already completed on a member [#1397]
+[core] Added check to making sure any explicit Hazelcast IMDG dependency has same version as the shaded one [#1412]
```
Serdaro: ... to ensure that any explicit ...

hazelcast-jet-distribution/src/main/resources/release_notes.txt
cangencer: how do you mean, fixed?

Serdaro: for Elasticsearch the link reads https://github.com/hazelcast/hazelcast-jet-contrib/tree/master/influxdb

and for the Influxdb it reads https://github.com/hazelcast/hazelcast-jet-contrib/tree/master/elasticsearch

they should be replaced with each other?

cangencer: Thanks, I didn't notice! fixed

PR #1452, opened 2019-06-18
[BACKPORT] Add default yaml configurations to the distribution

gurbuzali:
back-port of https://github.com/hazelcast/hazelcast-jet/pull/1451

### DIFF
--- a/hazelcast-jet-distribution/src/assembly/zip.xml
+++ b/hazelcast-jet-distribution/src/assembly/zip.xml
@@ -92,17 +92,17 @@
         <file>
             <source>${config.directory}/hazelcast-jet-default.yaml</source>
             <destName>hazelcast-jet.yaml</destName>
-            <outputDirectory>config</outputDirectory>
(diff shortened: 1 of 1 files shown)


hazelcast-jet-distribution/src/assembly/zip.xml
```diff
@@ -53,6 +54,23 @@
...
+        <!-- Default YAML config files -->
+        <file>
+            <source>${config.directory}/hazelcast-jet-default.yaml</source>
+            <destName>hazelcast-jet.yaml</destName>
+            <outputDirectory>config</outputDirectory>
+        </file>
+        <file>
+            <source>${config.directory}/hazelcast-jet-member-default.yaml</source>
+            <destName>hazelcast.yaml</destName>
+            <outputDirectory>config</outputDirectory>
```
cangencer: better to put them under config/examples, otherwise it's not clear which config files are loaded.

Issue #1454, opened 2019-06-24
com.hazelcast.jet.pipeline.SourceBuilderTest.test_faultTolerance_restartTwice[1: mode=client]


olukas:
_master_  (commit 69cd63dcf7f1f954533954422a526eed1145a332)

Failed on Sonar build: http://jenkins.hazelcast.com/view/Jet/job/jet-oss-master-sonar-nightly/157/testReport/com.hazelcast.jet.pipeline/SourceBuilderTest/test_faultTolerance_restartTwice_1__mode_client_/

Stacktrace:
```
java.lang.AssertionError: jobId=39f2-40a6-bdc2-2a16 expected:<RUNNING> but was:<COMPLETING>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at com.hazelcast.jet.core.JetTestSupport.lambda$assertJobStatusEventually$2(JetTestSupport.java:153)
	at com.hazelcast.jet.core.JetTestSupport$1.run(JetTestSupport.java:196)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1310)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1327)
	at com.hazelcast.jet.core.JetTestSupport.assertTrueEventually(JetTestSupport.java:161)
	at com.hazelcast.jet.core.JetTestSupport.assertJobStatusEventually(JetTestSupport.java:152)
	at com.hazelcast.jet.core.JetTestSupport.assertJobStatusEventually(JetTestSupport.java:147)
	at com.hazelcast.jet.pipeline.SourceBuilderTest.test_faultTolerance_restartTwice(SourceBuilderTest.java:416)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
[...]

PR #1456, opened 2019-06-24
Drop ParallelTest, clean up cluster after tests

viliam-durina:
Remove ParallelTest: the logs were mingled and we were not able to
investigate issues. Also ensure that the shared cluster is cleaned up of
all non-terminated jobs and non-destroyed distributed objects.

Related to #1454

### DIFF
--- a/hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/JetTestSupport.java
+++ b/hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/JetTestSupport.java
@@ -47,7 +47,6 @@ import java.io.PrintWriter;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Map;
-import java.util.concurrent.CancellationException;
 import java.util.concurrent.Future;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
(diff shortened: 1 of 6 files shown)


hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/JetTestSupport.java
```diff
@@ -247,4 +248,36 @@ public void waitForNextSnapshot(JobRepository jr, long jobId, int timeoutSeconds
...
+    /**
+     * Give this method a job and it will ensure it's no longer running. It
+     * will ignore if it's not running. If the cancellation fails, it will
+     * retry.
+     */
+    public void ditchJob(Job job) {
+        for (;;) {
+            Exception failure;
+            try {
+                job.cancel();
```
cangencer: is there a point to trying to cancel several times? if it gets stuck, we won't be able to get out of it.

viliam-durina: For example `cancel` fails when the job is restarting or terminating at the very moment. Exception is thrown from `cancel`. After the job starts, you'll be able to cancel. If the job is stuck, `join` won't return, which is what we want: the job that caused the stuck job will fail.

viliam-durina: Just realized that we'll sleep even in case of a normal job failure. I'll fix that tomorrow...

cangencer: Still if the job gets stuck on cancellation, will it timeout eventually?

viliam-durina: The entire test will time out, it's 5 mins a think. The cluster is probably in a bad shape anyway and subsequent tests on the same cluster might be affected. But if the test times out after 5 mins, more tests will be run. I probably should shut down the cluster...

PR #1457, opened 2019-06-24
Add test sources and sinks

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/SinkImpl.java
```diff
@@ -25,18 +25,25 @@
...
+        this.isSingleton = isSingleton;
     }
 
     @Nonnull
     public ProcessorMetaSupplier metaSupplier() {
         return metaSupplier;
     }
 
+    @Nonnull
+    public boolean isSingleton() {
```
viliam-durina: What about using "total parallelism one" instead of "singleton"?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/test/AssertionP.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public boolean tryProcess() {
+        maybeFireTimer();
+        return true;
+    }
+
+    private void maybeFireTimer() {
+        long now = System.currentTimeMillis();
```
viliam-durina: Better to use `nanoTime()`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/test/AssertionSinkBuilder.java
```diff
@@ -0,0 +1,167 @@
...
+import com.hazelcast.util.Preconditions;
+
+import javax.annotation.Nonnull;
+
+import static com.hazelcast.jet.impl.util.Util.checkSerializable;
+
+/**
+ * See {@link AssertionSinkBuilder#assertionSink(String, SupplierEx)} .
+ *
+ * @param <A> type of the state object
```
viliam-durina: Why don't we call the state `S`?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/test/AssertionP.java
```diff
@@ -0,0 +1,98 @@
...
+        long now = System.nanoTime();
+        if (nextTimerSchedule == 0 || now >= nextTimerSchedule) {
+            timerFn.accept(state);
+            nextTimerSchedule = now + TIMER_INTERVAL;
+        }
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    protected boolean tryProcess(int ordinal, @Nonnull Object item) {
```
viliam-durina: The processor should forward the items, also the WMs. Otherwise, if used in-line, the downstream stages won't receive anything.

cangencer: The processor is never used inline - it's always used as a sink

PR #1459, opened 2019-06-26
Metrics for Completed Jobs

jbartok:
Some initial changes, just the part where a metrics map gets communicated to members/clients. Works for both members and clients, but I needed to do a hack in `JetGetJobMetricsMessageTask.encodeResponse` which probably means I'm doing something wrong somewhere else. 

The server side is completely missing, that's where I turn next, but pls. comment about what is already not right or not acceptable for some reason.


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClientJobProxy.java
```diff
@@ -54,15 +56,24 @@
...
     }
 
-    ClientJobProxy(JetClientInstanceImpl client, long jobId, DAG dag, JobConfig config) {
+    ClientJobProxy(JetClientInstanceImpl client,
+                   SerializationService serializationService,
+                   long jobId,
+                   DAG dag,
+                   JobConfig config) {
         super(client, jobId, dag, config);
+        this.serializationService = serializationService;
```
cangencer: you can use `serializationService()`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
```diff
@@ -82,6 +83,12 @@ default String getName() {
     @Nonnull
     JobStatus getStatus();
 
+    /**
+     * Returns a copy of the current values of all job-specific metrics.
+     */
+    @Nonnull
+    Map<String, Long> getMetrics();
```
cangencer: it might be better to return some other class than Map, in case we wanted to add more information, such as timestamp.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClientJobProxy.java
jbartok: Didn't notice it, thanks!

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
jbartok: Done.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJobProxy.java
```diff
@@ -306,4 +311,37 @@ private void rejoinJob(Throwable t) {
...
+        @Nonnull
+        @Override
+        public Collection<String> getMetricNames() {
+            return metrics.keySet();
+        }
+
+        @Override
+        public Long getMetricValue(String name) throws IllegalArgumentException {
+            if (name == null) {
+                throw new IllegalArgumentException("Metric name can't be null!");
```
viliam-durina: Don't use exclamation marks for exceptions!! :)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClientJobProxy.java
```diff
@@ -202,7 +218,7 @@ public Void get() throws InterruptedException, ExecutionException {
 
         @Override
         public Void get(long timeout, @Nonnull TimeUnit unit)
-                throws InterruptedException, ExecutionException, TimeoutException {
+            throws InterruptedException, ExecutionException, TimeoutException {
```
viliam-durina: this looks weird

jbartok: Wrong setting for "continous indent" in IntelliJ, fixed it.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJobProxy.java
jbartok: Ok, fixed.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,39 @@
...
+     * Returns a collection containing the names of all job specific metrics available.
+     */
+    @Nonnull
+    Collection<String> getMetricNames();
+
+    /**
+     * Retruns the value of a job specific metric with the given name.
+     * @throws IllegalArgumentException if name is null or if no job metric with this name is available.
+     */
+    Long getMetricValue(String name) throws IllegalArgumentException;
```
cangencer: why not `long` ?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,39 @@
...
+
+package com.hazelcast.jet.core;
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+
+/**
+ * Contains information related to job-specific metrics.
+ */
+public interface JobMetrics {
```
cangencer: let's mark this as `@Beta` as well

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJobProxy.java
```diff
@@ -306,4 +311,37 @@ private void rejoinJob(Throwable t) {
...
+            }
+
+            Long value = metrics.get(name);
+            if (value == null) {
+                throw new IllegalArgumentException("Metric with name '" + name + "' not available for job!");
```
cangencer: it's better to return null here than throw

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJobProxy.java
```diff
@@ -306,4 +311,37 @@ private void rejoinJob(Throwable t) {
...
+
+        @Nonnull
+        @Override
+        public Collection<String> getMetricNames() {
+            return metrics.keySet();
+        }
+
+        @Override
+        public Long getMetricValue(String name) throws IllegalArgumentException {
+            if (name == null) {
```
cangencer: use `Objects.requireNonNull()`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClientJobProxy.java
```diff
@@ -62,7 +64,8 @@
         super(client, jobId, dag, config);
     }
 
-    @Nonnull @Override
+    @Nonnull
+    @Override
```
cangencer: can be on sameline

cangencer:
We also have various tags for metrics, so you might consider how to query by tag for example.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJobProxy.java
jbartok: Ok, but then I can't have a primitive "long" return type (just saying I can't comply to both requests).

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
jbartok: Ok, will do.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
jbartok: Will leave it as `Long` so that I can return null too.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJobProxy.java
jbartok: Ok.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClientJobProxy.java
jbartok: Ok.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,44 @@
...
+import com.hazelcast.spi.annotation.Beta;
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+
+/**
+ * Contains information related to job-specific metrics.
+ */
+@Beta
+public interface JobMetrics {
```
cangencer: do we need an interface here? it would just be a DTO.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobProxy.java
```diff
@@ -63,6 +67,15 @@ public JobStatus getStatus() {
         }
     }
 
+    @Nonnull @Override
+    public JobMetrics getMetrics() {
+        try {
+            return JobMetricsImpl.of(this.<Map<String, Long>>invokeOp(new GetJobMetricsOperation(getId())).get());
```
cangencer: JobMetrics instance should be returned inside the operation directly

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ExecutionContext.java
```diff
@@ -174,9 +181,20 @@ public void completeExecution(Throwable error) {
...
-        processors.forEach(metricsRegistry::deregister);
-        tasklets.forEach(metricsRegistry::deregister);
+        deregisterMetricsSources(
+                ((NodeEngineImpl) nodeEngine).getMetricsRegistry(),
+                jobMetricsRegistry
+        );
+    }
+
+    private void deregisterMetricsSources(MetricsRegistry... metricsRegistries) {
+        Arrays.stream(metricsRegistries).forEach(
```
cangencer: you can do a simple for loop here

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
```diff
@@ -134,7 +137,13 @@
         this.memberCount = memberCount;
     }
 
-    public void initialize(NodeEngine nodeEngine, long jobId, long executionId, SnapshotContext snapshotContext) {
+    public void initialize(
+            NodeEngine nodeEngine,
```
cangencer: we don't typically use this type of identation.. see constructor of `MasterContext` for example

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
```diff
@@ -625,4 +635,64 @@ public int getHigherPriorityVertexCount() {
...
+        public ProbeBuilder withTag(String tag, String value) {
+            for (int i = 0; i < builders.length; i++) {
+                builders[i] = builders[i].withTag(tag, value);
+            }
+            return this;
+        }
+
+        @Override
+        public <S> void register(
+                @Nonnull S source,
```
cangencer: see my comments about indentation above

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/operation/ReportJobMetricsOperation.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Override
+    public void run() {
+        JetService service = getService();
+        ExecutionContext executionContext = service.getJobExecutionService().getExecutionContext(executionId);
+        if (executionContext == null) {
+            response = Collections.emptyMap();
+        } else {
+            ToMapProbeRenderer renderer = new ToMapProbeRenderer();
+            executionContext.renderJobMetrics(renderer);
```
cangencer: metrics are already collected every 5 secs by default. do we have want to re-render them on demand ?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
jbartok: ok

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
jbartok: ok

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ExecutionContext.java
jbartok: ok

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobProxy.java
jbartok: Ok, next version will do so.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
jbartok: ok, will change

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/operation/ReportJobMetricsOperation.java
jbartok: I will look into this once I have tests in place.

jbartok: Should be doable to use the periodic rendering we already have, I'll attempt doing it.

jbartok:
Here's a rewrite which uses the periodically rendered metrics + a single triggered rendering on job completion. Let me know which version you would like to keep.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
```diff
@@ -77,6 +79,7 @@ public String toString() {
     private volatile JobStatus jobStatus = NOT_RUNNING;
     private volatile long executionId;
     private volatile Map<MemberInfo, ExecutionPlan> executionPlanMap;
+    private volatile JobMetrics jobMetrics;
```
viliam-durina: I think it's possible to remove this field. The value is never reused:
- if metrics is requested for a running job, `ReportJobMetricsOp` is invoked and when all responses are received they are returned to the client
- whe `CompleteExecutionOp` returned, we can also pass the final metrics to `JobCoordinatorService.completeJob()` and not unnecessarily store it to this field.

The lifecycle of the metrics will be simpler this way.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ExecutionContext.java
```diff
@@ -83,6 +85,8 @@
     private SnapshotContext snapshotContext;
     private JobConfig jobConfig;
 
+    private final Map<String, Long> jobMetrics = new HashMap<>();
```
viliam-durina: The maps can be concurrently read while we're updating it. Using `ConcurrentHashMap` isn't feasible either because it can be read just after it was cleared. We need to use volatile field to which we'll assign a built map.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/JetMetricsService.java
```diff
@@ -259,4 +267,40 @@ private void logError(String name, Object value, MetricsPublisher publisher, Exc
...
+
+        private final Map<String, Long> metrics = new HashMap<>();
+
+        public InternalJobMetricsPublisher(JobExecutionService jobExecutionService) {
+            this.jobExecutionService = jobExecutionService;
+        }
+
+        @Override
+        public void publishLong(String name, long value) {
+            metrics.put(name, value);
```
viliam-durina: We can directly collect to `Map<ExecId, Map<MetricName, Long>>` and in `whenComplete` just use that maps.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Contains information related to job-specific metrics.
+ */
+@Beta
+public final class JobMetrics implements Serializable {
+
+    /** Static empty instance, contains no metrics */
+    public static final JobMetrics EMPTY = new JobMetrics();
```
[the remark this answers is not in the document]
jbartok: Ok, moved to `JobMetricsUtil`, external usages replaced by `JobMetricsUtil.empy()`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,125 @@
...
+     * Retruns the value of a job specific metric with the given name.
+     * @throws IllegalArgumentException if name is null or if no job metric with this name is available.
+     */
+    Long getMetricValue(String name) {
+        Objects.requireNonNull(name);
+        return metrics.get(name);
+    }
+
+    /** Merges two immutable {@link JobMetrics} instances into a third */
+    public JobMetrics merge(JobMetrics other) {
```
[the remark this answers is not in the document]
jbartok: Ok, did that, but in order to be able to remain efficient I had to add a new `asMap()` method.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+/**
+ * Contains information related to job-specific metrics.
+ */
+@Beta
+public final class JobMetrics implements Serializable {
```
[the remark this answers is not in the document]
jbartok: We now have an `asMap()` method, presenting a `Map` view (backed by an unmodifiable map) which can be iterated. Isn't that enough? Do you think I should still implement one of the above suggestions? (I got rid of `merge` btw.)

jbartok:
Well... ok, but nothing really works anymore after your touchups... Let me see if I can sort it out...

jbartok:
Ok, it was just a small mistake in `JobMetricsUtil.METRIC_KEY_EXEC_ID_PATTERN`. We should really consider making our tests run in reasonable time so that people don't skip running them...

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
```diff
@@ -82,6 +83,12 @@ default String getName() {
     @Nonnull
     JobStatus getStatus();
 
+    /**
+     * Returns a copy of the current values of all job-specific metrics.
```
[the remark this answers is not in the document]
jbartok: Yes, I've added a test for restarted jobs. The existent tests were already checking the metrics after job completion.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ExecutionContext.java
jbartok: Ok, have done this change. Tried to do it as efficiently as possible, but still, it's a constant source of garbage production. Maybe that's not a problem though.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/JetMetricsService.java
jbartok: Yes, done, see previous.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
jbartok: Let's discuss this and your comment on races when you will be available.

hazelcast-jet-client-protocol/src/main/java/com/hazelcast/client/impl/protocol/template/JetCodecTemplate.java
```diff
@@ -69,4 +69,7 @@
     @Request(id = 15, retryable = true, response = ResponseMessageConst.STRING)
     Object getMemberXmlConfiguration();
 
+    @Request(id = 16, retryable = false, response = ResponseMessageConst.DATA)
+    Object getJobMetrics(long jobId);
```
cangencer: why not retryable?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
```diff
@@ -82,6 +83,12 @@ default String getName() {
     @Nonnull
     JobStatus getStatus();
 
+    /**
+     * Returns a snapshot of the current values of all job-specific metrics.
```
cangencer: javadoc needs to be expanded. What does it return for completed jobs?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import static com.hazelcast.jet.Util.entry;
+
+/**
+ * An immutable snapshot of job-specific metrics.
```
cangencer: javadoc needs expanding

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,122 @@
...
+ */
+@Beta
+public final class JobMetrics implements Serializable {
```
cangencer: we should implement custom serialization (i.e. IdentifiedDataSerializable) for this class

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,122 @@
...
+    private static final JobMetrics EMPTY = new JobMetrics(Collections.emptyMap());
+
+    private final Map<String, Long> metrics;
+
+    private JobMetrics(@Nonnull Map<String, Long> metrics) {
+        this.metrics = Collections.unmodifiableMap(metrics);
+    }
+
+    /** Builds a {@link JobMetrics} object based on a key-value map of metrics data. */
+    public static JobMetrics of(@Nonnull Map<String, Long> metrics) {
```
cangencer: missing `@Nonnull` for return value
javadoc is not in standard format

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,122 @@
...
+            return EMPTY;
+        }
+        return new JobMetrics(new HashMap<>(metrics));
+    }
+
+    /**
+     * Returns an unmodifiable map view of these metrics.
+     */
+    @Nonnull
+    public Map<String, Long> asMap() {
```
cangencer: `toMap()`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,122 @@
...
+        return metrics;
+    }
+
+    /**
+     * Returns a collection containing the names of all job-specific metrics
+     * available.
+     */
+    @Nonnull
+    public Collection<String> getMetricNames() {
```
cangencer: should return `Set` ?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,122 @@
...
+    @Nullable
+    public Long getMetricValue(@Nonnull String name) {
+        return metrics.get(name);
+    }
+
+    /**
+     * Returns a subset of this metrics which have the specified value for the
+     * specified tag.
+     */
+    public JobMetrics withTag(String tag, String value) {
```
cangencer: missing `Nonnull` annotations

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,122 @@
...
+        Entry<String, String> tagValue = entry(tag, value);
+        Map<String, Long> filteredMetrics = metrics.entrySet().stream()
+                .filter(entry -> MetricsUtil.parseMetricName(entry.getKey()).contains(tagValue))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+        return new JobMetrics(filteredMetrics);
+    }
+
+    /** Returns the number of metrics present.*/
```
cangencer: javadoc non-standard format

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobExecutionService.java
```diff
@@ -327,6 +329,24 @@ public void completeExecution(long executionId, Throwable error) {
...
+        Map<ExecutionContext, Map<String, Long>> metricsOfExecutions = Collections.emptyMap();
+
+        for (Entry<String, Long> entry : metrics.entrySet()) {
+            String metricName = entry.getKey();
+            Long executionId = JobMetricsUtil.getExecutionIdFromMetricName(metricName);
+            if (executionId != null) {
+                ExecutionContext executionContext = executionContexts.get(executionId);
+                if (executionContext != null) {
+                    Long metricValue = entry.getValue();
+                    metricsOfExecutions = putIntoMultiMap(metricsOfExecutions, executionContext, metricName, metricValue);
```
cangencer: why do we need the complexity of this method? you can just use `map.compute()` instead.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobMetricsUtil.java
```diff
@@ -0,0 +1,61 @@
...
+            return null;
+        }
+        return Util.idFromString(m.group(1));
+    }
+
+    public static long toLongMetricValue(double value) {
+        return Math.round(value * DOUBLE_TO_LONG);
+    }
+
+    static JobMetrics mergeMetrics(Collection<Object> metrics) {
```
cangencer: why `Object`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/JetInitDataSerializerHook.java
```diff
@@ -188,6 +192,10 @@ public IdentifiedDataSerializable create(int typeId) {
                     return new ClusterMetadata();
                 case GET_CLUSTER_METADATA_OP:
                     return new GetClusterMetadataOperation();
+                case GET_JOB_METRICS_OP:
+                    return new GetJobMetricsOperation();
+                case GET_JOB_METRICS_FROM_MEMBER_OP:
+                    return new GetJobMetricsFromMemberOperation();
```
cangencer: GetLocalJobMetrics would be better name

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/operation/GetJobMetricsFromMemberOperation.java
```diff
@@ -0,0 +1,63 @@
...
+    public GetJobMetricsFromMemberOperation(long jobId, long executionId) {
+        super(jobId);
+        this.executionId = executionId;
+    }
+
+    @Override
+    public void run() {
+        JetService service = getService();
+        ExecutionContext executionContext = service.getJobExecutionService().getExecutionContext(executionId);
+        if (executionContext == null) {
```
cangencer: you can use `response = ? :`

cangencer:
Also PR is missing labels / milestone

cangencer:
You can do this in a separate PR, but we also need some classes with full lists of metric names and tag names.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
jbartok: Ok, done in next push.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobMetricsUtil.java
jbartok: It was a call dealing with `invokeOnParticipant` responses, which can for example be `Throwable` instances too. Will modify it to be closer to the source, makes more sense there.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobExecutionService.java
jbartok: Yeah, I just thought to create a smaller number of full blown `Map` instances (ie. `Collections.singletonMap` instead of `HashMap`), but was probably a bad idea. Will rewrite.

jbartok:
> There are still races. For example, in `JobCoordinatorService.getJobMetrics()` we can observe that `JobResult` isn't there, but then a moment later master context might not be there as well. Or it might be there, but after `finalizeJob` or restarting. We should always return some metrics: either current snapshot, or the ones from JobResult or last collected (in case of restart). For this we can use the `jobMetrics` field in `MasterContext`, which I proposed to remove.
> 
> It's a complex change, I can't describe it here, but we need to address it. Currently we can randomly fail, even though no member failed nor did the job fail.

@viliam-durina I'm making a change to `JobCoordinationService.getJobMetrics`, inspired by how `getJobStatus` works + the `getJobMetrics` request is retryable now, so maybe this will alleviate the problem. But, when you have time we should talk, because I don't really understand the scenarios/problem you are mentioning.

As far as I know this remains the only outstanding TODO in this PR.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobMetrics.java
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * <p>
+     *     Builds a {@link JobMetrics} object based on a key-value map of metrics data.
+     *     The key Strings in the map should be well formed metric names (see class javadoc
+     *     for details).
+     * </p>
+     */
+    @Nonnull
+    public static JobMetrics of(@Nonnull Map<String, Long> metrics) {
```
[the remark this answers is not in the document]
jbartok: Well, I had a public EMPTY constant at some point, but got boo-d out during review... On the other hand `null` parameters are a bad idea in general (as it turned out historically). Plus there's always `Collections.emptyMap()`, which doesn't really cost more than a `null', so I'm leaning towards not changing anything. 

Creation of `JobMetrics` objects is not something a user should be doing anyways.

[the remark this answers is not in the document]
cangencer: It's a readonly class, so why would you want to create an empty one?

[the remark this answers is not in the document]
olukas: Agree. My concern was invalid here - I've just asked.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/JobMetricsTest.java
```diff
@@ -0,0 +1,234 @@
...
+import static com.hazelcast.jet.core.JobStatus.SUSPENDED;
+import static com.hazelcast.jet.core.TestUtil.assertExceptionInCauses;
+import static com.hazelcast.jet.impl.util.ExceptionUtil.peel;
+import static java.util.Collections.singletonList;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+@RunWith(HazelcastSerialClassRunner.class)
+public class JobMetricsTest extends TestInClusterSupport {
```
viliam-durina: We should add a stress test: one thread will keep restarting the job, another thread will query the metrics. It must never fail. 

Ideally we should also check that the metrics aren't stale. This might be the way:
- use a processor that will increment a counter in a static field
- that field will be exposed using a `@Probe` on the processor class, this works now
- the test will check that it sees monotonically increasing value for this metric. It must never miss this metric once it was present in the result.
Second test might test the same, but instead of restarting it can do `suspend+sleep(100ms)+resume+sleep(100ms)`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterContext.java
viliam-durina: This was wrong. We need the field while the job is restarting or suspended.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/JobMetricsTest.java
olukas: Agree that these stress tests would be useful. I will prepare them.

olukas:
@viliam-durina it seems that failed/stucked tests from PR builder are related to your latest changes. At least failures in `SourceBuilder_TopologyChangeTest` can be reproduced locally. Could you please fix it?

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/JobMetrics_StreamTest.java
```diff
@@ -0,0 +1,287 @@
...
+        job.cancel();
+        assertJobStatusEventually(job, FAILED);
+        // Then
+        assertTrueEventually(() -> assertMetrics(job.getMetrics(), 5, 2));
+    }
+
+    @Test
+    @Ignore("https://hazelcast.atlassian.net/wiki/spaces/JET/pages/1758560997/"
+            + "Metrics+for+Completed+Jobs+Design?focusedCommentId=1767506913#comment-1767506913")
+    public void metricsExistWhenJobSuspendedAndResumed() {
```
viliam-durina: I'm for removing this test. We should not try to preserve metrics from previous execution. Metrics are local for a member and members can come and go, how would we repartition the metrics?

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/JobMetrics_StreamTest.java
```diff
@@ -0,0 +1,287 @@
...
+
+        job.cancel();
+        assertJobStatusEventually(job, FAILED);
+        assertMetrics(job.getMetrics(), 2, 1);
+    }
+
+    @Test
+    @Ignore("https://hazelcast.atlassian.net/wiki/spaces/JET/pages/1758560997/"
+            + "Metrics+for+Completed+Jobs+Design?focusedCommentId=1767866594#comment-1767866594")
+    public void notResetMetricsForExactlyOnceProcessingGuarantee() {
```
viliam-durina: Remove this too, same reason.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/JobMetrics_StreamTest.java
olukas: I've update this test to be aligned with current behaviour.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/JobMetrics_StreamTest.java
olukas: Removed

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/JobMetrics.java
```diff
@@ -0,0 +1,233 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static java.util.function.Function.identity;
+import static java.util.stream.Collectors.toSet;
+
+/**
+ * An immutable collection of job-specific metrics, pairs of metric names and sets of associated {@link Measurement}s.
```
cangencer: All javadoc needs to be limited to 72 net characters (excluding any prefix/indent at the beginning. See for example it's done correctly in `MetricTags`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/JobMetrics.java
```diff
@@ -0,0 +1,233 @@
...
+
+    private static final JobMetrics EMPTY = new JobMetrics(Collections.emptyMap());
+
+    private static final Collector<Measurement, ?, TreeMap<String, Set<Measurement>>> COLLECTOR = Collectors.groupingBy(
+            measurement -> measurement.getTag(MetricTags.METRIC),
+            TreeMap::new,
+            Collectors.mapping(identity(), toSet())
+    );
+
+    private Map<String, Set<Measurement>> metrics; //metric name -> set of measurements
```
cangencer: is there a reason to use Set here instead of List? We don't expect duplicate of same measurement anywhere?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MeasurementFilters.java
```diff
@@ -0,0 +1,110 @@
...
+
+package com.hazelcast.jet.core.metrics;
+
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+
+/**
+ * Static utility class for creating various {@link Measurement} filtering predicates.
+ */
+public final class MeasurementFilters {
```
cangencer: MeasurementPredicates

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MeasurementFilters.java
```diff
@@ -0,0 +1,110 @@
...
+    private MeasurementFilters() { }
+
+    /**
+     * Matches a {@link Measurement} which contain the specified tag.
+     *
+     * @param tag the tag of interest
+     * @return a filtering predicate
+     */
+    public static Predicate<Measurement> containsTag(String tag) {
+        return new IsPresent(tag);
```
cangencer: you can just inline all these lambdas, no need for creating a class for each

hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/log-debug.adoc
```diff
@@ -218,5 +219,39 @@ available and are observable as MBeans under the node
...
+
+While the job is running the metric values are updated periodically (see
+{jet-javadoc}/config/MetricsConfig.html#setCollectionIntervalSeconds-int-[MetricsConfig.setCollectionIntervalSeconds]).
+
+Once a job stops executing (successfully, after a failure, cancellation,
+or temporarily while suspended) the metrics will have their most
+recent values (i.e. the last metric values from the moment before the
+job completed).
+
+When a job is restarted then the metrics are reset too, their values
```
olukas: We should mention here that metrics are also reset after calling `resume` (after previous `suspend`).

jbartok:
verify

PR #1460, opened 2019-06-27
Add workaround for StreamSocketPTest

cangencer:
The test fails when run twice with maven due to newline in parameters. See:
https://issues.apache.org/jira/browse/SUREFIRE-1662


hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/StreamSocketPTest.java
```diff
@@ -82,12 +83,13 @@ public void before() {
...
         context = new TestProcessorContext();
         bucket = outbox.queue(0);
+        input = input.replaceAll("%n", "\n").replaceAll("%r", "\r");
     }
 
     @Test
     public void smokeTest() throws Exception {
         // we'll test the input as if it is split at every possible position. This is to test the logic that input can be
-        // split at any place: between \r\n, between the bytes of utf-16 sequence etc
+        // split at any place: between %r%n, between the bytes of utf-16 sequence etc
```
viliam-durina: Should not have been replaced here

Issue #1472, opened 2019-07-05
Cannot mix and match YAML and XML on Jet and IMDG configs


eminn:
When Jet uses default (XML) configuration and provided with YAML configuration for IMDG, it tries to parse the YAML IMDG configuration as XML. There shouldn't be restriction to have them both in the same type.

```
objc[41663]: Class JavaLaunchHelper is implemented in both /Library/Java/JavaVirtualMachines/jdk1.8.0_91.jdk/Contents/Home/bin/java (0x1031b64c0) and /Library/Java/JavaVirtualMachines/jdk1.8.0_91.jdk/Contents/Home/jre/lib/libinstrument.dylib (0x10327e4e0). One of the two will be used. Which one is undefined.
Jul 05, 2019 6:58:39 PM com.hazelcast.config.AbstractConfigLocator
INFO: Loading 'hazelcast-jet-default.xml' from the classpath.
Jul 05, 2019 6:58:39 PM com.hazelcast.jet.config.JetConfig
INFO: jet.home is /Users/emindemirci/Development/testbed
Jul 05, 2019 6:58:40 PM com.hazelcast.config.AbstractConfigLocator
INFO: Loading configuration '/Users/emindemirci/Development/testbed/src/main/resources/hazelcast.yaml' from System property 'hazelcast.config'
Jul 05, 2019 6:58:40 PM com.hazelcast.config.AbstractConfigLocator
INFO: Using configuration file at /Users/emindemirci/Development/testbed/src/main/resources/hazelcast.yaml
[Fatal Error] :1:1: Content is not allowed in prolog.
Jul 05, 2019 6:58:40 PM com.hazelcast.config.XmlConfigBuilder
SEVERE: Failed to parse the inputstream
Exception: Content is not allowed in prolog.
Hazelcast startup interrupted.
Exception in thread "main" com.hazelcast.config.InvalidConfigurationException: Content is not allowed in prolog.
	at com.hazelcast.config.XmlConfigBuilder.parse(XmlConfigBuilder.java:211)
	at com.hazelcast.config.XmlConfigBuilder.parseAndBuildConfig(XmlConfigBuilder.java:154)
	at com.hazelcast.config.XmlConfigBuilder.build(XmlConfigBuilder.java:146)
	at com.hazelcast.config.XmlConfigBuilder.build(XmlConfigBuilder.java:139)
[...]

PR #1474, opened 2019-07-05
Introduced ability to mix and match YAML and XML configs for Jet and …

eminn:
…IMDG

Fixes https://github.com/hazelcast/hazelcast-jet/issues/1472

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/config/ConfigProvider.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/config/ConfigProvider.java
@@ -19,6 +19,9 @@ package com.hazelcast.jet.impl.config;
 import com.hazelcast.client.config.ClientConfig;
 import com.hazelcast.client.config.XmlClientConfigBuilder;
 import com.hazelcast.client.config.YamlClientConfigBuilder;
+import com.hazelcast.config.Config;
+import com.hazelcast.config.XmlConfigBuilder;
+import com.hazelcast.config.YamlConfigBuilder;
 import com.hazelcast.jet.config.JetConfig;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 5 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/config/ConfigProvider.java
```diff
@@ -102,4 +105,35 @@ public static ClientConfig locateAndGetClientConfig() {
...
+
+    @Nonnull
+    public static Config locateAndGetMemberConfig(@Nullable Properties properties) {
+        Config config;
+        XmlJetMemberConfigLocator xmlConfigLocator = new XmlJetMemberConfigLocator();
+        YamlJetMemberConfigLocator yamlConfigLocator = new YamlJetMemberConfigLocator();
+
+        if (yamlConfigLocator.locateFromSystemProperty()) {
+            // 1. Try loading config if provided in system property and it is an YAML file
+            config = new YamlConfigBuilder(yamlConfigLocator.getIn()).setProperties(properties).build();
```
gurbuzali: does yaml have priority over xml ?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/config/YamlJetMemberConfigLocator.java
```diff
@@ -30,7 +30,7 @@
     private static final String HAZELCAST_ENTERPRISE_MEMBER_DEFAULT_YAML = "hazelcast-jet-enterprise-member-default.yaml";
 
     public YamlJetMemberConfigLocator() {
-        super(true);
+        super(false);
```
gurbuzali: I guess this means `dont fail if the suffix is not expected`, why ?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/config/ConfigProvider.java
eminn: This is related to https://github.com/hazelcast/hazelcast-jet/issues/1380, if it doesn't have yaml extensions we are treating it as XML one. So we are checking if the file has yaml extensions first.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/config/YamlJetMemberConfigLocator.java
eminn: It shouldn't fail to be able to fallback to other methods like loading from classpath etc.

PR #1475, opened 2019-07-05
Introduced YAML configuration for CLI client

### LINKED ISSUE #1473: Jet CLI accept only XML client configuration
It should also accept YAML configuration

eminn:
Fixes #1473

### DIFF
--- a/hazelcast-jet-all/src/test/java/com/hazelcast/jet/server/JetCommandLineTest.java
+++ b/hazelcast-jet-all/src/test/java/com/hazelcast/jet/server/JetCommandLineTest.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.jet.server;
 
+import com.hazelcast.client.config.ClientConfig;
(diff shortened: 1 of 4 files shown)


cangencer:
is it possible to add some tests for this?

eminn:
Let me try to add one

hazelcast-jet-all/src/test/java/com/hazelcast/jet/server/JetCommandLineTest.java
```diff
@@ -92,10 +100,13 @@ public static void afterClass() {
     public void before() {
         JetConfig cfg = new JetConfig();
         cfg.getHazelcastConfig().addEventJournalConfig(new EventJournalConfig().setMapName(SOURCE_NAME));
+        cfg.getHazelcastConfig().getNetworkConfig().setPort(9000);
```
cangencer: will this cause a problem with parallel tests? maybe better to change group name?

PR #1479, opened 2019-07-16
Jet User Metrics

jbartok:
Trying to implement user metrics for Jet the way that @cangencer suggested. By using a marker interface to mark user code and then somehow ending up registering such classes cleanly with the MetricsRegistry on each cluster member. 

It was clear from the start that this might get messy, this PR is meant to show concretely how and what gets messy and to provide a means for further discussion which hopefully will produce an acceptable solution.

Current solution covers only MapTransform, but the core of it is there and it's easily extendable to other transforms.

So, pls. take a look and comment.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/MetricsSupplier.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/MetricsSupplier.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright (c) 2008-2019, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.jet.core;
+
+/**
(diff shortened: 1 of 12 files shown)


cangencer:
I don't understand how this is meant to be used? What I had mind was something completely different. For example for a map operation, if the supplied FunctionEx implements some interface it would also register as metrics. Nothing to do with transforms themselves.

jbartok:
I meant it to be used like that. I've experimented with filter, not map, but that's the same. So I have some pipeline code like this: `p.drawFrom(...).filter(new Filter())`, where `Filter` implements `PredicateEx` and the `MetricsSupplier` marker. **Because it implements the marker** it will get some special handling as it goes through `StreamStageImp.filter ... attachFilter ... attach` with the aim of getting an instance of `Filter` into `ExecutionPlan` so that it can be registered with the `MetricsRegistry` of each cluster member. 

The `@Probe` annotation from `Filter` come into play only at this step, at the registration into `MetricsRegistry`. 

The "vehicle" I've chosen to transport the instances is `Vertex`, in particular the `ProcessorMetaSupplier` from it, by using a wrapper with extra information, `ProcessorWithMetricsMetaSupplier`. 

I fully realise that this is not ok as is, but maybe it helps to break the problem into more manageable pieces. So: 
* first, is it ok to try to carry instances of classes marked by such a marker interface over into members for registration?
* second, if the previous answer is yes, what other, more civilised way could we use for transport?

cangencer:
It could be done at Processor level? All processors have access to the user passed functions.

jbartok:
Screenshot 2019-07-29 at 09 09 52

You are correct, the processors do have access, but that access it not really practical to use. Even in the simplest example I can think of, one that has a filter marked for metrics provision, that filter is hidden very deeply in the function reference of its processor. See attached screenshot.

I don't really see how I could find the original user functions in all kinds of processors without reflection based searching... I could do that, but seems risky.

Hence my previous approach of trying to isolate the original user functions before the DAG is sent to the cluster and transmitting them separately. 

Let's discuss it a bit when you have time.

PR #1495, opened 2019-07-22
Cap number of job results in cluster to 10,000

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -353,6 +355,14 @@ void cleanup(NodeEngine nodeEngine) {
                 }
             }
         }
+
+        int maxNoResults = nodeEngine.getProperties().getInteger(JetProperties.JOB_RESULTS_MAX_SIZE);
```
viliam-durina: We can add `Math.max(1, ___)` here to avoid the case when the users uses a value <=0. Otherwise we'll clean up every time (with 0) or will fail (if negative value).

PR #1500, opened 2019-07-25
Light job, faster job initialization

viliam-durina:


viliam-durina:
We're not doing this for now

PR #1507, opened 2019-07-29
Make idling min and max both configurable

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Processor.java
```diff
@@ -61,7 +60,7 @@
      * operations. Violations of this rule will manifest as less than 100% CPU
      * usage under maximum load (note that this is possible for other reasons too,
      * for example if the network is the bottleneck or if {@linkplain
-     * JetProperties#JET_MINIMUM_IDLE_MICROSECONDS parking time} is too high).
+     * JetProperties#JET_IDLE_COOPERATIVE_MIN_MICROSECONDS parking time} is too high).
```
mtopolnik: Why do we link to the minimum parking time here? Is it more relevant than the max parking time?

cangencer: Originally we only had the minimum as configurable. I'll update this part too.

PR #1508, opened 2019-07-29
Simplify creation of queues when edge is not partitioned

cangencer:
Previously partitions were calculated and assigned for every edge whether it was
partitioned or not and this caused unnecessary overhead during job initialisation.


cangencer:
verify

PR #1509, opened 2019-07-29
Make enabling job metrics optional

cangencer:
Registering metrics for running very small jobs incurs additional overhead without
bringing much value. It should be made optional


cangencer:
Yes I tried to write a test but it wasn't so straightforward, but we should add a test to the other PR.

PR #1511, opened 2019-07-30
Regression test for job submission slowness

olukas:
This test case is not part of default profile (i.e. tests which are executed by PR builder). You can execute it locally by `mvn test -Pnightly -Dtest=JobSubmissionSlownessRegressionTest`.

### DIFF
--- a/hazelcast-jet-core/src/test/java/com/hazelcast/jet/job/JobSubmissionSlownessRegressionTest.java
+++ b/hazelcast-jet-core/src/test/java/com/hazelcast/jet/job/JobSubmissionSlownessRegressionTest.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright (c) 2008-2019, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/test/java/com/hazelcast/jet/job/JobSubmissionSlownessRegressionTest.java
```diff
@@ -0,0 +1,162 @@
...
+        double measurementARate = measurementARateSum / MEASUREMENT_A_CYCLE_COUNT;
+        double measurementBRate = measurementBRateSum / MEASUREMENT_B_CYCLE_COUNT;
+
+        assertTrue("Job submission rate should not decrease. First rate: " + measurementARate
+                + ", second rate: " + measurementBRate, measurementARate * 0.8 < measurementBRate);
+    }
+
+    private static DAG twoVertex() {
+        DAG dag = new DAG();
+        Vertex v1 = dag.newVertex("v", () -> new SleepingP(0));
```
cangencer: we can remove `SleepingP` and replace it with `Processors.noop()`. It was originally used to test when the processor was not completing immediately.

olukas: You're right. I updated it.

PR #1513, opened 2019-07-31
Add adapter from AggregateOperation1 to Aggregator

cangencer:


cangencer:
Breaking change: `AggregateOperation1.toCollector` method has been moved `AggregateOperations` class and the old one deprecated. The reason is that the type inference isn't able to work correctly when it's used like `stream.collect(AggregateOperations...toCollector())`

PR #1514, opened 2019-08-01
Improve map reader

cangencer:
Add new map reader which is able to read multiple partitions 
in parallel and is cooperative.


jbartok:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadMapP.java
```diff
@@ -0,0 +1,885 @@
...
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Private API, see methods in {@link SourceProcessors}.
+ * <p>
+ * The number of Hazelcast partitions should be configured to at least
+ * {@code localParallelism * clusterSize}, otherwise some processors will
+ * have no partitions assigned to them.
+ */
+public final class ReadMapP<R, E> extends AbstractProcessor {
```
cangencer: we shouldn't call it ReadMapP

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadMapP.java
```diff
@@ -0,0 +1,885 @@
...
+                return new ReadMapP<>(
+                        reader,
+                        partitions,
+                        migrationWatcher.createWatcher()
+                );
+            }
+        }
+    }
+
+    protected interface Reader<R, E> extends Serializable {
```
cangencer: why do you need both interface and abstract class here when interface is never implemented directly?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadParallelPartitionsP.java
```diff
@@ -0,0 +1,879 @@
...
+        static final long serialVersionUID = 1L;
+        private Reader<R, E> reader;
+        private transient Map<Address, List<Integer>> addrToPartitions;
+
+        LocalProcessorMetaSupplier(@Nonnull Reader<R, E> reader) {
+            this.reader = reader;
+        }
+
+        @Override
+        public int preferredLocalParallelism() {
```
cangencer: this is leftover, should be deleted

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadParallelPartitionsP.java
```diff
@@ -0,0 +1,879 @@
...
+                throw new IllegalArgumentException("index=" + index + ", count=" + count);
+            }
+            return IntStream.range(0, list.size())
+                    .filter(i -> i % count == index)
+                    .mapToObj(list::get)
+                    .collect(toList());
+        }
+
+        @Override
+        public int preferredLocalParallelism() {
```
cangencer: likewise, should be deleted

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadParallelPartitionsP.java
```diff
@@ -0,0 +1,879 @@
...
+                return new ReadParallelPartitionsP<>(
+                        reader,
+                        partitions,
+                        migrationWatcher.createWatcher()
+                );
+            }
+        }
+    }
+
+    private abstract static class Reader<R, E> implements Serializable {
```
cangencer: I would move readers to another file, to clean up this big file.

jbartok: I have not extracted `Reader` because as far as I can tell all the other processors are in a single file too. We can, just not too sure what its semantics would be, as a stand alone class. 

Reader is passed into `ProcessorMetaSupplier` constructors. Those get serialized and sent to the cluster, no? If yes, then `Reader` needs to be serializable too, right?

I have not done these two changes.

jbartok:
verify

jbartok:
verify

jbartok:
Not sure what's going on with these test failures... For example locally I caught a case now when two separate processes are running tests from `SourcesTest` and the whole thing is stuck. I think this class can't deal with parallel runs because of the remote cluster it creates. Pls. take a look, I don't think I will manage to squeeze it in today.

Screenshot 2019-08-10 at 09 00 15

[dump_16238.txt](https://github.com/hazelcast/hazelcast-jet/files/3488559/dump_16238.txt)
[dump_47211.txt](https://github.com/hazelcast/hazelcast-jet/files/3488560/dump_47211.txt)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadParallelPartitionsP.java
cangencer: You can pass SupplierEx<Reader> which seems more correct. You're just telling it what reader to use, not the actual reader instance.

PR #1516, opened 2019-08-02
{map/filter/flatMap}Stateful

mtopolnik:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/StatefulFlatMapP.java
```diff
@@ -0,0 +1,157 @@
...
+        currentWm = watermark.timestamp();
+        assert currentWm != Long.MIN_VALUE : "Received WM == Long.MIN_VALUE";
+        long countEvicted = 0;
+        for (Iterator<Entry<K, TimestampedItem<S>>> iter = keyToState.entrySet().iterator(); iter.hasNext();) {
+            long timestamp = iter.next().getValue().timestamp();
+            if (timestamp == Long.MIN_VALUE || timestamp >= currentWm - ttl) {
+                break;
+            }
+            iter.remove();
+            if (++countEvicted == MAX_ITEMS_TO_EVICT) {
```
jerrinot: What if you have a high throughput (of unique keys), but low number of watermarks? 
Could it be you have millions of items / s, but just 1 watermark per second?

cangencer:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/StatefulFlatMapP.java
mtopolnik: `LruHashMap` evicts the eldest item each time you put in a new one, and this mechanism acts on top of that to reduce the map's size when transitioning from a period of high activity to low activity.

cangencer:
verify

PR #1533, opened 2019-08-07
Drop CompleteExecutionOperation

viliam-durina:
It was only for cleanup. Now we cleanup immediately after the execution
completes on member or immediately when execution termination is
requested, if execution didn't begin yet.


viliam-durina:
We're not doing this for now.

PR #1538, opened 2019-08-08
Add s3 module

gurbuzali:


hazelcast-jet-s3/src/test/java/com/hazelcast/jet/s3/SinkTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @BeforeClass
+    public static void setup() {
+        accessKeyId = getSystemPropertyOrEnv("S3_ACCESS_KEY_ID");
+        accessKeySecret = getSystemPropertyOrEnv("S3_ACCESS_KEY_SECRET");
+        client = S3Utils.client(accessKeyId, accessKeySecret, US_EAST_1);
+    }
+
+    @Before
+    public void deleteObjects() {
```
eminn: Does this leaves any files on the S3 bucket when a test fails for example ? So I think it is better to add also an `@After` hook to cleanup the bucket.

hazelcast-jet-reference-manual/src/main/asciidoc/source-sink-connectors/s3.adoc
```diff
@@ -0,0 +1,62 @@
+= Amazon AWS S3
+
+Amazon AWS S3 Object Storage is a production-worthy choice for both a
+data source and a data sink in a batch computation job.
+
+The S3 Source and Sink uses Amazon S3 SDK to connect to the storage.
+The connectors expect the user to provide either an
+https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3.java[AmazonS3]
+client or credentials (Access key ID, Secret Access Key) to create the
+client. The source and sink assume the data is in the form of  plain
```
eminn: double space before plain.

cangencer:
verify

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
```diff
@@ -0,0 +1,153 @@
...
+        int itemCounter;
+        long objectCounter;
+        StringBuilder buffer = new StringBuilder();
+
+        S3Context(
+                String bucketName,
+                int processorIndex,
+                int batchSize,
+                SupplierEx<? extends AmazonS3> clientSupplier,
+                FunctionEx<T, String> toStringFn
```
cangencer: ? super T

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * Convenience for {@link #s3(String, int, SupplierEx, FunctionEx)}.
+     * Creates an S3 client with given credentials and region. Uses {@link
+     * Object#toString()} for {@code toStringFn}.
+     */
+    @Nonnull
+    public static <T> Sink<? super T> s3(
+            @Nonnull String bucketName,
+            @Nonnull String accessKeyId,
```
cangencer: I think it would be better to group accessKeyId / secret / region into an object?

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
```diff
@@ -0,0 +1,153 @@
...
+     * followed by an incremented counter, for example the processor having the
+     * index 2 will create these consecutive files {@code 2-0, 2-1, 2-2 ...}.
+     * <p>
+     * No state is saved to snapshot for this sink. If the job is restarted
+     * previously written files will be overwritten.
+     * <p>
+     * The default local parallelism for this sink is 1.
+     *
+     * @param bucketName     the name of the bucket
+     * @param batchSize      the size of the batch
```
cangencer: I think the name of the parameter is confusing. It should be something like number of lines per file?

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sources.java
```diff
@@ -0,0 +1,240 @@
...
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Collections.singletonList;
+import static java.util.stream.Collectors.toMap;
+
+/**
+ * Contains factory methods for creating AWS S3 sources.
+ */
+public final class S3Sources {
+
+    private static final int LOCAL_PARALLELISM = 1;
```
cangencer: we should use 2

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sources.java
```diff
@@ -0,0 +1,240 @@
...
+    /**
+     * Creates an AWS S3 {@link BatchSource} which lists all the objects in the
+     * bucket-list using given {@code prefix}, reads them line by line,
+     * transforms each line to the desired output object using given {@code
+     * mapFn} and emits them to downstream.
+     * <p>
+     * The source does not save any state to snapshot. If the job is restarted,
+     * it will re-emit all entries.
+     * <p>
+     * The default local parallelism for this processor is 1.
```
cangencer: we should use 2

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sources.java
```diff
@@ -0,0 +1,240 @@
...
+                .destroyFn(S3Context::close)
+                .build();
+    }
+
+    private static final class S3Context<T> {
+
+        private static final ObjectListing EMPTY_LISTING = new ObjectListing();
+        private static final int BATCH_COUNT = 10;
+
+        final String prefix;
```
cangencer: why aren't these private?

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sources.java
```diff
@@ -0,0 +1,240 @@
...
+                .<T>fillBufferFn(S3Context::fillBuffer)
+                .distributed(LOCAL_PARALLELISM)
+                .destroyFn(S3Context::close)
+                .build();
+    }
+
+    private static final class S3Context<T> {
+
+        private static final ObjectListing EMPTY_LISTING = new ObjectListing();
+        private static final int BATCH_COUNT = 10;
```
cangencer: we can use larger batch size I think, 1024?

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
```diff
@@ -0,0 +1,153 @@
...
+import com.hazelcast.jet.pipeline.SinkBuilder;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Contains factory methods for creating AWS S3 sinks.
+ */
+public final class S3Sinks {
+
+    private static final int DEFAULT_BATCH_SIZE = 1024;
```
cangencer: I think we need higher default. This will create many very small files.

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sources.java
```diff
@@ -0,0 +1,240 @@
...
+                if (line == null) {
+                    reader.close();
+                    reader = null;
+                    return;
+                }
+                buffer.add(mapFn.apply(objectName, line));
+            }
+        }
+
+        S3Object nextObject() {
```
cangencer: this method should be inlined

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
gurbuzali: like `S3Credentials` or `S3Parameter` ?

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
gurbuzali: so batchSize -> linesPerFile ?

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
gurbuzali: What do you have in mind ?

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
cangencer: Yeah S3Credentials is ok I think. They can also be defined by environment variables? Are these optional then?

cangencer:
Some interesting comments about ResetException here: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/best-practices.html
Do we need to handle these / retry ?

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
cangencer: what about the `AWSCredentials` object thats comes in the API

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
```diff
@@ -0,0 +1,151 @@
...
+     * @param clientSupplier S3 client supplier
+     * @param toStringFn     the function which converts each item to its
+     *                       string representation
+     * @param <T>            type of the items the sink accepts
+     */
+    @Nonnull
+    public static <T> Sink<? super T> s3(
+            @Nonnull String bucketName,
+            int linesPerFile,
+            @Nonnull SupplierEx<? extends AmazonS3> clientSupplier,
```
cangencer: I was looking at the S3 API and there's already a good builder for S3 client in the API. I don't think we need any other overload and this one. We can remove the S3Utils/S3Parameters and with examples in the Javadoc. This way we don't need so many overloads.

examples/s3/src/main/java/com/hazelcast/jet/examples/s3/S3WordCount.java
```diff
@@ -0,0 +1,113 @@
...
+ * <p>
+ * {@link S3Sinks#s3(String, int, SupplierEx, FunctionEx)}
+ * writes the output to the given output bucket, with each
+ * processor writing to a batch of files within the bucket. The files are
+ * identified by the global processor index and an incremented value.
+ */
+public class S3WordCount {
+
+    private static Pipeline buildPipeline(
+            String accessKey, String secretKey, Regions region,
```
cangencer: i think you can make these static variables

gurbuzali:
verify

cangencer:
can we move the s3 example to `files`? instead?

cangencer:
It would be good check for bucket's existence in the sink so it fails fast. I was testing with an invalid bucket and it didn't fail until it actually started writing. Also shouldn't there be an option for the prefix on the sink as well?

hazelcast-jet-s3/src/test/java/com/hazelcast/jet/s3/S3MockTest.java
```diff
@@ -0,0 +1,141 @@
...
+        try {
+            jet.newJob(p).join();
+        } catch (CompletionException e) {
+            assertTrue(e.getCause().getCause() instanceof AssertionCompletedException);
+        }
+    }
+
+    private int generateAndUploadObjects() {
+        Random random = new Random();
+        int objectCount = random.nextInt(10) + 5;
```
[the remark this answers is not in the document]
gurbuzali: looks like the failure is related with S3Mock, it does not list more than 1000 objects.
I've tried with actual S3 and it passed.
I'll remove the randomness

gurbuzali:
prefix for the sink ? like a prefix for the names of the files created ?
For the source it is used for filtering.

gurbuzali:
verify

cangencer:
> prefix for the sink ? like a prefix for the names of the files created ?
> For the source it is used for filtering.

Yes, to used for writing, so you don't write to the root of the bucket and also can name the files appropriately.

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
```diff
@@ -0,0 +1,167 @@
...
+            buffer.append(toStringFn.apply(item))
+                  .append(System.lineSeparator());
+            if (++itemCounter == linesPerFile) {
+                amazonS3.putObject(bucketName, nextKey(), buffer.toString());
+                buffer.setLength(0);
+                itemCounter = 0;
+            }
+        }
+
+        private void close() {
```
cangencer: `close()` is called after job is completed, but here we are writing. I'm not sure if this is a good idea.

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sources.java
```diff
@@ -0,0 +1,222 @@
...
+                    .iterator();
+        }
+
+
+        private boolean belongsToThisProcessor(String key) {
+            return Math.floorMod(key.hashCode(), totalParallelism) == processorIndex;
+        }
+
+        private void close() {
+            amazonS3.shutdown();
```
cangencer: any open readers (if there was some exception, or job is cancelled) should also be closed here

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
```diff
@@ -161,7 +167,7 @@ private void close() {
 
 
         private String nextKey() {
-            return processorIndex + "-" + (++objectCounter);
+            return prefix + processorIndex + "-" + (++objectCounter);
```
gurbuzali: how about adding `/` between prefix and processorIndex
this way s3 will create a folder inside the bucket group all objects under that folder

eminn: Then may be we need to call it folder rather than a prefix if we take that approach ?

gurbuzali: yes, a `folder` or `directory` maybe

eminn: @cangencer what is your take on this ?

gurbuzali: I guess S3 web UI using `folder`

cangencer: hmm I think s3 terminology is prefix, so I'd prefer that, and it's used in source as well. Also you can put like `folder/file-` and then the files be called `folder/file-0-1` etc

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
```diff
@@ -0,0 +1,240 @@
...
+     *                       the strings
+     * @param <T>            type of the items the sink accepts
+     */
+    @Nonnull
+    public static <T> Sink<? super T> s3(
+            @Nonnull String bucketName,
+            @Nullable String prefix,
+            @Nonnull SupplierEx<? extends AmazonS3> clientSupplier,
+            @Nonnull FunctionEx<? super T, String> toStringFn,
+            @Nonnull String charset
```
cangencer: lambda parameters should be last

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        private void checkIfBucketExists() {
+            if (!s3Client.doesBucketExistV2(bucketName)) {
+                throw new IllegalArgumentException("Bucket [" + bucketName + "] does not exist");
+            }
+        }
+
+        private void receive(T item) {
+            buffer.append(toStringFn.apply(item))
```
cangencer: wouldn't it be better to create a byte buffer and write it to that directly? now we first write to string buffer, and then copy the whole thing again into a byte buffer

hazelcast-jet-s3/src/main/java/com/hazelcast/jet/s3/S3Sinks.java
```diff
@@ -0,0 +1,240 @@
...
+                completeActiveRequest();
+                fileCounter++;
+                partCounter = 0;
+                initUploadRequest();
+            }
+            if (buffer.length() <= DEFAULT_MINIMUM_UPLOAD_PART_SIZE) {
+                return;
+            }
+            UploadPartRequest uploadRequest = createUploadRequestFromBuffer();
+            UploadPartResult uploadResult = s3Client.uploadPart(uploadRequest);
```
cangencer: also see this: 

```* When supplying an {@link InputStream} using {@link
     * UploadPartRequest#withInputStream(InputStream)} or {@link
     * UploadPartRequest#setInputStream(InputStream)}, the stream will only be
     * closed by the client if {@link UploadPartRequest#isLastPart()} is {@code
     * true}. If this is not the last part, the stream will be left open.```

eminn: I've seen it when I was exploring the parameters. Thanks for the remark.

PR #1546, opened 2019-08-16
Add code samples

cangencer:


cangencer:
verify

examples/co-group/src/main/java/com/hazelcast/jet/examples/cogroup/BatchCoGroup.java
```diff
@@ -0,0 +1,221 @@
...
+            jet.newJob(coGroupBuild()).join();
+            validateCoGroupResults();
+        } finally {
+            Jet.shutdownAll();
+        }
+    }
+
+    private void validateCoGroupResults() {
+        IMap<Integer, Tuple3<List<PageVisit>, List<AddToCart>, List<Payment>>> result = jet.getMap(RESULT);
+        printImap(result);
```
jbartok: The result looks like this:
```
result:
12->([PageVisit{3}, PageVisit{4}], [AddToCart{23}, AddToCart{24}], [Payment{33}, Payment{34}])
11->([PageVisit{1}, PageVisit{2}], [AddToCart{21}, AddToCart{22}], [Payment{31}, Payment{32}])
```
Would be good if the `Event` objects would print their `userId`s too, because then we could see that the result is correct.

jbartok:
When looking at these (and other code samples) what often comes to mind is that it would be really nice if the samples would directly and explicitly link to the features they demonstrate in the user manual. 

Maybe even have some advanced documentation format in the future where you could easily jump between manual and samples simply by clicking links. Not sure if it's worth our time now, but it would be surely nice.

jbartok:
One more non-practical suggestion: maybe the/some code samples could/should contain not only a way to show that this or that thing is computed correctly but also a performance version, an optional run that illustrates how fast it runs, even on a local cluster, for a large data volume. I'm thinking in terms of a standardised `runSimple` and `runLarge` methods for samples.

Being new to this I sometimes find myself not realising/suspecting how fast Jet actually is. Maybe having samples that show this to anybody who takes the time to run them would be good advertising. Just an idea.

cangencer:
Regarding the other comments, they're far outside the scope of the PR, but feel to free to create issues.

PR #1556, opened 2019-08-21
Command line interface fixes

viliam-durina:


hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetCommandLine.java
```diff
@@ -164,9 +165,7 @@ public void run() {
         // top-level command, do nothing
     }
 
-    @Command(description = "Submits a job to the cluster",
-            mixinStandardHelpOptions = true
```
cangencer: are these mixed in by default? so this option was redundant?

hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetCommandLine.java
```diff
@@ -482,8 +477,16 @@ private void println(String msg) {
         out.println(msg);
     }
 
-    private static String shorten(String name, int length) {
-        return name.substring(0, Math.min(name.length(), length));
+    private static String shortenAndSanitize(String name, int length) {
+        String result = name
+                .substring(0, Math.min(name.length(), length))
+                .replaceAll("\\s", " ")
+                .replaceAll("[^\\p{Print}]", "?");
```
cangencer: what does this do? can we write an explanation?

viliam-durina: `\p{Print}` are printable characters. Wrapped in `[^]` negates it. That's standard regex.

hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetCommandLine.java
viliam-durina: They are not. But we don't need it for individual commands. We have:
```
jet help COMMAND
```
This mixin will allow also this with the same effect:
```
jet COMMAND -h
```

hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetCommandLine.java
cangencer: We shouldn't change the job's name. I think best to replace \n and \t with their equivalents in the printed out string. If you only see the sanitzed name how can you restore from that snapshot without seeing the actual one? Also the escaping is only applied to SS name and not job name.

cangencer: Also there is no test testing this method since it's only applied to list-snapshots and the test for list-snapshots only tests with empty list.

viliam-durina: Yeah, I first sanitized all names. Then I realized I should not escape the name you need, so I only escaped and shortened the job name in `list-snapshots`. The snapshot name is printed last and is not sanitized. Also the job name in `list-jobs` is not sanitized and printed last. If the sanitization changed anything, an asterisk is appended.

But using shell escaping is interesting idea, you could copy-paste it to a command.

PR #1563, opened 2019-08-22
Match upstream parallelism when adding timestamp transform

### LINKED ISSUE #1527: TimestampTransform should have same parallelism as upstream
When a batch stream is added timestamps currently it causes addition of a unicast edge and maximum parallelism. This should instead use same parallelism as upstream and isolated edge to avoid shuffling the data.

eminn:
Fixes #1527

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Vertex.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Vertex.java
@@ -29,6 +29,7 @@ import java.util.function.UnaryOperator;
 
 import static com.hazelcast.jet.impl.util.Util.checkSerializable;
 import static com.hazelcast.util.Preconditions.checkNotNull;
+import static java.lang.Math.min;
 
 /**
  * Represents a unit of data processing in a Jet computation job. Conceptually,
@@ -122,6 +123,24 @@ public class Vertex implements IdentifiedDataSerializable {
             throw new IllegalArgumentException("Parallelism must be either -1 or a positive number");
         }
         return parallelism;
+    }
+
+    /**
+     * Says whether the given integer is valid as the value of {@link
+     * #localParallelism(int) localParallelism}.
+     */
+    public static int determineLocalParallelism(Vertex vertex, int defaultParallelism) {
+        int localParallelism = vertex.localParallelism;
+        int preferredLocalParallelism = vertex.metaSupplier.preferredLocalParallelism();
+        Vertex.checkLocalParallelism(preferredLocalParallelism);
+        Vertex.checkLocalParallelism(localParallelism);
+        return localParallelism != LOCAL_PARALLELISM_USE_DEFAULT
+                ? localParallelism
+                : preferredLocalParallelism != LOCAL_PARALLELISM_USE_DEFAULT
+                    ? defaultParallelism == -1
+                        ? preferredLocalParallelism
+                        : min(preferredLocalParallelism, defaultParallelism)
+                    : defaultParallelism;
     }
 
     /**
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Vertex.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Vertex.java
@@ -127,8 +127,8 @@ public class Vertex implements IdentifiedDataSerializable {
 
     /**
      * Determines the local parallelism value for the vertex by looking
-     * it's local parallelism and meta supplier's preferred local parallelism.
-     *
+     * its local parallelism and meta supplier's preferred local parallelism.
+     * <p>
      * If none of them is set, returns the provided default parallelism
      */
     public static int determineLocalParallelism(Vertex vertex, int defaultParallelism) {
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Vertex.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Vertex.java
@@ -131,15 +131,15 @@ public class Vertex implements IdentifiedDataSerializable {
      * <p>
      * If none of them is set, returns the provided default parallelism
      */
-    public static int determineLocalParallelism(Vertex vertex, int defaultParallelism) {
-        int localParallelism = vertex.localParallelism;
-        int preferredLocalParallelism = vertex.metaSupplier.preferredLocalParallelism();
-        Vertex.checkLocalParallelism(preferredLocalParallelism);
-        Vertex.checkLocalParallelism(localParallelism);
+    public int determineLocalParallelism(int defaultParallelism) {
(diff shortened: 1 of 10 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Vertex.java
```diff
@@ -124,6 +125,26 @@ public static int checkLocalParallelism(int parallelism) {
         return parallelism;
     }
 
+    /**
+     * Determines the local parallelism value for the vertex by looking
+     * it's local parallelism and meta supplier's preferred local parallelism.
```
cangencer: "its"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Vertex.java
```diff
@@ -124,6 +125,26 @@ public static int checkLocalParallelism(int parallelism) {
...
+     *
```
cangencer: missing <p>

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/transform/TimestampTransform.java
```diff
@@ -41,10 +43,12 @@ public TimestampTransform(
...
     @Override
     public void addToDag(Planner p) {
+        PlannerVertex upstream = p.xform2vertex.get(this.upstream().get(0));
+        int localParallelism = determineLocalParallelism(upstream.v, upstream.v.getLocalParallelism());
         PlannerVertex pv = p.addVertex(
-                this, name(), localParallelism(), insertWatermarksP(eventTimePolicy)
+                this, name(), localParallelism, insertWatermarksP(eventTimePolicy)
         );
-        p.addEdges(this, pv.v);
+        p.dag.edge(between(upstream.v, pv.v).isolated());
```
cangencer: The right way to configure edge is to use p.addEdges(this, pv.v, edge -> edge.isolated()), this doesn't take into account multiple output ordinals.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Vertex.java
```diff
@@ -124,6 +125,26 @@ public static int checkLocalParallelism(int parallelism) {
         return parallelism;
     }
 
+    /**
+     * Determines the local parallelism value for the vertex by looking
+     * its local parallelism and meta supplier's preferred local parallelism.
+     * <p>
+     * If none of them is set, returns the provided default parallelism
+     */
+    public static int determineLocalParallelism(Vertex vertex, int defaultParallelism) {
```
viliam-durina: This should be an instance method now.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/BatchStageTest.java
```diff
@@ -870,4 +871,19 @@ public void customTransform_keyed() {
         // multiple processors will observe the same keys and the counts won't match.
         assertEquals("0\n1", streamToString(sinkStreamOf(Integer.class), Object::toString));
     }
+
+    @Test
+    public void batchAddTimestampPreserveOrderTest() {
```
viliam-durina: This can be a unit test I think. It can create a Pipeline, convert it to a DAG and the compare the determined LP of the vertices. It should also have 2 variants: whether the upstream does or does not have a preferred LP in the metaSupplier.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/BatchStageTest.java
```diff
@@ -870,4 +876,73 @@ public void customTransform_keyed() {
...
+        // Then
+        Vertex tsVertex = dag.getVertex("add-timestamps");
+        assertEquals(lp, tsVertex.getLocalParallelism());
+    }
+
+
+    @Test
+    public void batchAddTimestampMatchingUpstreamParallelism_whenUpstreamDoesntHavePreferredLocalParallelism() {
+        // Given
+        BatchSource<Integer> items = TestSources.items(1, 2, 3);
```
viliam-durina: `TestSources.items` _has_ preferred LP=1.
Use 
```java
ProcessorMetaSupplier.of(noopP())
```

hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/BatchStageTest.java
```diff
@@ -870,4 +876,74 @@ public void customTransform_keyed() {
...
+        p.drawFrom(source)
+         .addTimestamps(o -> 0L, 0)
+         .drainTo(assertOrdered(items));
+
+        // Then
+        execute();
+    }
+
+    @Test
+    public void batchAddTimestampMatchingUpstreamParallelism_whenUpstreamHavePreferredLocalParallelism() {
```
cangencer: `addTimestamps_whenUpstreamHavePreferredLocalParallelism_thenMatchUpstream` or similar for other tests

viliam-durina: `addTimestamps_when_upstreamHasPreferredLocalParallelism_then_lpMatchUpstream`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/transform/StreamSourceTransform.java
```diff
@@ -85,6 +86,9 @@ public void addToDag(Planner p) {
             String v1name = name();
             Vertex v1 = p.dag.newVertex(v1name, metaSupplierFn.apply(eventTimePolicy))
                              .localParallelism(localParallelism());
+            if (localParallelism() == LOCAL_PARALLELISM_USE_DEFAULT) {
```
cangencer: it would be better to explicitly set it to a local variable and then assign it to pv2 rather than changing parallelism of the transform, as done in TimestampTransform.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/transform/StreamSourceTransform.java
```diff
@@ -85,8 +86,11 @@ public void addToDag(Planner p) {
             String v1name = name();
             Vertex v1 = p.dag.newVertex(v1name, metaSupplierFn.apply(eventTimePolicy))
                              .localParallelism(localParallelism());
+            int localParallelism = localParallelism() != LOCAL_PARALLELISM_USE_DEFAULT
```
cangencer: can be simplified as `int localParallelism = v1.determineLocalParallelism(localParallelism());`

PR #1571, opened 2019-08-26
Default to a lower parallelism(2) in case of non-cooperative context factory

eminn:
Users still have the ability to override the default

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/transform/ProcessorTransform.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/transform/ProcessorTransform.java
@@ -36,17 +36,14 @@ import static com.hazelcast.jet.core.processor.Processors.mapUsingContextP;
 public class ProcessorTransform extends AbstractTransform {
     public static final int NON_COOPERATIVE_DEFAULT_LOCAL_PARALLELISM = 2;
     final ProcessorMetaSupplier processorSupplier;
-    final boolean isCooperative;
 
     ProcessorTransform(
             @Nonnull String name,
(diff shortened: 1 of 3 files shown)


hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/ProcessorTransformParallelismTest.java
```diff
@@ -0,0 +1,344 @@
...
+
+    private static final int localParallelism = 11;
+
+
+    @Parameters(name = "{index}: transform={4}")
+    public static Collection<Object[]> data() {
+        return Arrays.asList(
+                new Object[][] {
+                        {
+
```
cangencer: indentation looks strange here

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/transform/ProcessorTransform.java
```diff
@@ -89,12 +94,23 @@ public static ProcessorTransform customProcessorTransform(
...
     }
 
     @Override
     public void addToDag(Planner p) {
-        PlannerVertex pv = p.addVertex(this, name(), localParallelism(), processorSupplier);
+        PlannerVertex pv = p.addVertex(this, name(), determineLocalParallelism(), processorSupplier);
         p.addEdges(this, pv.v);
     }
+
+    protected int determineLocalParallelism() {
```
viliam-durina: Rather than this it would be better to use ProcessorMetaSupplier with a preferred parallelism. The difference is that if, for some reason, there's only 1 cpu core, it will use 1 thread instead of 2.

Issue #1569, opened 2019-08-26
Don't render metrics when completing a job if they are not saved


viliam-durina:
Currently the metrics are rendered as a part of `CompleteExecutionOperation` even if we're not saving them. This is wasteful, esp. in very short-lived jobs.
Credits: @cangencer

PR #1576, opened 2019-08-27
Stop rendering metrics for jobs that don't have them enabled

jbartok:
Fixes https://github.com/hazelcast/hazelcast-jet/issues/1569
Fixes https://github.com/hazelcast/hazelcast-jet/issues/1580

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
@@ -95,8 +95,7 @@ public interface Job {
      * are enabled. Otherwise empty metrics will be returned.
      * <p>
      * Keep in mind that the collections may occur at different times on
-     * each member, metrics from various members aren't from the same instant
-     * of time.
+     * each member, metrics from various members aren't from the same instant.
      * <p>
      * When a job is restarted (or resumed after being previously suspended)
      * the metrics are reset too, their values will reflect only updates
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
@@ -88,29 +88,26 @@ public interface Job {
      * Returns a snapshot of the current values of all job-specific metrics.
      * <p>
      * While the job is running the metric values are updated periodically
-     * (see {@link MetricsConfig#setCollectionIntervalSeconds}).
+     * (see {@link MetricsConfig#setCollectionIntervalSeconds metrics
+     * collection interval}), assuming
+     * that both {@link MetricsConfig#setEnabled global metrics collection}
+     * and {@link JobConfig#isMetricsEnabled() per job metrics collection}
+     * are enabled. Otherwise empty metrics will be returned.
      * <p>
-     * Once a job stops executing (successfully, after a failure, cancellation,
-     * or temporarily while suspended) the metrics will have their most
-     * recent values (i.e. the last metric values from the moment before the
-     * job completed).
+     * Keep in mind that the collections may occur at different times on
+     * each member, metrics from various members aren't from the same instant
+     * of time.
      * <p>
      * When a job is restarted (or resumed after being previously suspended)
      * the metrics are reset too, their values will reflect only updates
      * from the latest execution of the job.
      * <p>
-     * The method returns empty metrics if either
-     * {@link MetricsConfig#setEnabled global metrics collection} or
-     * {@link JobConfig#isMetricsEnabled() per job metrics collection}
-     * are disabled or until the first metric collection takes place.
-     * <p>
-     * For completed jobs, metrics are only persisted if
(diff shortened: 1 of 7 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/operation/CompleteExecutionOperation.java
```diff
@@ -44,14 +44,16 @@
 public class CompleteExecutionOperation extends Operation implements IdentifiedDataSerializable {
 
     private long executionId;
+    private boolean jobMetricsEnabled;
```
cangencer: shouldn't it be `saveMetrics` ?

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/JobMetrics_MiscTest.java
```diff
@@ -63,6 +63,25 @@ public void setup() {
...
     }
 
+    @Test
+    public void when_jobMetricsDisabled_then_emptyMetrics() throws Throwable {
+        DAG dag = new DAG();
+        dag.newVertex("v1", MockP::new);
+        dag.newVertex("v2", (SupplierEx<Processor>) NoOutputSourceP::new);
+        Job job = jet().newJob(dag);
+
+        //when
```
viliam-durina: The `// When` comment should be at the place where you do the _when_ action from the method name. Here it's where you disable the job metrics.

Also it's not clear to me whether you disable metrics collection or metrics saving. Looks like collection, but this PR is about metrics saving. Neither is disabled explicitly in this test.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/JobMetrics_MiscTest.java
```diff
@@ -63,6 +63,25 @@ public void setup() {
...
+        NoOutputSourceP.executionStarted.await();
+        assertJobStatusEventually(job, JobStatus.RUNNING);
+        //then
+        assertTrueAllTheTime(() -> assertEmptyJobMetrics(job), 1);
```
viliam-durina: 1 second is pretty short, collection interval is 5 seconds by default, I don't see it shortened in this test.

jbartok:
My understanding was lacking about the whole point of this issue. Will push improved changes in a sec.

cangencer:
Please add tags + milestone

jbartok:
Fixes https://github.com/hazelcast/hazelcast-jet/issues/1580

hazelcast-jet-core/src/main/java/com/hazelcast/jet/Job.java
```diff
@@ -88,29 +88,26 @@ default String getName() {
...
+     * and {@link JobConfig#isMetricsEnabled() per job metrics collection}
+     * are enabled. Otherwise empty metrics will be returned.
      * <p>
-     * Once a job stops executing (successfully, after a failure, cancellation,
-     * or temporarily while suspended) the metrics will have their most
-     * recent values (i.e. the last metric values from the moment before the
-     * job completed).
+     * Keep in mind that the collections may occur at different times on
+     * each member, metrics from various members aren't from the same instant
+     * of time.
```
viliam-durina: "instant of time" -> "instant"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -451,7 +451,14 @@ public boolean isMetricsEnabled() {
 
     /**
      * Returns whether metrics should be stored in the cluster after the job
-     * completes. If enabled, metrics can be retrieved by calling {@link Job#getMetrics()}.
+     * completes. If enabled, metrics can be retrieved for all non-running
+     * jobs (have completed successfully, have failed, have been cancelled
```
viliam-durina: 1. We normally document the set-methods and the get-methods refer to those. The reason is that the user mostly uses only the set-methods, the get-methods are used by Jet internally.

2. it doesn't apply to "all non-running jobs" but only to this job, when it's not running.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/operation/CompleteExecutionOperation.java
```diff
@@ -44,14 +44,16 @@
...
     private long executionId;
+    private boolean saveMetrics;
     private Throwable error;
     private RawJobMetrics response;
 
     public CompleteExecutionOperation() {
     }
 
-    public CompleteExecutionOperation(long executionId, Throwable error) {
+    public CompleteExecutionOperation(long executionId, boolean saveMetrics, Throwable error) {
```
viliam-durina: I'd rename `saveMetrics` to `collectMetrics`. Metrics are not saved here.

Issue #1565, opened 2019-08-23
change `cluster.sh` default group name


cangencer:
cluster.sh uses default group name of `dev` and `dev-pass` where it should use jet specific ones.

PR #1577, opened 2019-08-27
Fix group name, password, remove tab indentation

viliam-durina:


cangencer:
Fixes #1565

hazelcast-jet-distribution/src/main/resources/bin/cluster.sh
```diff
@@ -60,21 +60,21 @@ if [ -z "$PORT" ]; then
...
+    PASSWORD="jet-pass"
 fi
 
 if [ -z "$ADDRESS" ]; then
     echo "No specific ip address is defined, running script with default ip : '127.0.0.1'."
     ADDRESS="127.0.0.1"
 fi
 
-command -v curl >/dev/null 2>&1 || { echo >&2 "Cluster state script requires curl but it's not installed. Aborting."; exit -1; }
+command -v curl >/dev/null 2>&1 || { echo >&2 "Cluster state script requires curl but it's not installed. Aborting."; exit 255; }
```
cangencer: why not `1`?

viliam-durina: `shellchecker` told me that -1 isn't valid. Why not 255 then? :)

cangencer: seems like a non-standard error. 1 is typically used a a general error exit code. `255` means "exit status out of range"

cangencer:
verify

Issue #1584, opened 2019-08-29
Harmonize the semantics of various metrics related config params


jbartok:
We have multiple configuration parameters related to metrics right now and it's not straightforward to make sense of what they do. We should bring them in line a bit. I propose their meanings become following (changes relative to current behaviour are minimal):

1. `MetricsConfig.enable` - master switch for everything metrics related; if disabled no metrics related activity whatsoever will be going on, regardless of the settings of any other metrics related flags; defaults to `true`

2. `MetricsConfig.jmxEnabled` - can be used to toggle metrics publishing over JMX on/off, if metrics are turned on by the master switch; no effect otherwise; defaults to `true`

3. `JobConfig.enableMetrics` - can be used to toggle job metrics publishing on/off (via `Job` objects), if metrics are turned on by the master switch; no effect otherwise; defaults to   
`true`

4. `JobConfig.storeMetricsAfterJobCompletion` - can be used to toggle saving final metrics for finished jobs on/off (to be accessed via `Job` objects), `if` the metrics master switch is on `and if` job metrics are also turned on; no effect otherwise; defaults to `false`

Btw. seems to me that for completeness we are missing one flag, which would be `MetricsConfig.mcEnabled` which could be used to toggle publishing to the MC on/off, independently from JMX and Job publishing.

PR #1586, opened 2019-08-29
Harmonize metrics related config params

jbartok:
Fixes https://github.com/hazelcast/hazelcast-jet/issues/1584


hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/MetricsConfig.java
```diff
@@ -46,9 +46,10 @@
     private int intervalSeconds = DEFAULT_METRICS_COLLECTION_SECONDS;
 
     /**
-     * Sets whether metrics collection should be enabled for the node. If
-     * enabled, Hazelcast Jet Management Center will be able to connect to this
-     * member. It's enabled by default.
+     * Master switch for all metrics related functionality. Enabling it is
```
cangencer: this javadoc is too technical, I would keep the old one and mention that this option is required for other metrics functionality.

PR #1589, opened 2019-08-29
Update job metrics info in RefMan

jbartok:


cangencer:
Could the list of metrics moved to own section, rather than being under JMX?

jbartok:
I've reorganised the metrics section (3.8) of the reference manual completely, pls. take a look. I personally think it's much better now, but have added Viliam as a reviewer, because I'm sure he will spot things which need a bit of "touching up".

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
```diff
@@ -281,7 +281,7 @@ public void run() {
 
         @Probe(name = "taskletCount")
         private final CopyOnWriteArrayList<TaskletTracker> trackers;
-        @Probe
+        @Probe(name = "iterationCount")
```
cangencer: are these not in `MetricNames`?

hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/log-debug.adoc
```diff
@@ -187,71 +180,116 @@ summed. If the processor has input edges with different priority, only
...
 low load.
-
+|**proc=<N>, ordinal=<M>** +
+Each Jet member will have one instance of these metrics for each edge _M_ (input or
+output) of each processor _N_. _N_ is the global processor index and _M_ is either
+the ordinal of the edge or has the value _snapshot_ for output items
+written to state snapshot.
 |===
 
 === Hazelcast IMDG Metrics
```
cangencer: this may be worth moving up, or mentioning earlier since there is a lot of metrics from IMDG. We can give examples of what kind of metrics are available (i.e. map stats, connection stats etc)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
jbartok: I've only added those metric names to `MetricNames` which are retrievable via API, so only the job specific ones. 

Do you want me to add them? 

Actually there are a lot more non-job specific metrics we define and which weren't mentioned in the manual... They are specific to certain processors, but still, they are there. Should I add those too?

cangencer: I don't think we need to mention the processor-specific ones for now, but the other general metrics should be added to the manual since they're still retrievable via JMX

jbartok: Ok.

PR #1595, opened 2019-09-03
Add causes from members to job failed with a TCE

### LINKED ISSUE #1594: MasterJobContext.getResult hides exception
MasterJobContext hides any exception which is one of `(e instanceof CancellationException
                        || e instanceof TerminatedWithSnapshotException
                        || isTopologyException(e))` and wraps it with `TopologyException` which may hide the real cause, which can be non-topology change related.

viliam-durina:
Multiple exceptions are translated into a TopologyChangeException (TCE)
in `MasterJobContext.getResult`. Sometimes it's not clear what
exceptions they were. This commit adds them to the message of the TCE.

Fixes #1594

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterJobContext.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterJobContext.java
@@ -43,6 +43,7 @@ import com.hazelcast.jet.impl.util.LoggingUtil;
 import com.hazelcast.jet.impl.util.NonCompletableFuture;
 import com.hazelcast.jet.impl.util.Util;
 import com.hazelcast.logging.ILogger;
+import com.hazelcast.nio.Address;
 import com.hazelcast.spi.ExecutionService;
 import com.hazelcast.spi.Operation;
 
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterJobContext.java
```diff
@@ -526,18 +525,22 @@ private Throwable getResult(String opName, Collection<Map.Entry<MemberInfo, Obje
...
+                .collect(Collectors.groupingBy(
+                        e -> e instanceof CancellationException
+                                || e instanceof TerminatedWithSnapshotException
+                                || isTopologyException(e),
+                        Collectors.toList()));
+        List<Throwable> topologyFailures = splitFailures.getOrDefault(true, emptyList());
+        List<Throwable> otherFailures = splitFailures.getOrDefault(false, emptyList());
+
+        if (!otherFailures.isEmpty()) {
+            return otherFailures.get(0);
```
cangencer: would it be better to list all of them, along them with the member info?

viliam-durina:
verify

viliam-durina:
The failure was unrelated to this PR, logged the issue in #1598.

PR #1600, opened 2019-09-04
Fix different exception from join if problem is same

### LINKED ISSUE #1598: Different error thrown when joining before or after a job completes
When `Job.join()` is called while the job is running, the exception is the exception actually thrown by the job. If it's called after the job failed, the exception is recreated as `new JetException(JobResult.failureText)`, doesn't contain the original exception in causes and the original stack trace.

viliam-durina:
Fixes #1598

One difference still remains: if a temporary master context is completed
in `JobCoordinatorService.completeMasterContextIfJobAlreadyCompleted()`,
it will have different error.

Also contains refactor of `ExecutionLifecycleTest` to use
`SimpleTestInClusterSupport`: some tests created a second cluster.

The test `when_job_withNoSnapshots_completed_then_noSnapshotMapsLeft`
was outdated.


viliam-durina:
verify

viliam-durina:
the build failed due to network issues talking to GH, S3

viliam-durina:
verify

viliam-durina:
verify

PR #1601, opened 2019-09-04
Copy-paste system properties list to refman

### LINKED ISSUE #1541: Create section about Jet System Properties in Reference Manual
Jet uses some system properties (see `com.hazelcast.jet.core.JetProperties`). They should be added to reference manual to show them to users. It can be done in the same way as it is for IMDG reference manual, see https://docs.hazelcast.org/docs/latest/manual/html-single/index.html#system-properties.

viliam-durina:
Fixes #1541

### DIFF
--- a/hazelcast-jet-reference-manual/src/main/asciidoc/appendix.adoc
+++ b/hazelcast-jet-reference-manual/src/main/asciidoc/appendix.adoc
@@ -84,7 +84,123 @@ for the previous version can be used without any modification.
 
 Jet currently provides metrics to Management Center and also through
 other means such as JMX. The metrics names may change between MINOR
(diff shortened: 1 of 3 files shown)


hazelcast-jet-reference-manual/src/main/asciidoc/appendix.adoc
```diff
@@ -84,7 +84,123 @@ for the previous version can be used without any modification.
...
+|===
+|Property Name
+| Default Value
+| Type
+| Description
+
+|`jet.job.scan.period`
+|5000
+|long
+|Jet will periodically check for new jobs to start and perform cleanup
```
olukas: Please add information that time is in ms.

PR #1604, opened 2019-09-05
Change max parallel ops limit to 1 per processor for writeMapP

cangencer:
Currently the limit is higher per processor, which may overload the cluster if the writes are slow. This allows the parallelism of the sink to be controlled in a more fine-grained way.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteMapP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteMapP.java
@@ -36,6 +36,7 @@ import static com.hazelcast.jet.impl.connector.HazelcastWriters.handleInstanceNo
 import static com.hazelcast.jet.impl.util.ExceptionUtil.sneakyThrow;
 
 public final class WriteMapP<K, V> implements Processor {
+    private static final int MAX_LOCAL_PARALLEL_OPS = 1;
     private final AtomicInteger numParallelOps = new AtomicInteger();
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteMapP.java
```diff
@@ -64,7 +64,6 @@ public void init(@Nonnull Outbox outbox, @Nonnull Context context) {
         // get 1. The putAll operation is already bulky, it doesn't help to have many in
```
viliam-durina: This comment should be moved to the constant

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteMapP.java
```diff
@@ -76,7 +75,7 @@ public boolean tryProcess() {
     @Override
     public void process(int ordinal, @Nonnull Inbox inbox) {
         checkFailure();
-        if (Util.tryIncrement(numParallelOps, 1, parallelOpsLimit)) {
+        if (Util.tryIncrement(numParallelOps, 1, MAX_LOCAL_PARALLEL_OPS)) {
```
viliam-durina: We can also replace this with `AtomicBoolean opPending` and `if (opPending.compareAndSet(false, true))`

cangencer: I'll keep it for now in case we decide to increase the limit in the future, or make it configurable in some other way.

PR #1612, opened 2019-09-09
Do not rewrap exception in JetException if already a JetException

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -383,7 +383,17 @@ void cleanup(NodeEngine nodeEngine) {
         logger.fine("Job cleanup took " + TimeUnit.NANOSECONDS.toMillis(elapsed) + "ms");
     }
 
-    private long jobIdFromMapName(String map, String prefix) {
+    private static String toErrorMsg(@Nullable Throwable error) {
+        if (error == null) {
+            return null;
+        }
+        if (error instanceof JetException) {
+            return error.toString().substring((JetException.class.getName() + ": ").length());
```
viliam-durina: This can be just `return error.getMessage();`. Or better
```java
return error.getCause() != null ? error.getCause().toString() : error.getMessage();
```

cangencer: Wouldn't you lose the message of the parent exception that way?

viliam-durina: Ah, that's true. Then `getMessage` would suffice because `Throwable.toString` is `className: message`. Also currently if the message is null, there's no `": :" and we'll fail.

cangencer: OK fixed that, however the test comparing exceptions before & after execution still fails because after execution, we lose the original call site of the exception.

PR #1613, opened 2019-09-09
Fix javadoc for assertionCollectedEventually

### LINKED ISSUE #1609: AssertionCompletedException not propagated from Sinks
I am trying to use the assertion support in Jet 3.2, but it's not working as expected. I can see `com.hazelcast.jet.pipeline.test.AssertionCompletedException: Assertion passed successfully
` in a log, but the `AssertionCompletedException` is not propagated successfully. 

The reproduce bellow is expected to pass, but it fails inside the `assertInstanceOf` as the `actual` is set to `null`. 

```java
public class TestTest {
    @Test
    public void foo() {
        Pipeline pipeline = Pipeline.create();
        pipeline.drawFrom(TestSources.itemStream(1))
                .withIngestionTimestamps()
                .drainTo(AssertionSinks.assertCollectedEventually(60, e -> {} /*always accepted*/));

        JetInstance instance = Jet.newJetInstance();

        Job job = instance.newJob(pipeline);

        try {
            job.join();
        } catch (CompletionException e) {
            Throwable jetException = e.getCause();
            Throwable cause = jetException.getCause();
            assertInstanceOf(AssertionCompletedException.class, cause);
        }
    }

    private static <E> void assertInstanceOf(Class<E> expected, Object actual) {
        assertNotNull(actual);
        assertTrue(actual + " is not an instanceof " + expected.getName(), expected.isAssignableFrom(actual.getClass()));
    }
}
```

cangencer:
Fixes #1609

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/test/AssertionSinks.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/test/AssertionSinks.java
@@ -163,7 +163,7 @@ public final class AssertionSinks {
      * } catch (CompletionException e) {
      *     String errorMsg = e.getCause().getMessage();
      *     Assert.assertTrue(
-     *             "Job failed with another reason then AssertionCompletedException",
+     *             "Job was expected to complete with AssertionCompletedException, but completed with: " + e.getCause(),
      *             errorMsg.contains(AssertionCompletedException.class.getName())
      *     );
      * }</pre>
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/test/AssertionSinks.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/test/AssertionSinks.java
@@ -153,6 +153,20 @@ public final class AssertionSinks {
      * used to terminate the job so that you can {@code join()} it. This also
      * requires that there are no other assertions in the job as this one can
      * complete the job before the other ones succeeded.
+     * <p>
+     * The assertion can be validated as follows:
+     * <pre>{@code
+     * try {
+     *     jetInstance.newJob(p).join();
+     *     Assert.fail("Job should have completed with an AssertionCompletedException, " +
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/test/AssertionSinks.java
```diff
@@ -153,6 +153,20 @@ private AssertionSinks() {
...
+     * The assertion can be validated as follows:
+     * <pre>{@code
+     * try {
+     *     jetInstance.newJob(p).join();
+     *     Assert.fail("Job should have completed with an AssertionCompletedException, " +
+     *             "but completed normally");
+     * } catch (CompletionException e) {
+     *     String errorMsg = e.getCause().getMessage();
+     *     Assert.assertTrue(
+     *             "Job failed with another reason then AssertionCompletedException",
```
viliam-durina: "then" -> "than"
We should also include the actual reason. Maybe `"Job was expected to complete with AssertionCompletedException, but completed with: " + e.getCause()`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/test/Assertions.java
```diff
@@ -226,14 +226,15 @@ private Assertions() {
...
+     *     jetInstance.newJob(p).join();
+     *     Assert.fail("Job should have completed with an AssertionCompletedException, " +
+     *             "but completed normally");
      * } catch (CompletionException e) {
-     *     Throwable jetException = e.getCause();
-     *     assertInstanceOf(AssertionCompletedException.class, jetException.getCause());
-     * }
+     *     String errorMsg = e.getCause().getMessage();
+     *     Assert.assertTrue(
+     *             "Job failed with another reason then AssertionCompletedException",
```
viliam-durina: Again

cangencer:
Currently it's marked as `@Beta`, but we may consider such a utility method in next release.

PR #1624, opened 2019-09-11
Fix few Job methods when master is going down

viliam-durina:
`Job.getJobStatus` can fail if executed when the master is going down and the client doesn't know yet. This PR fixes 3 similar methods on the client, but similar issue remains on member and with other methods. `newJob` handles it on its own and differently.

Also rewrite the `JobTest` to use shared cluster, went from 130s to 11+13 seconds locally.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClientJobProxy.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClientJobProxy.java
@@ -39,18 +39,25 @@ import com.hazelcast.jet.impl.client.protocol.codec.JetSubmitJobCodec;
 import com.hazelcast.jet.impl.client.protocol.codec.JetTerminateJobCodec;
 import com.hazelcast.logging.LoggingService;
 import com.hazelcast.nio.serialization.Data;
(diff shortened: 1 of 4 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClientJobProxy.java
```diff
@@ -183,6 +181,21 @@ private ClientInvocation invocation(ClientMessage request, Address invocationAdd
...
     }
 
+    private <T> T callAndRetryIfTargetNotFound(Callable<T> action) {
+        for (;;) {
+            try {
+                return action.call();
+            } catch (Exception e) {
+                if (e instanceof ExecutionException && e.getCause() instanceof TargetNotMemberException) {
+                    // ignore the TargetNotMemberException and retry with new master
+                    LockSupport.parkNanos(RETRY_DELAY_NS);
```
cangencer: we should probably have some fallback so that it doesn't retry indefinitely?

viliam-durina:
verify

cangencer:
verify

cangencer:
can we merge master into this?

PR #1632, opened 2019-09-12
Code sample for stateful mapping

mtopolnik:


examples/pattern-matching/src/main/java/com/hazelcast/jet/examples/patternmatching/TransactionTracking.java
```diff
@@ -0,0 +1,139 @@
...
+    private static final int SLIDING_STEP_MILLIS = 20;
+    private static final int EVENTS_PER_SECOND = 2_000;
+
+    private static Pipeline buildPipeline() {
+        Pipeline p = Pipeline.create();
+
+        // This stage emits pairs (transactionId, transactionDurationMillis).
+        // Transaction duration can also be the special value TIMED_OUT.
+        StreamStage<Entry<String, Long>> transactionOutcomes =
+                p.drawFrom(eventSource(EVENTS_PER_SECOND)).withTimestamps(TransactionEvent::timestamp, MAX_LAG)
```
cangencer: could we describe what's emitted from the source?

examples/pattern-matching/src/main/java/com/hazelcast/jet/examples/patternmatching/TransactionTracking.java
```diff
@@ -0,0 +1,139 @@
...
+                         (startEnd, transactionId, wm) -> (startEnd[0] == null || startEnd[1] == null)
+                                 ? entry(transactionId, TIMED_OUT)
+                                 : null
+                 );
+
+        // This pipeline branch computes the mean duration of transactions over a
+        // 10-second sliding window:
+        transactionOutcomes
+                .filter(e -> e.getValue() != TIMED_OUT)
+                .window(sliding(SECONDS.toMillis(SLIDING_WINDOW_SIZE), SLIDING_STEP_MILLIS))
```
cangencer: I think the windowing part and the graph didn't add much value for me. It would be maybe enough just to output the matched transactions and also the timed out ones?

PR #1636, opened 2019-09-13
HDFS new MapReduce API

eminn:
Introduce a new variant in the HDFS source which uses new API

Todo:
- [x] Implement a Sink with the new API


eminn:
verify

eminn:
verify

viliam-durina:
verify

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/HdfsProcessors.java
```diff
@@ -69,12 +67,10 @@ private HdfsProcessors() {
             @Nonnull FunctionEx<? super E, V> extractValueFn
     ) {
         configuration = SerializableConfiguration.asSerializable(configuration);
-        if (configuration instanceof JobConf) {
-            return new WriteHdfsOldApiP.MetaSupplier<>((JobConf) configuration,
-                    extractKeyFn, extractValueFn);
+        if (configuration.get("mapreduce.job.outputformat.class") != null) {
```
eminn: Can we replace the string with the `MRJobConfig#OUTPUT_FORMAT_CLASS_ATTR`?

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/HdfsProcessors.java
```diff
@@ -51,10 +49,10 @@ private HdfsProcessors() {
             @Nonnull Configuration configuration, @Nonnull BiFunctionEx<K, V, R> mapper
     ) {
         configuration = SerializableConfiguration.asSerializable(configuration);
-        if (configuration instanceof JobConf) {
-            return new ReadHdfsOldApiP.MetaSupplier<>((JobConf) configuration, mapper);
-        } else {
+        if (configuration.get("mapreduce.job.inputformat.class") != null) {
```
eminn: Can we replace the string with the MRJobConfig#INPUT_FORMAT_CLASS_ATTR?

eminn:
verify

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
```diff
@@ -139,6 +140,7 @@ private static InputFormat getInputFormat(Configuration configuration) {
 
         static final long serialVersionUID = 1L;
 
+        @SuppressFBWarnings("SE_BAD_FIELD")
```
cangencer: what's the reason for this?

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/HdfsSinks.java
```diff
@@ -38,48 +40,54 @@ private HdfsSinks() {
...
+     * the key which stores the {@code OutputFormat} configuration. If it's
+     * stored under {@value MRJobConfig#OUTPUT_FORMAT_CLASS_ATTR}}, the new API
+     * will be used. Otherwise, the old API will be used. If you get the
+     * configuration from {@link Job#getConfiguration()}, the new API will be
+     * used.
      * <p>
      * No state is saved to snapshot for this sink. After the job is restarted,
      * the items will likely be duplicated, providing an <i>at-least-once</i>
-     * guarantee.
+     * guarantee. TODO [viliam] wrong, don't we overwrite files?
```
cangencer: there is a TODO here

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
```diff
@@ -0,0 +1,255 @@
...
+            throw sneakyThrow(e);
+        }
+
+        return () -> {
+            try {
+                while (reader.nextKeyValue()) {
+                    // we have to clone the the value returned by `reader.getCurrentKey()` because
+                    // the next call to `reader.nextKeyValue()` will mutate the returned value instead
+                    // of creating a new value. Same for value.
+                    K key = serdeWritable(reader.getCurrentKey());
```
cangencer: this seems like a really bad idea and will be very slow. isn't there some other way to do it? In old API you could use .createKey(). Each Reader should have its own createKey/Value implementation.

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
```diff
@@ -0,0 +1,255 @@
...
+                return null;
+            } catch (Exception e) {
+                throw sneakyThrow(e);
+            }
+        };
+    }
+
+    @SuppressWarnings("unchecked")
+    private <T> T serdeWritable(T o) throws IOException, IllegalAccessException, InstantiationException {
+        ((Writable) o).write(cloneBufferDataOutput);
```
cangencer: there needs to be constraint on K, V that they extend `Writable`

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
cangencer: I had another look and seems this method is removed in new api. There's another method called `ReflectionUtils.cloneWritableInto`, but still it will be slow to serialize/deserialize each key/value.

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
eminn: It requires non serializable classes to be marked as transient, but the instances of the subclasses used here is always `Serializable`.

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
```diff
@@ -0,0 +1,255 @@
...
+        Writable newO = (Writable) o.getClass().newInstance();
+        newO.readFields(cloneBuffer.getDataInputStream());
+        o = (T) newO;
+        cloneBuffer.reset();
+        return o;
+    }
+
+    private static InputFormat getInputFormat(Configuration configuration) {
+        Class<?> inputFormatClass = configuration.getClass(MRJobConfig.INPUT_FORMAT_CLASS_ATTR, TextInputFormat.class);
+        return (InputFormat) ReflectionUtils.newInstance(inputFormatClass, configuration);
```
cangencer: looks like this might leak the classloader, since it uses a static cache

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
cangencer: I had one more look, and actually this is completely wrong because there is no guarantee that the items are Writable. When using Avro/Parquet then this is not the case.

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
```diff
@@ -150,7 +153,21 @@ public boolean complete() {
...
         Class<?> inputFormatClass = configuration.getClass(MRJobConfig.INPUT_FORMAT_CLASS_ATTR, TextInputFormat.class);
-        return (InputFormat) ReflectionUtils.newInstance(inputFormatClass, configuration);
+        try {
+            return (InputFormat) ReflectionUtils.newInstance(inputFormatClass, configuration);
+        } finally {
+            removeInputFormatFromCache(inputFormatClass);
+        }
+    }
+
+    private static void removeInputFormatFromCache(Class<?> inputFormatClass) {
```
eminn: May be it is better to clear the cache rather than removing single class from there in the close method ? WDYT?

hazelcast-jet-hadoop/pom.xml
```diff
@@ -59,19 +59,24 @@
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-hdfs</artifactId>
             <version>${hadoop.version}</version>
-            <scope>provided</scope>
+<!--            <scope>provided</scope>-->
```
eminn: Leftover ?

gurbuzali: yes, leftover

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
gurbuzali: since it is a static field, I didn't want to touch more than what we put there. maybe other parts of hadoop uses this util internally

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
```diff
@@ -0,0 +1,289 @@
...
+            cloneBuffer.reset();
+            return o;
+        }
+        return serializationService.toObject(serializationService.toData(o));
+    }
+
+    private static InputFormat getInputFormat(Configuration configuration) {
+        Class<?> inputFormatClass = configuration.getClass(MRJobConfig.INPUT_FORMAT_CLASS_ATTR, TextInputFormat.class);
+        try {
+            return (InputFormat) ReflectionUtils.newInstance(inputFormatClass, configuration);
```
cangencer: why not just create the instance ourselves, rather than creating and then deleting from cache? It doesn't seem to do any magic.

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
cangencer: ok, might be good to add the reasoning as comments

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/HdfsSources.java
```diff
@@ -36,49 +41,81 @@
...
+     * avoid unnecessary copying.
+     * <p>
+     * The source copies the objects by serializing and de-serializing it. The
+     * objects should be either {@link Writable} or serializable in a way that
+     * Jet instance can serialize/de-serialize.
+     * <p>
+     * Here is how you can configure the source. Default value is {@code false}.
+     * <pre>{@code
+     * Configuration conf = new Configuration();
+     * conf.set(HdfsSources.REUSE_OBJECT, "true");
```
cangencer: this setting name is misleading. we NEVER want to reuse same objects when reading as it will break many things with respect to snapshotting, among others.

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
gurbuzali: it does do some magic actually, it tries to set `Configuration` object to the created object if it is configurable and it tries to call `configure` method, again using reflection.

cangencer: but we can also call ReflectionUtils.setConf() explicitly after creation.

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/HdfsSources.java
cangencer: I would suggest something like "CLONE_ON_READ" with default to set to true. With some record readers this is not necessary and we should document which readers clone or not.

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/HdfsSources.java
```diff
@@ -36,49 +41,81 @@
...
+     * <p>
+     * Here is how you can configure the source. Default value is {@code true}.
+     * <pre>{@code
+     * Configuration conf = new Configuration();
+     * conf.set(HdfsSources.COPY_ON_READ, "false");
+     *
+     * BatchSource<Entry<K, V>> source = HdfsSources.hdfs(conf);
+     * }</pre>
+     */
+    public static final String COPY_ON_READ = "JET_HDFS_SOURCE_COPY_ON_READ";
```
cangencer: I think the convention for hadoop properties is like `jet.source.copyonread`

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
```diff
@@ -0,0 +1,293 @@
...
+                    // record-readers return the same object for `reader.getCurrentKey()`
+                    // and `reader.getCurrentValue()` which is mutated for each `reader.nextKeyValue()`
+                    K key = copyIfConfigured(reader.getCurrentKey());
+                    V value = copyIfConfigured(reader.getCurrentValue());
+                    R projectedRecord = projectionFn.apply(key, value);
+                    if (projectedRecord != null) {
+                        return projectedRecord;
+                    }
+                }
+                reader.close();
```
cangencer: this reader won't be closed if exception

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
```diff
@@ -0,0 +1,293 @@
...
+            newO.readFields(cloneBuffer.getDataInputStream());
+            o = (T) newO;
+            cloneBuffer.reset();
+            return o;
+        }
+        return serializationService.toObject(serializationService.toData(o));
+    }
+
+    private static InputFormat getInputFormat(Configuration configuration) {
+        try {
```
cangencer: you can remove this try block and replace the call to getInputFormat() with `uncheckCall()`

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/WriteHdfsNewApiP.java
```diff
@@ -0,0 +1,217 @@
...
+        @Override
+        public int preferredLocalParallelism() {
+            return 2;
+        }
+
+        @Override
+        public void init(@Nonnull Context context) throws Exception {
+            jobContext = new JobContextImpl(configuration, new JobID());
+            OutputFormat outputFormat = getOutputFormat(configuration);
+            // TODO [viliam] remove the use of UUID
```
cangencer: there is a TODO here

gurbuzali: I'll talk with @viliam-durina

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
viliam-durina: Also won't be closed when the job is cancelled.

gurbuzali: extracted to a field, now closed in `close` method

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHdfsNewApiP.java
```diff
@@ -0,0 +1,296 @@
...
+        if (reader != null) {
+            reader.close();
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private Traverser<R> traverseSplit(InputSplit split) {
+        try {
+            TaskAttemptContextImpl attemptContext = new TaskAttemptContextImpl(configuration, new TaskAttemptID());
+            reader = inputFormat.createRecordReader(split, attemptContext);
```
viliam-durina: Now we don't close the reader in the normal case

hazelcast-jet-hadoop/src/test/java/com/hazelcast/jet/hadoop/impl/HdfsTestSupport.java
```diff
@@ -16,16 +16,15 @@
...
-public class HdfsTestSupport extends JetTestSupport {
+public abstract class HdfsTestSupport extends SimpleTestInClusterSupport {
 
     @Before
-    public void before() {
+    public void hdfsSupportBefore() {
         // Tests fail on windows. If you want to run them, comment out this line and
         // follow this instructions: https://stackoverflow.com/a/35652866/952135
-        assumeThatNoWindowsOS();
+//        assumeThatNoWindowsOS(); // TODO [viliam] uncomment
```
cangencer: there is a todo here

examples/hadoop/src/main/java/com/hazelcast/jet/examples/hadoop/HadoopWordCount.java
```diff
@@ -45,12 +45,12 @@
...
  * for the {@code WordCount} class in {@code wordcount} sample.
  * <p>
- * {@link HdfsSources#hdfs(JobConf, BiFunctionEx)} is a source
+ * {@link HadoopSources#inputFormat(Configuration, BiFunctionEx)} is a source
  * that can be used for reading from HDFS given a {@code JobConf}
  * with input paths and input formats. The files in the input folder
  * will be split among Jet processors, using {@code InputSplit}s.
  * <p>
- * {@link HdfsSinks#hdfs(JobConf, FunctionEx, FunctionEx)}
+ * {@link HadoopSinks#hdfs(JobConf, FunctionEx, FunctionEx)}
```
cangencer: should be #outputFormat

examples/hadoop/src/main/java/com/hazelcast/jet/examples/hadoop/avro/HadoopAvro.java
```diff
@@ -50,10 +51,10 @@
 
     private static Pipeline buildPipeline(JobConf jobConfig) {
```
cangencer: we should change this to `Configuration` also

hazelcast-jet-hadoop/src/test/java/com/hazelcast/jet/hadoop/impl/WriteHadoopPTest.java
```diff
@@ -0,0 +1,169 @@
...
+            job.setOutputKeyClass(IntWritable.class);
+            job.setOutputValueClass(IntWritable.class);
+            org.apache.hadoop.mapreduce.lib.output.FileOutputFormat.setOutputPath(job, path);
+            if (outputFormatClass.equals(org.apache.hadoop.mapreduce.lib.output.LazyOutputFormat.class)) {
+                org.apache.hadoop.mapreduce.lib.output.LazyOutputFormat.setOutputFormatClass(job,
+                        org.apache.hadoop.mapreduce.lib.output.TextOutputFormat.class);
+            }
+            return job.getConfiguration();
+        } else {
+            JobConf conf = new JobConf();
```
cangencer: do we still need this part?

examples/hadoop/src/main/java/com/hazelcast/jet/examples/hadoop/avro/HadoopAvro.java
```diff
@@ -50,10 +51,10 @@
...
         Pipeline p = Pipeline.create();
-        p.drawFrom(HdfsSources.<AvroWrapper<User>, NullWritable>hdfs(jobConfig))
+        p.drawFrom(HadoopSources.<AvroWrapper<User>, NullWritable>inputFormat(jobConfig))
```
cangencer: it would make sense to unwrap it at the source here.

hazelcast-jet-hadoop/src/main/java/com/hazelcast/jet/hadoop/impl/ReadHadoopNewApiP.java
```diff
@@ -0,0 +1,284 @@
...
+
+        return () -> {
+            try {
+                while (reader.nextKeyValue()) {
+                    // we clone the key/value if configured so because some of the
+                    // record-readers return the same object for `reader.getCurrentKey()`
+                    // and `reader.getCurrentValue()` which is mutated for each `reader.nextKeyValue()`
+                    R projectedRecord = projectionFn.apply(reader.getCurrentKey(), reader.getCurrentValue());
+                    if (projectedRecord != null) {
+                        return copyIfConfigured(projectedRecord);
```
cangencer: we are missing in the documentation that copy is done after projection

cangencer: also, this method is called for each record, which will do a hash lookup in the configuration

PR #1637, opened 2019-09-13
S3 sink should handle items larger than buffer size

### LINKED ISSUE #1634: File is not uploaded to S3 bucket if it contains long line
In case when item in pipeline results to long line which should be drained to S3 bucket then that file is not uploaded at all. No exception is thrown, job status is set `COMPLETED` but file does not appear in S3 bucket.

To reproduce this issue use part of code like this:
```
StringBuilder sb = new StringBuilder();
String[] items = new String[3];
items[0] = "a";
items[1] = "b";
for (int i = 0; i < 1_000_000; i++) {
    sb.append("1234567890");
}
items[2] = sb.toString();
p.drawFrom(TestSources.items(items))
        .drainTo(S3Sinks.s3(BUCKET_NAME, PREFIX, UTF_8, Main::client, Object::toString));
```
If `for` cycle generates smaller `String` or when it is drained to `logger()` then everything works as expected.

cangencer:
Fixes #1634


cangencer:
verify

PR #1641, opened 2019-09-17
User metrics (OBSOLETE)

jbartok:
This code is meant as a solution proposal for user metrics. 

It covers only the filter operation for now (if accepted we will need to do explicit adjustments for each operation that has to handle user metrics). 

It works (see `UserMetricsTest` unit test) and is the simplest solution I could find until now.

Not much code is involved, should be simple to understand. 

Pls. let me know if this direction is acceptable and if I can go ahead refining it.

I have not put this proposal in the TDD (yet!) because it shouldn't need much explaining and I'm not sure at all, that you will like it and I don't have to find something else.


jbartok:
Have pushed the changes we have discussed after the review. 

There is one problem in this version: multiple different processor instances (`TransformP`s in case of `UserMetricsTest`) get the same instance of `MetricsOperator` (`FilterWithMetrics` in the case of the before mentioned test). So `MetricsOperator.init(MetricsContext)` gets called multiple times on one instance. I will solve this by Serde/cloning the `mapFn` param in `TransformP` constructor.

Then I will update the TDD and start looking into other operations that need to support user metrics.

Anyways, sorry for spamming, this is a stable, reviewable state.

jbartok:
Hmm... should I worry about users overwriting built-in metrics with their custom versions having the same name? I didn't try it, but it should be possible. Do we want to protect them from their own stupidity?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MetricsOperator.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.jet.core.metrics;
+
+/**
+ * Marker interface to be implemented by all classes that want to be able
+ * to retrieve and work with custom metrics.
+ */
+public interface MetricsOperator {
```
cangencer: could we find a different name? What does MetricsOperator mean?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MetricsContext.java
```diff
@@ -0,0 +1,36 @@
...
+ * as Processors.
+ */
+public interface MetricsContext {
+
+    /**
+     * Return a counter that can be used to set values for a custom, per
+     * Processor metric.
+     */
+    @Nonnull
+    AtomicLong getCounter(String name);
```
cangencer: why AtomicLong? isn't it too low level?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MetricsOperator.java
jbartok: `MetricsOperator` is supposed to be an/any entity which can set the value of user metrics; that's why I'm trying names like `MetricsSource`, `MetricsSetter`, `MetricsOperator` and such, none of them particularly good.. will try to find a better name.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MetricsContext.java
jbartok: I think it has to be something thread safe... Should I put some kind of generic wrapper on top of it?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MetricsOperator.java
jbartok: `MetricsGenerator`?

cangencer: something like HasMetrics or ProvidesMetrics

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MetricsContext.java
cangencer: yes, it should be wrapped. we can also support a lambda.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsOperatorUtil.java
```diff
@@ -0,0 +1,123 @@
...
+
+        @Override
+        public P getEx() throws Exception {
+            return supplierEx.getEx();
+        }
+
+        @Override
+        public void init(MetricsContext context) {
+            for (MetricsOperator operator : metricsOperators) {
+                operator.init(context);
```
cangencer: wouldn't they overwrite each other? also wouldn't it be better to merge the counters in one, and still keep one of the "operator" (to be renamed) object.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/ComputeStageImplBase.java
```diff
@@ -114,8 +115,9 @@
     @SuppressWarnings("unchecked")
     <RET> RET attachFilter(@Nonnull PredicateEx<T> filterFn) {
         checkSerializable(filterFn, "filterFn");
-        PredicateEx<T> adaptedFn = (PredicateEx<T>) fnAdapter.adaptFilterFn(filterFn);
-        return (RET) attach(new MapTransform<T, T>("filter", transform, t -> adaptedFn.test(t) ? t : null), fnAdapter);
+        PredicateEx<T> adaptedPred = (PredicateEx<T>) fnAdapter.adaptFilterFn(filterFn);
```
cangencer: wouldn't it be better to do the wrapping as part of the adaptFilterFn?

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/UserMetricsTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+
+public class UserMetricsTest extends JetTestSupport {
+
+    private static final JobConfig JOB_CONFIG_WITH_METRICS = new JobConfig().setStoreMetricsAfterJobCompletion(true);
+
+    @Test
+    public void filter() {
```
cangencer: can you add a test with a stream source with timestamps and windowing as well?

cangencer:
We should list out which functions and where user metrics can be applied to. It doesn't seem reasonable to me that it can be applied to any function there is in the pipeline API.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsOperatorUtil.java
jbartok: Sorry, but this doesn't make sense to me, our mental models diverge somewhere. We should talk.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/ComputeStageImplBase.java
jbartok: I have moved the metrics related wrapping inside the function adapter. Looks indeed much better now, but had to make a bit of a compromise to do it for filter (due to adapting a predicate but needing a function in the end). I think this "compromise" thing I'm mentioning is the same thing your "have to do it before" comment is about too. Take a look at `ComputeStageImpBase.InternalFunctionAdapter.filterPredicateToFn`.

jbartok:
> We should list out which functions and where user metrics can be applied to. It doesn't seem reasonable to me that it can be applied to any function there is in the pipeline API.

Most definitely we should. Even simple aggregation on a batch stage can have 6 different places where user metrics can be set (see `BatchStageImpl.aggregate.metricsProviderCandidates`) and I'm pretty sure that's not a pragmatic thing to do. Most likely the reasonable option is `AggregateOperation1.accumulateFn()` alone. 

I will go ahead and try to make a list for your approval before I implement any more cases.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MetricsContext.java
jbartok: What you said here makes sense to me, but let's say I add a method like `setCounterSource(String name, SupplierEx<Long> supplier)` to `MetricsContext` which would facilitate the user providing a lambda, how would I enforce thread safety? We do such things internally, but then we control the threads. Here we don't though... Have to think about it a bit more.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsOperatorUtil.java
cangencer: I mean if two lambdas provide metrics with same name, they would be overwriten?

jbartok: Metrics with the same name would be definitely overwritten, because eventually they will end up in `MetricsRegistryImplementation.registerInternal`. The overwrite would be logged but that's it. That being said, once I figure out how to wrap counters and use lambdas I'll take another look if I could improve something on my side (ie. above code being commented on).

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MetricsContext.java
jbartok: Have pushed **A** solution, which hides the `AtomicLong` and allows both for explicit setting of values and implicit provision of lambda sources, but it's not perfect (in particular I mean that I wrote a comment that the lambda has to be thread-safe, but we can't enforce it). Take a look when you can and we'll see if it's acceptable.

jbartok:
See TDD for the proposed list of user functions to support.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -203,6 +205,21 @@ public static void checkSerializable(Object object, String objectName) {
         }
     }
 
+    /** Returns a deep clone of an object by serializing and deserializing it (ser-de). */
+    public static <T> T serde(T object) {
```
cangencer: shouldn't be this constrained to `Serializable` ? How will it work with classes loaded from job class loader

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/ComputeStageImplBase.java
```diff
@@ -429,4 +433,178 @@ static void ensureJetEvents(@Nonnull ComputeStageImplBase stage, @Nonnull String
...
+
+    private static InternalFunctionAdapter extendFunctionAdapter(FunctionAdapter fnAdapter) {
+        if (fnAdapter instanceof InternalFunctionAdapter) {
+            return (InternalFunctionAdapter) fnAdapter;
+        } else {
+            return new InternalFunctionAdapter(fnAdapter);
+        }
+    }
+
+    private static final class InternalFunctionAdapter extends FunctionAdapter {
```
cangencer: what is internalFunctionAdapter? Name is not very descriptive.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsUtil.java
```diff
@@ -0,0 +1,401 @@
...
+            return new WrappedFunctionEx<>(functionEx,
+                    Collections.singletonList((ProvidesMetrics) metricsProviderCandidate));
+        } else {
+            return functionEx;
+        }
+    }
+
+    public static <T, U, R> BiFunctionEx<T, U, R> wrap(BiFunctionEx<T, U, R> biFunctionEx,
+                                                                    Object metricsProviderCandidate) {
+        if (metricsProviderCandidate instanceof ProvidesMetrics) {
```
cangencer: why do you need to wrap it here? I find the names very confusing, it's just about copying the metrics from the original function right?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsUtil.java
```diff
@@ -0,0 +1,401 @@
...
+    public static <T, U> BiPredicateEx<T, U> wrap(BiPredicateEx<T, U> biPredicateEx, Object metricsProviderCandidate) {
+        if (metricsProviderCandidate instanceof ProvidesMetrics) {
+            return new WrappedBiPredicateEx<>(biPredicateEx,
+                    Collections.singletonList((ProvidesMetrics) metricsProviderCandidate));
+        } else {
+            return biPredicateEx;
+        }
+    }
+
+    public static <T, R> FunctionEx<T, R> wrap(FunctionEx<T, R> functionEx, Object metricsProviderCandidate) {
```
cangencer: the candidate is only type Object, it can be constrained stronger

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsUtil.java
```diff
@@ -0,0 +1,401 @@
...
+            return (ProvidesMetrics) metricsProviderCandidate;
+        } else {
+            return DUMMY_METRICS_PROVIDER;
+        }
+    }
+
+    public static <T, U> BiPredicateEx<T, U> wrap(BiPredicateEx<T, U> biPredicateEx, Object metricsProviderCandidate) {
+        if (metricsProviderCandidate instanceof ProvidesMetrics) {
+            return new WrappedBiPredicateEx<>(biPredicateEx,
+                    Collections.singletonList((ProvidesMetrics) metricsProviderCandidate));
```
cangencer: you can move the singletonList logic to AbstractWrapper constructor

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsUtil.java
jbartok: I'm not sure I understand the "why here" question. Are you asking about one particular usage of this particular `wrap` method?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsUtil.java
jbartok: I'm not sure it can... This particular wrap for example has usages with both `FunctionEx` and `PredicateEx` candidates. 
`extends ProvidesMetrics` is also not a good constraint, because most of the time candidates don't extend it and no wrapping will take place. 
If I'm misunderstanding something pls. provide an example of what you meant.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
viliam-durina: This method is called when creating the DAG, not when running the job. It's called on the job submitter, and it has the classes on its classpath, so there should not be a problem

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/ProvidesMetrics.java
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.jet.core.metrics;
+
+/**
+ * Marker interface to be implemented by all classes that want to be able
```
jbartok: Comment from Viliam: `ProvidesMetrics` is not technically a "marker interface". Marker interface has no methods. `Serializable` is a marker interface, for example.

jbartok: True, but I think hiding the method would just make the code more awkward and not calling it a "marker interface" would just make it harder to understand/explain. I would not change anything. @cangencer pls. comment.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/TransformStatefulP.java
```diff
@@ -93,6 +96,11 @@ protected boolean tryProcess(int ordinal, @Nonnull Object item) {
         return flatMapper.tryProcess((T) item);
     }
 
+    @Override
+    public void registerMetrics(MetricsContext context) {
+        UserMetricsUtil.cast(statefulFlatMapFn).registerMetrics(context);
```
viliam-durina: Would be better if all functions in which we support user metrics would be registered here. We rely on the code in `Processors.mapStatefulP()` that registers metrics for both the `createFn` and `statefulFlatMapFn` in `statefulFlatMapFn`.

Also, we should support metrics in `onEvictFn`.

PR #1666, opened 2019-10-01
Add s3 to the index of source/sink connectors

gurbuzali:


cangencer:
It needs to be added to these places as well: 

https://docs.hazelcast.org/docs/jet/latest-dev/manual/#pipeline-source-sink
https://docs.hazelcast.org/docs/jet/latest-dev/manual/#overview-of-sources-and-sinks

PR #1671, opened 2019-10-02
Fix code sample

### LINKED ISSUE #1670: Fix sample in refman
The refman has a sample that teaches not to mutate an item after emitting it. The sample uses `mapUsingContext` to do rolling aggregation, which is wrong. It should be changed to use `mapStateful`.

mtopolnik:
Fixes #1670

### DIFF
--- a/hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/performance-considerations.adoc
+++ b/hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/performance-considerations.adoc
@@ -66,7 +66,7 @@ emitting is rolling aggregation. Consider this example where we use
 include::{javasource}/Serialization.java[tag=modify-emitted]
 ----
 
-<1>  You keep emitting the same list, changing it in-place
+<1>  Here you both modify and emit the same list
 
 The downstream stage receives the same object each time, but it's
 supposed to see it in the state as it was sent. In reality, the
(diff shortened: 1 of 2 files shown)


hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/performance-considerations.adoc
```diff
@@ -66,7 +66,7 @@ emitting is rolling aggregation. Consider this example where we use
 include::{javasource}/Serialization.java[tag=modify-emitted]
 ----
 
-<1>  You keep emitting the same list, changing it in-place
+<1>  Here you both modify and emit the same list
```
viliam-durina: The previous version was clearer to me. What about:
> You should not continue to modify the item that you emit.

mtopolnik: This is just the first sentence of the full explanation that continues below.

viliam-durina: What about:
> Here you emit an object that's your mutable state.

In "Here you both modify and emit the same list" - we don't _modify_ it here.

mtopolnik: I updated the wording again, you can review.

hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/performance-considerations.adoc
```diff
@@ -59,21 +59,24 @@ data, if the object is not thread-safe.
...
-The downstream stage receives the same object each time, but it's
-supposed to see it in the state as it was sent. In reality, the
-aggregating stage keeps changing the object concurrently while the
-downstream stage tries to observe it. In the best case it will crash
-with a `ConcurrentModificationException`, but it may also observe `null`
-or partially initialized objects in the list and a number of other bugs.
+<1> Modify the state object
+<2> Return the state object
+
+Jet keeps your state object passes it to each invocation of the mapping
```
viliam-durina: "Jet keeps your state object **and** passes it"

PR #1681, opened 2019-10-07
Introduce ability to add Jars in ZIP files as job resources

eminn:
The JAR files found inside the ZIP package unpacked and the classes are loaded with the Jet classloader.

I hesitated to rename `addJar` method to `addArchive` on `JobConfig` class since it was used a lot of places and became a vital part of the configuration workflow. Instead I added `addZip` method which is a copy of `addJar`, any feedback is welcomed on this manner.


hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -277,6 +277,50 @@ public JobConfig addJar(@Nonnull String path) {
         }
     }
 
+    /**
+     * Adds the ZIP file identified by the supplied URL to the list of archives
+     * that will be a part of the job's classpath while it's executing in the
+     * Jet cluster.
```
viliam-durina: From this description it wasn't clear to me what's the difference between jar and zip. We load all `jar` files inside the zip file, which was surprising to me.

When you are specifying classpath for `java` and you specify a zip file, afaik, it will be treated just like a jar file - java ignores the extension. Actually you can't specify zip-with-jars file using `java -cp`.

Also what's surprising that it doesn't matter whether you use `addJar` or `addZip`, only the actual file extension matters. If you add a file ending with `zip` using the `addJar` method because it has methods directly in it, no error will be reported and no classes will be loaded. Also, the extension checking is case sensitive even on windows. It would be better to handle the file according to which method was used and ignore the actual file extension. And I'd rename this method to `addJarsInZip`.

mtopolnik: I propose closing this PR and merging this work with my Python work, there's a lot of overlap and we have to solve these two use cases so they don't collide.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/ResourceConfig.java
```diff
@@ -75,12 +78,17 @@ public String getId() {
     }
 
     /**
-     * Whether this entry is an Jar archive or a single resource element.
+     * Whether this entry is an archive file or a single resource element.
      */
     public boolean isArchive() {
```
viliam-durina: This method should be removed, it's superseded by `ResourceType`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -204,25 +212,51 @@ long uploadJobResources(JobConfig jobConfig) {
...
+            ZipEntry zipEntry;
+            while ((zipEntry = zis.getNextEntry()) != null) {
+                if (zipEntry.isDirectory()) {
                     continue;
                 }
-                readStreamAndPutCompressedToMap(jarEntry.getName(), map, jis);
+                if (zipEntry.getName().endsWith(".jar")) {
+                    loadJarFromInputStream(map, zis);
+                }
+                readStreamAndPutCompressedToMap(zipEntry.getName(), map, zis);
```
viliam-durina: `else` should be here, otherwise we'll add resource to classpath with the jar file, but it will be empty because the `zis` is already read. But I would consider removing it altogether. We should only add the jar files from the zip, not all files. The packages of java products typically contains samples, javadocs, licenses, which are not needed and even should not be on the classpath, they take precious space in IMap at best.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -124,8 +126,8 @@
      * Prefix for internal IMaps which store snapshot data. Snapshot data for
      * one snapshot is stored in either of the following two maps:
      * <ul>
-     *     <li>{@code _jet.snapshot.<jobId>.0}
-     *     <li>{@code _jet.snapshot.<jobId>.1}
+     * <li>{@code _jet.snapshot.<jobId>.0}
+     * <li>{@code _jet.snapshot.<jobId>.1}
```
viliam-durina: Also here the indent was better when reading in plain text

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/ResourceType.java
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Represent a file of any type.
+     */
+    REGULAR_FILE,
+    /**
+     * Represents a JAR file.
+     */
+    JAR,
+    /**
+     * Represents a ZIP file contains JAR files.
```
viliam-durina: "contains" -> "containing" or "that contains"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -204,22 +217,38 @@ long uploadJobResources(JobConfig jobConfig) {
...
-     * Unzips the Jar archive and processes individual entries using
-     * {@link #readStreamAndPutCompressedToMap(String, Map, InputStream)}.
+     * Unzips the ZIP archive and processes JAR files
      */
-    private void loadJar(Map<String, byte[]> map, URL url) throws IOException {
-        try (JarInputStream jis = new JarInputStream(new BufferedInputStream(url.openStream()))) {
-            JarEntry jarEntry;
-            while ((jarEntry = jis.getNextJarEntry()) != null) {
-                if (jarEntry.isDirectory()) {
+    private void loadZip(Map<String, byte[]> map, URL url) throws IOException {
```
viliam-durina: Method should be `loadJarsInZip`

PR #1688, opened 2019-10-11
Add exception logging to JUL logging.properties

mtopolnik:
Our default logging.properties file inside the Jet JAR sets up the log line format that doesn't include the exception stacktrace. This PR add that. Hopefully we can squeeze in this small detail to the 3.2 release and make life of the users of our ZIP file easier.


cangencer:
PR is against master, but has 3.2 tag, if it should be in 3.2 it needs to go into 3.2 branch.

PR #1692, opened 2019-10-15
User metrics

jbartok:


jbartok:
@viliam-durina pls. take a look. Keep in mind that this is just a sketch/POC, don't go about writing javadoc and such, I'll do that, I'll detail it and add what's missing if you agree with the general idea.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
```diff
@@ -160,6 +163,8 @@ public ProcessorTasklet(@Nonnull Context context,
         if (probeBuilder != null) {
             registerMetrics(instreams, probeBuilder);
         }
+
+        this.userMetricsContext = new UserMetricsContext(probeBuilder, this, ProbeLevel.INFO, ProbeUnit.MS);
```
viliam-durina: `ProbeUnit.MS`? We don't know.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
```diff
@@ -26,6 +26,8 @@
 import com.hazelcast.jet.core.Processor.Context;
 import com.hazelcast.jet.core.Watermark;
 import com.hazelcast.jet.core.metrics.MetricTags;
+import com.hazelcast.jet.impl.metrics.UserMetricsContext;
+import com.hazelcast.jet.impl.metrics.UserMetricsImp;
```
viliam-durina: Should be `UserMetricsImpl`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsContext.java
```diff
@@ -0,0 +1,174 @@
...
+public class UserMetricsContext {
+
+    private final Function<String, Counter> counterSupplier;
+
+    private String onlyMetric;
+    private Counter onlyCounter;
+    private Map<String, Counter> metrics;
+
+    public UserMetricsContext(ProbeBuilder probeBuilder, ProcessorTasklet source, ProbeLevel level, ProbeUnit unit) {
+        this.counterSupplier = metric -> {
```
viliam-durina: Why do we need a supplier? It's used only once.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsContext.java
```diff
@@ -0,0 +1,174 @@
...
+            if (probeBuilder == null) {
```
viliam-durina: When can the probe builder be null? Why do we need `UnregisteredCounter `?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsContext.java
```diff
@@ -0,0 +1,174 @@
...
+                    onlyMetric = null;
+                    onlyCounter = null;
+                }
+                metrics.put(metric, res);
+            }
+        }
+        return res;
+    }
+
+    private static final class RegisteredCounter implements Counter {
```
viliam-durina: `UnregisteredCounter` isn't thread-safe, this impl is. Neither should be, we should use `lazySet` to set the value for performance.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsImp.java
```diff
@@ -0,0 +1,37 @@
...
+    private UserMetricsImp() {
+    }
+
+    public static void set(UserMetricsContext context) {
+        CONTEXT.set(context);
+    }
+
+    public static Counter get(String name) {
+        UserMetricsContext context = CONTEXT.get();
+        return context.get(name);
```
viliam-durina: We should throw some more informative exc if `context` is null, like "This thread is not a processor thread, metrics can be used only on processor threads"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsContext.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        @Override
+        public void decrement() {
+            Util.lazyAdd(value, -1L);
+        }
+
+        @Override
+        public void decrement(long decrement) {
+            Util.lazyAdd(value, -1L * decrement);
```
cangencer: I think can be `-decrement`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
```diff
@@ -232,10 +237,17 @@ public void init() {
...
+
+        if (!processor.isCooperative()) {
+            UserMetricsImp.set(userMetricsContext);
+        }
     }
 
     @Override @Nonnull
     public ProgressState call() {
+        if (processor.isCooperative()) {
+            UserMetricsImp.set(userMetricsContext);
```
cangencer: we should be able to avoid this by having a reference to the TL object, and updating a field on it.

viliam-durina:
Other requested changes:

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsContext.java
jbartok: Will inline.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsContext.java
jbartok: As discussed this case covers the situation when metric are disabled completely. Will write a test for it.

jbartok:
@viliam-durina the null traverser appearing can be observed in `UserMetrics_FilterTest.filterUsingContextAsync_stream` on the `user-metrics` branch. I'm checking if it also happens on this branch or not. 

Unfortunately this check showed a much bigger problem, our current approach doesn't work at all with these "use context async" operations, when the user code gets executed on the ForkJoinPool's treads. 

Looking for a way to set the metrics context on those threads.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
```diff
@@ -270,6 +262,7 @@ public void run() {
                 tracker.executionTracker.exception(new JetException("Exception in " + t + ": " + e, e));
             } finally {
                 blockingWorkerCount.decrementAndGet();
+                UserMetricsImpl.getContainer().setContext(null);
```
cangencer: the logic here is racy. if the setContext() throws then count will get decretmented without having been incremented.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsContext.java
```diff
@@ -0,0 +1,208 @@
...
+            VOLATILE_VALUE_UPDATER.lazySet(this, value - decrement);
+        }
+
+        @Override
+        public long get() {
+            return value;
+        }
+    }
+
+    private static final class NonRegisteredMetric implements Metric {
```
cangencer: why do you need this? It would be simpler just to use the same object even if metrics are disabled. Seems like unnecessary over-optimization.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsContext.java
```diff
@@ -0,0 +1,208 @@
...
+            case MS:
+                return ProbeUnit.MS;
+            case BYTES:
+                return ProbeUnit.BYTES;
+            default:
+                throw new RuntimeException("Unhandled metrics unit " + unit);
+        }
+    }
+
+    private interface Metric extends Counter, Gauge {
```
cangencer: this interface should be removed

jbartok: Why is it a problem? It's private and simplifies the code in this class a great deal.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/UserMetrics.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private UserMetrics() {
+    }
+
+    /**
+     * Returns a counter type handler for manipulating the metric with
+     * the specified name. Is equivalent with a gauge of the unit
+     * {@link Unit#COUNT}.
+     */
+    public static Counter getCounter(String name) {
```
cangencer: let's merge these as one thing, and call the method `metric` (not getMetric)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsContext.java
cangencer: I thought the class impl is simple enough we don't need two different interfaces for it.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/UserMetrics.java
```diff
@@ -0,0 +1,58 @@
...
+ * they do in pipeline user code they write.
+ * <p>
+ * This class provides methods for creating various types of handlers
+ * for setting up and manipulating the values of such metrics. There is
+ * a version which makes it possible to count things (see {@link Counter}),
+ * or one that is more generic and can be used to set all types of values,
+ * including counters (see {@link Gauge}).
+ */
+@SuppressWarnings("WeakerAccess")
+public final class UserMetrics {
```
cangencer: I'm not sure about the name "UserMetrics", we should call it something more descriptive. Just `Metrics` maybe?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsContext.java
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        @Override
+        public void set(long newValue) {
+            VOLATILE_VALUE_UPDATER.lazySet(this, newValue);
+        }
+
+        @Override
+        public void inc() {
+            VOLATILE_VALUE_UPDATER.lazySet(this, value + 1);
```
cangencer: this approach means that counter can only be incremented by a single thread (like in mapUsingContextAsync). This should be documented, in this case.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsImpl.java
```diff
@@ -0,0 +1,71 @@
...
+
+        Container() {
+        }
+
+        public UserMetricsContext getContext() {
+            return context;
+        }
+
+        public void setContext(Tasklet tasklet) {
+            if (tasklet instanceof ProcessorTasklet) {
```
cangencer: can we avoid hardcoding the instance check here somehow? i.e. a new method on tasklet?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsImpl.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public static Gauge getGauge(String name, Unit unit) {
+        return getContext().getGauge(name, unit);
+    }
+
+    private static UserMetricsContext getContext() {
+        Container container = CONTEXT.get();
+        UserMetricsContext context = container.getContext();
+        if (context == null) {
+            throw new RuntimeException("User metrics are accessible only from internal worker threads");
```
cangencer: the exception needs to be more descriptive.

hazelcast-jet-reference-manual/src/main/java/LogDebug.java
```diff
@@ -66,18 +75,69 @@ static void s3() {
...
+            //tag::s7[]
+            .filterUsingContextAsync(
+                ContextFactory.withCreateFn(i -> 0L),
+                (ctx, l) -> {
+                    Counter dropped = UserMetrics.getCounter("dropped");
+                    return CompletableFuture.supplyAsync(
+                        () -> {
+                            boolean pass = l % 2L == ctx;
+                            if (!pass) {
+                                dropped.inc();
```
cangencer: this won't work because of lazySet

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsContext.java
jbartok: Even if we document it, we still won't have a workaround for the ForkJoinPool case... Should we maybe give up on `lazySet`?

cangencer: We could have a different thread-safe counter, perhaps.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/UserMetricsImpl.java
jbartok: I agree in principle, have given it another try but I don't think it's much good... Just don't seem to be able to formulate a message which is meaningful from the user's perspective. Suggestions welcome.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsImpl.java
```diff
@@ -16,55 +16,51 @@
...
-        public UserMetricsContext getContext() {
+        public MetricsContext getContext() {
             return context;
         }
 
-        public void setContext(Tasklet tasklet) {
-            if (tasklet instanceof ProcessorTasklet) {
-                this.context = ((ProcessorTasklet) tasklet).getUserMetricsContext();
-            }
+        public void setContext(@Nullable Tasklet tasklet) {
```
cangencer: wouldn't it be better to pass the context directly here instead from the ExecutionService?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/Metrics.java
```diff
@@ -0,0 +1,93 @@
...
+     *           );
+     *       }
+     *   )
+     *   .drainTo(...);
+     * }
+     * </pre>
+     *
+     *
+     */
+    public static Metric metric(String name, Unit unit, boolean threadSafe) {
```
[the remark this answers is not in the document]
jbartok: :) Exactly what I did initially, changing it back!

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsImpl.java
jbartok: Not sure I understand, pls. rephrase.

cangencer: Set the call to `MetricsImpl.container().setContext(t.metricsContext())` rather than passing `Tasklet`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsContext.java
```diff
@@ -0,0 +1,209 @@
...
+                    onlyUnit = null;
+                }
+                metrics.put(name, res);
+            }
+        }
+        return res;
+    }
+
+    public void collectMetrics(MetricTagger tagger, MetricsCollectionContext context) {
+        if (onlyMetric != null) {
```
cangencer: we should also maybe add a tag specific for user metrics?

jbartok: Ok, added.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/MetricsTest.java
```diff
@@ -0,0 +1,279 @@
...
+    public void usingContextAsync() {
+        pipeline.drawFrom(TestSources.items(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11))
+                .addTimestamps(i -> i, 0L)
+                .window(WindowDefinition.tumbling(3))
+                .aggregate(summingLong((ToLongFunctionEx<Integer>) Integer::longValue))
+                .map(WindowResult::result)
+                .filterUsingServiceAsync(
+                        ServiceFactory.withCreateFn(i -> 0L),
+                        (ctx, l) -> {
+                            Metric dropped = Metrics.metric("dropped");
```
viliam-durina: We should use the thread-safe version here. We could actually stress-test it a bit by having lot more items and check that the total matches.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsContext.java
```diff
@@ -0,0 +1,215 @@
...
+
+                    Metric metric = metricSupplier.apply(name, unit);
+                    metrics.put(name, metric);
+                    return metric;
+                }
+            }
+        } else { //multiple metrics already defined
+            Metric metric = metrics.get(name);
+            if (metric == null) { //requested metric not yet defined
+                metric = metricSupplier.apply(name, unit);
```
viliam-durina: This branch is not covered by tests

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
```diff
@@ -163,6 +165,8 @@ public ProcessorTasklet(@Nonnull Context context,
         waitForAllBarriers = ssContext.processingGuarantee() == ProcessingGuarantee.EXACTLY_ONCE;
 
         watermarkCoalescer = WatermarkCoalescer.create(instreams.size());
+
+        metricsContext = new MetricsContext();
```
cangencer: can be moved to field initializer

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
```diff
@@ -249,12 +240,13 @@ private BlockingWorker(TaskletTracker tracker, CountDownLatch startedLatch) {
...
         public void run() {
             final ClassLoader clBackup = currentThread().getContextClassLoader();
-            final Tasklet t = tracker.tasklet;
+            Tasklet t = tracker.tasklet;
             currentThread().setContextClassLoader(tracker.jobClassLoader);
             IdleStrategy idlerLocal = idlerNonCooperative;
 
             try {
                 blockingWorkerCount.incrementAndGet();
+                MetricsImpl.container().setContext(t.getMetricsContext());
```
cangencer: minor, but you can assign MetricsImpl.container() to a variable instead of calling it twice

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
```diff
@@ -274,6 +266,7 @@ public void run() {
                 tracker.executionTracker.exception(new JetException("Exception in " + t + ": " + e, e));
             } finally {
                 blockingWorkerCount.decrementAndGet();
+                MetricsImpl.container().setContext(null);
```
cangencer: I would prefer to have a `clear()` method here, rather than setting it to null.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsContext.java
```diff
@@ -0,0 +1,215 @@
...
+
+public class MetricsContext {
+
+    private String onlyName;
+    private Metric onlyMetric;
+
+    private Map<String, Metric> metrics;
+
+    Metric metric(String name, Unit unit) {
+        return metric(name, unit, SimpleMetric::new);
```
cangencer: you can assign the lambdas to constant fields to avoid lambda creation each time this is invoked

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/Metric.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Decrements the current value by 1.
+     */
+    void dec();
+
+    /**
+     * Decrements the current value by the specified amount.
+     */
+    void subtract(long amount);
```
cangencer: let's call this `dec(amount)` ?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsContext.java
```diff
@@ -0,0 +1,215 @@
...
+            case MS:
+                return ProbeUnit.MS;
+            case BYTES:
+                return ProbeUnit.BYTES;
+            default:
+                throw new RuntimeException("Unhandled metrics unit " + unit);
+        }
+    }
+
+    private static final class SimpleMetric implements Metric {
```
cangencer: What's simple about it? `SingleWriterMetric` is a better name

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsContext.java
```diff
@@ -0,0 +1,215 @@
...
+        public long get() {
+            return value;
+        }
+    }
+
+    private static final class ThreadSafeMetric implements Metric {
+
+        private final String name;
+        private final Unit unit;
+        private AtomicLong value = new AtomicLong();
```
cangencer: you can also use the AtomicFieldUpdater approach here to reduce garbage

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
jbartok: Ok, I did that, but then it seems to me that I shouldn't any longer allow null `MetricsContext` param to be passed to `setContext`.  But if I do that I have to check for null, before calling `setContext` and that will make the caller code uglier. What do you think?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/Metrics.java
```diff
@@ -0,0 +1,106 @@
...
+     *   .filterUsingServiceAsync(
+     *       ServiceFactory.withCreateFn(i -> "foo"),
+     *       (ctx, item) -> {
+     *           Metric dropped = Metrics.threadSafeMetric("dropCount", Unit.COUNT);
+     *           return CompletableFuture.supplyAsync(
+     *               () -> {
+     *                   // the callback runs on another thread
+     *                   boolean pass = item % 2L == 0;
+     *                   if (!pass) {
+     *                       dropped.inc();
```
viliam-durina: Old name

jbartok: I will call you Viliam Eagle-Eye from now on!

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsContext.java
jbartok: Hmm... I fixed the coverage, but there is a problem... So I've added a new metric to `MetricsTest.counter`, called "last", like this:

                .drawFrom(TestSources.items(0L, 1L, 2L, 3L, 4L))
                .filter(l -> {
                    boolean pass = l % 2 == 0;

                    if (!pass) {
                        Metrics.metric("dropped").increment();
                    }
                    Metrics.metric("total").increment();
                    System.out.println("l = " + l); //todo: remove
                    Metrics.metric("last").set(l);

                    return pass;
                })

I was naively expecting a value of 4 or at least one of the input values... But it's far from it, the result is 10, because each input goes to a different processor and the various "last" values from the multiple processors get summed up.

This behaviour is ok, when we count things and use the increment and decrement operations, but with this "set" operation and other such intentions summing up the values per processor is not really what we need... 

What do you think we should do? Maybe implement some sort of option for metric where we specify how they should be aggregated?

jbartok:
I think it's ready for final reviews.

PR #1700, opened 2019-10-22
More efficient empty traverser

viliam-durina:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/Traversers.java
```diff
@@ -38,7 +39,44 @@
  */
 public final class Traversers {
 
-    private static final Traverser<Object> EMPTY_TRAVERSER = () -> null;
+    private static final Traverser<Object> EMPTY_TRAVERSER = new Traverser<Object>() {
```
cangencer: wouldn't it be better to create a named class here?

PR #1708, opened 2019-10-30
Increased Metrics test coverage

olukas:


hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/MetricsTest.java
```diff
@@ -205,6 +210,193 @@ public void usingContextAsync() {
         job.join();
     }
 
+    @Test
+    public void availableDuringJobExecution() {
```
jbartok: From the name of the test it seems like you would like to assert the metrics while the job is running. If that's so, this test is really undeterministic, you will probably catch the job running at most some of the time. I think you might want to use something like `TestProcessors.NoOutputSourceP` and its `proceedLatch` to make it deterministic.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/MetricsTest.java
```diff
@@ -205,6 +210,193 @@ public void usingContextAsync() {
...
+        assertTrueEventually(() -> {
+            assertMetricsProduced(job, "total", generatedItems);
+        });
+    }
+
+    @Test
+    public void when_metricsDisabled_then_unavailableDuringJobExecution() {
+        int generatedItems = 1000;
+
+        pipeline.drawFrom(TestSources.itemStream(1_000))
```
jbartok: Same undeterminism as with `availableDuringJobExecution`. Also you might want to use `assertTrueAllTheTime` to check that there aren't any metrics, ever.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/MetricsTest.java
olukas: I think it is deterministic - `TestSources.itemStream` creates `StreamStage` which means that job will be still running during metrics are asserted. `filter` stage stop to emit items when it reaches `generatedItems` amount. It means that certain number can be checked in `assertTrueEventually` part. I believe this deterministic and sufficient, WDYT?

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/MetricsTest.java
olukas: Ok, I will use `assertTrueAllTheTime` for 3 seconds instead of current `assertTrue`.

olukas:
@jbartok I used `assertTrueAllTheTime` as you suggested + added test for using Jet predefined metric in custom metric. Please review it.

PR #1710, opened 2019-10-30
Fixed MapSourceAndSinks mapWithUpdating example

olukas:
It failed because there was `Map<Integer, Integer>` which was used as source and sink but entries `<Integer, String>` tried to be written into it.

### DIFF
--- a/examples/imdg-connectors/src/main/java/com/hazelcast/jet/examples/imdg/MapSourceAndSinks.java
+++ b/examples/imdg-connectors/src/main/java/com/hazelcast/jet/examples/imdg/MapSourceAndSinks.java
@@ -65,17 +65,22 @@ public class MapSourceAndSinks {
     }
 
     /**
-     * This will take the contents of source map, converts values to the string and
-     * suffixes the value with {@code odd} if the key is odd and with {@code event} if the key is even.
+     * This will take the contents of a map and suffixes the value with
+     * {@code odd} if the key is odd and with {@code even} if the key is even.
      */
-    private static Pipeline mapWithUpdating(String sourceMapName, String sinkMapName) {
+    private static Pipeline mapWithUpdating(String mapName) {
         Pipeline pipeline = Pipeline.create();
 
-        pipeline.drawFrom(Sources.<Integer, Integer>map(sourceMapName))
-                .map(e -> entry(e.getKey(), String.valueOf(e.getValue())))
+        // Note that we use the same map for source and sink. The map reader
+        // requires the map to not change while reading or it might produce
+        // incorrect results. But our job does not add or remove any keys in
+        // the map, it only changes the values, so we can use it. Note that
+        // you can also use IMap.executeOnEntries() for this simple case,
+        // but Jet pipeline has more features.
+        pipeline.drawFrom(Sources.<Integer, String>map(mapName))
                 .drainTo(
                         Sinks.mapWithUpdating(
-                                sinkMapName,
+                                mapName,
                                 (oldValue, item) ->
                                         item.getKey() % 2 == 0
                                                 ? oldValue + "-even"
@@ -134,7 +139,7 @@ public class MapSourceAndSinks {
         try {
 
             System.out.println("----------Map Source and Sink ----------------");
-            // insert sequence 1..10 into map as (1,1) , (2,2) .... ( 9,9)
+            // insert sequence 0..9 into map as (0,0) , (1,1) .... ( 9,9)
             prepareSampleInput(jet, MAP_SOURCE);
             // execute the pipeline
             jet.newJob(mapSourceAndSink(MAP_SOURCE, MAP_SINK)).join();
@@ -144,7 +149,7 @@ public class MapSourceAndSinks {
 
 
             System.out.println("--------------Map with Merging----------------");
-            // insert sequence 1..10 into map as (1,1) , (2,2) .... ( 9,9)
+            // insert sequence 0..9 into map as (0,0) , (1,1) .... ( 9,9)
             prepareSampleInput(jet, MAP_WITH_MERGING_SOURCE);
             // execute the pipeline
             jet.newJob(mapWithMerging(MAP_WITH_MERGING_SOURCE, MAP_WITH_MERGING_SINK)).join();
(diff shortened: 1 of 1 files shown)


olukas:
@viliam-durina I rewrote it to use different maps for source and sink. I also set value in source to value `notUsed` and initialize values in sink map to demostrate that is uses `Sinks.mapWithUpdating`.

hhromic:
> We read and write from and to the same map, that was the issue. We must not write to the map we're reading from, that works only because the map has few entries.

Sorry to hijack this PR, but I'm precisely trying to do this.

So it is **not** recommended to do the following (pseudo-code)?
```
drawFrom(Sources.map(MAP_NAME))
drainTo(Sinks.mapWithUpdating(MAP_NAME))
```
I thought that should work fine as long as we don't change the map keys, i.e. add or remove.

What would be the recommended way to update map entries then using a Jet pipeline?
Should I just use plain IMDG `executeOnKeys()` ?  Thanks.

viliam-durina:
@hhromic Your case should be OK, changing just the value isn't a structural map mutation, but plain `executeOnEntries()` might be even better for you, you can even delete entries with it. Just make sure that you return `null` in the entry processor because otherwise you'll be sending all the keys to the client and they can be too many.

We wanted to avoid reading and writing the same map in the sample because people might not realize this issue. The job might even not fail if you happen to actually add/remove keys, but produce wrong results.

hhromic:
Thanks for the feedback, much appreciated!

Given that we don't plan to mutate the map structurally, thanks for confirming it should work fine for now. Anyway as you suggested I'll investigate more on using an `EntryProcessor` directly (no need for Jet pipeline) to see if it also gives more performance for our case.

> We wanted to avoid reading and writing the same map in the sample because people might not realize this issue. The job might even not fail if you happen to actually add/remove keys, but produce wrong results.

Absolutely agree with that, clear documentation and examples are most important.

Just to confirm, is the following an accurate explanation of the issue?
> When streaming map-keys using `Sources.map()` you are streaming a *view* at the time of the source creation. If the map is modified after creating the source, you can expect inconsistencies because the source might be giving you keys that no longer exist, or if you add keys, the source will not stream these keys that were recently added. Therefore leading to inconsistent results.

viliam-durina:
@hhromic Regarding that explanation, where is it from? It's totally incorrect. Correct explanation is in the [javadoc](https://github.com/hazelcast/hazelcast-jet/blob/7850954ec8d943218f4868152621d66a0b9e703b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/Sources.java#L157-L162). If it was the way your text is relating it would be ideal. Even if the map is just migrated (after member addition, for example), we can emit some items duplicately and completely miss other.

examples/imdg-connectors/src/main/java/com/hazelcast/jet/examples/imdg/MapSourceAndSinks.java
```diff
@@ -65,17 +65,20 @@ private static Pipeline mapSourceAndSink(String sourceMapName, String sinkMapNam
     }
 
     /**
-     * This will take the contents of source map, converts values to the string and
-     * suffixes the value with {@code odd} if the key is odd and with {@code event} if the key is even.
+     * This will take the contents of source map and suffixes the value with {@code odd} if the key is odd and with
```
viliam-durina: - We wrap javadoc to 72 characters net. 
- We can also say "This will take the contents of **a** map", i.e. not use "source", because there's no source and sink map, just one map.

examples/imdg-connectors/src/main/java/com/hazelcast/jet/examples/imdg/MapSourceAndSinks.java
```diff
@@ -65,17 +65,20 @@ private static Pipeline mapSourceAndSink(String sourceMapName, String sinkMapNam
...
+     * {@code even} if the key is even.
      */
-    private static Pipeline mapWithUpdating(String sourceMapName, String sinkMapName) {
+    private static Pipeline mapWithUpdating(String sourceSinkMapName) {
```
viliam-durina: Also this could be just `mapName`

examples/imdg-connectors/src/main/java/com/hazelcast/jet/examples/imdg/MapSourceAndSinks.java
```diff
@@ -154,10 +157,10 @@ private void go() {
 
 
             System.out.println("------------Map with Updating ----------------");
-            // insert sequence 1..10 into map as (1,1) , (2,2) .... ( 9,9)
-            prepareSampleInput(jet, MAP_WITH_UPDATING_SOURCE_SINK);
+            // insert sequence 1..10 into map as (1,"1") , (2,"2") .... ( 9,"9")
```
viliam-durina: We start with `(0, "0")`

examples/imdg-connectors/src/main/java/com/hazelcast/jet/examples/imdg/MapSourceAndSinks.java
```diff
@@ -200,6 +203,17 @@ private static void prepareSampleInput(JetInstance instance, String sourceMapNam
         }
     }
 
+    /**
+     * Inserts the sequence 1..ITEM_COUNT into {@code sourceMapName} as (1,"1") , (2,"2") ....(ITEM_COUNT-1,
```
viliam-durina: again

hhromic:
> @hhromic Regarding that explanation, where is it from? It's totally incorrect. Correct explanation is in the [javadoc](https://github.com/hazelcast/hazelcast-jet/blob/7850954ec8d943218f4868152621d66a0b9e703b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/Sources.java#L157-L162). If it was the way your text is relating it would be ideal. Even if the map is just migrated (after member addition, for example), we can emit some items duplicately and completely miss other.

Apologies, my bad as I used bad wording. That "explanation" was my own personal understanding and wanted to confirm if I was correct. I see now that it wasn't and I was thinking (wrongly) on an *ideal case* instead of how it actually works.

I totally missed the javadoc, my bad again, thanks for reminding me to check it out. I mostly follow the Jet Reference Manual and ocassionally go to the javadoc for the fine-details.

Speaking of which, in the [reference manual itself](https://docs.hazelcast.org/docs/jet/latest/manual/#connector-imdg) the first example depicts usage of the same map as both Source and Sink. While this is not the same as using a `mapWithUpdating` Sink, I now realise I was doing wrong assumptions about using the same map for source/sink in *any* context.

I appreciate the help and thanks again for the patience. I won't be polluting this PR anymore.

viliam-durina:
@hhromic Thanks for pointing to the refman, I fixed that (#1728).

We also have this old issue: hazelcast/hazelcast#11106. It's a tricky one to fix. Currently we read the map in batches and keep the index into the internal hashtable to read the next batch from. But if the map is rehashed (after adding some number of entries or after migration), entries can be in very different order internally. Fixing that requires using different type of internal data structure...

PR #1715, opened 2019-11-04
Populate metrics based on job & snapshotting statistics

jbartok:
The aim is to extend our built-in metrics with useful statistical information about jobs, executions and snapshotting.


jbartok:
According to @viliam-durina it doesn't make sense for me to add the snapshot statistics metrics right now because he has a pending PR (https://github.com/hazelcast/hazelcast-jet/pull/1438) which will radically change how snapshots work and I would have to redo the whole thing after that's merged in. So I will just move on to other things until that PR is merged in.

jbartok:
Ready for review. There will be two follow-ups, but those will have their own PRs:
- snapshotting statistics metrics
- optimisation to enable us to enable `JobConfig.storeMetricsAfterJobCompletion` by default

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetService.java
```diff
@@ -104,6 +104,7 @@ public void init(NodeEngine engine, Properties hzProperties) {
         jobRepository = new JobRepository(jetInstance);
         jobExecutionService = new JobExecutionService(nodeEngine, taskletExecutionService, jobRepository);
         jobCoordinationService = createJobCoordinationService();
+        nodeEngine.getMetricsRegistry().registerDynamicMetricsProvider(jobCoordinationService);
```
cangencer: can we group the registrations together(with jobExecutionService)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -998,4 +1012,53 @@ NodeEngineImpl nodeEngine() {
...
+            executionStarted.incrementAndGet();
+        }
+
+        void executionTerminated() {
+            executionTerminated.incrementAndGet();
+        }
+
+        @Override
+        public void provideDynamicMetrics(MetricDescriptor descriptor, MetricsCollectionContext context) {
+            context.collect(descriptor, "jobs.submitted", ProbeLevel.INFO, ProbeUnit.COUNT,
```
cangencer: these are not dynamic metrics. You can just annotate them with `@Probe` and then collect the object directly

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterJobContext.java
```diff
@@ -681,20 +684,63 @@ void finalizeJob(@Nullable Throwable failure) {
...
         }
-        logger.severe(String.format("Execution of %s failed after %,d ms", mc.jobIdString(), elapsed), failure);
+        logger.severe(formatExecutionSummary("failed"), failure);
         return false;
     }
 
+    private String formatExecutionSummary(String conclusion) {
+        long executionEndTime = System.currentTimeMillis();
+        StringBuilder sb = new StringBuilder(String.format("Running %s %s", mc.jobIdString(), conclusion));
+        sb.append("\n\t").append("Start time: ").append(Util.toLocalDateTime(executionStartTime));
```
cangencer: is this start time or submission time

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ExecutionContext.java
```diff
@@ -61,6 +64,8 @@
     private final Set<Address> participants;
     private final Object executionLock = new Object();
     private final ILogger logger;
+    private final AtomicLong startTime = new AtomicLong();
```
cangencer: how is this different than the other start time above?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/MasterJobContext.java
jbartok: It's set initially on `MasterJobContext` creation, which is the moment of submission, but it's also updated in `MasterJobContext.tryStartJob`, which makes it actual start time. When it gets printed on execution summary it's definitely START time. 

This is how it was before too, I just changed it from nano to millis time. I think we could remove the initialisation on submission.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
jbartok: Ok, done in next push.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetService.java
jbartok: No longer needed, `JobCoordinationService` will not be a dynamic metrics provider.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ExecutionContext.java
jbartok: You are comparing this agains `MasterJobContext.executionStartTime`, right? Then, assuming I'm not grossly misunderstanding something, the difference is that `ExecutionContext` has one instance for each cluster member so it's suitable for sourcing metrics (which are per cluster member) and that's what we do with the `startTime` from it. 

On the other hand `MasterJobContext.executionStartTime` exists only on the master node so we couldn't use it for the "execution start time" metric.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -998,4 +1016,39 @@ NodeEngineImpl nodeEngine() {
...
         assert IS_JOB_COORDINATOR_THREAD.get() : "not on coordinator thread";
     }
+
+    private static class Stats {
+
+        @Probe(name = "jobs.submitted")
+        private final AtomicInteger jobSubmitted = new AtomicInteger();
+        @Probe(name = "jobs.completed_successfully")
+        private final AtomicInteger jobCompletedSuccessfully = new AtomicInteger();
+        @Probe(name = "jobs.completed_with_failure")
```
cangencer: looks like we don't use consistent naming with metrics. See `MetricNames`

jbartok: Fixed.

cangencer:
can you add the milestone & labels?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/OutboxImpl.java
```diff
@@ -56,6 +57,7 @@
     private Object unfinishedSnapshotKey;
     private Object unfinishedSnapshotValue;
     private final AtomicLong lastForwardedWm = new AtomicLong(Long.MIN_VALUE);
+    private final AtomicLong snapshotSize = new AtomicLong(0);
```
cangencer: this metric doesn't make much sense to me. It seems to be always incremented, and not reset between snapshots, so how would it be useful?

jbartok: My thinking at the time was that it would give an estimate about the total amount of snapshot data written out by a `Processor` during the lifetime of a job. But now I realise that this is not data which accumulates, but data that continuously overwrites itself (ie. same keys all the time probably), so indeed, not useful...

jbartok: Have removed these changes.

jbartok:
Left to do:
* add these metrics into the RefMan
* clean/finish-up TDD

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/StoreSnapshotTasklet.java
```diff
@@ -165,4 +184,27 @@ public String toString() {
...
         DONE
     }
+
+    private static class Metrics implements DynamicMetricsProvider {
+
+        private long bytes;
+        private long keys;
+        private long chunks;
+
+        synchronized void set(long bytes, long keys, long chunks) {
```
cangencer: why use `synchronized` here?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -968,4 +986,39 @@ NodeEngineImpl nodeEngine() {
...
+        @Probe(name = MetricNames.JOB_EXECUTIONS_TERMINATED)
+        private final AtomicInteger executionTerminated = new AtomicInteger();
+
+        void jobSubmitted() {
+            jobSubmitted.incrementAndGet();
+        }
+
+        void jobCompleted(Throwable error) {
+            if (error == null) {
+                jobCompletedSuccessfully.incrementAndGet();
```
cangencer: it's better to always use `lazySet` when incrementing, see `Util.lazyIncrement`

cangencer: ok, take it back as it can be incremented from multiple threads

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/StoreSnapshotTasklet.java
jbartok: The intention is to update/query the 3 values in an atomic fashion, ie. never publish metrics in which for example the byte count is from one snapshot and the keys from another. But not sure if these benefits outweigh the costs (costs which in turn shouldn't be high, since this is not really a contended lock).

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/StoreSnapshotTasklet.java
```diff
@@ -165,4 +183,24 @@ public String toString() {
...
+        private long keys;
+
+        synchronized void set(long bytes, long keys) {
+            this.bytes = bytes;
+            this.keys = keys;
+        }
+
+        @Override
+        public synchronized void provideDynamicMetrics(MetricDescriptor descriptor, MetricsCollectionContext context) {
+            if (bytes > 0) {
```
cangencer: I don't think we need the conditional logic here. It'll be confusing if metric is sometimes missing and sometimes not.

jbartok: Ok, I can remove it, it's just that for most vertices there are no snapshots, ever, so there are a lot of zeroes we publish (and I have not found a better way to tell about a vertex if that's the case or not). Anyways, will remove.

jbartok:
As far as I'm concerned this is ready for final review/merge in (barring the above outstanding discussion on using synchronization).

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/StoreSnapshotTasklet.java
```diff
@@ -165,4 +183,22 @@ public String toString() {
...
+
+        synchronized void set(long bytes, long keys) {
+            this.bytes = bytes;
+            this.keys = keys;
+        }
+
+        @Override
+        public synchronized void provideDynamicMetrics(MetricDescriptor descriptor, MetricsCollectionContext context) {
+            context.collect(descriptor, SNAPSHOT_BYTES, ProbeLevel.INFO, ProbeUnit.COUNT, bytes);
+            context.collect(descriptor, SNAPSHOT_KEYS, ProbeLevel.INFO, ProbeUnit.COUNT, keys);
```
viliam-durina: Why use `synchronized` and not just `volatile`? To avoid observing keys updated, but not bytes? This problem is all over the metrics, I wouldn't be bothered by that.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -239,11 +239,11 @@ public void write(int b) {
     }
 
     public static String jobNameAndExecutionId(String jobName, long executionId) {
-        return "job '" + jobName + "', execution " + idToString(executionId);
+        return "job '" + jobName + "', execution '" + idToString(executionId) + "'";
```
viliam-durina: We put the name to quotes, but not ID. As in code, strings are quoted, not numbers.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -239,11 +239,11 @@ public void write(int b) {
...
     }
 
     public static String jobIdAndExecutionId(long jobId, long executionId) {
-        return "job " + idToString(jobId) + ", execution " + idToString(executionId);
+        return "job " + idToString(jobId) + ", execution '" + idToString(executionId) + "'";
```
viliam-durina: Again. You also quote only execution ID and not job ID.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/StatisticalMetricsTest.java
```diff
@@ -0,0 +1,214 @@
...
+import static com.hazelcast.jet.core.TestProcessors.MockPMS;
+import static com.hazelcast.jet.core.TestProcessors.MockPS;
+import static com.hazelcast.jet.core.TestProcessors.reset;
+import static com.hazelcast.jet.core.test.JetAssert.assertTrue;
+import static com.hazelcast.test.HazelcastTestSupport.assertTrueEventually;
+import static java.util.Collections.singletonList;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+public class StatisticalMetricsTest {
```
viliam-durina: What's "statistical metrics"? Better name would be "JobLifecycleMetricsTest".

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/StatisticalMetricsTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+    private static final int MEMBER_COUNT = 1;
```
viliam-durina: We should test on 2 members and check the non-master for execution-related metrics, esp. after fixing `JOB_EXECUTIONS_STARTED`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MetricNames.java
```diff
@@ -95,26 +92,59 @@
...
+     */
+    public static final String EXECUTION_START_TIME = "executionStartTime";
+
+    /**
+     * Tracks the total amount of time since a given execution of a specific
+     * job is running. The execution and the job can be identified based
+     * on the {@link MetricTags#EXECUTION} & {@link MetricTags#JOB} tags
+     * of the metric.
+     */
+    public static final String EXECUTION_DURATION = "executionDuration";
```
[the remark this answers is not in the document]
cangencer: For completed jobs, I don't think you can get it any other way? might be better to store completion time in that case, though.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ExecutionContext.java
```diff
@@ -311,6 +318,13 @@ public void provideDynamicMetrics(MetricDescriptor descriptor, MetricsCollection
         }
         descriptor = descriptor.withTag(MetricTags.JOB, idToString(jobId))
                                .withTag(MetricTags.EXECUTION, idToString(executionId));
+
+        long executionStartTime = startTime.get();
+        if (executionStartTime != 0) {
+            context.collect(descriptor, EXECUTION_START_TIME, ProbeLevel.INFO, ProbeUnit.MS, executionStartTime);
+            context.collect(descriptor, EXECUTION_DURATION, ProbeLevel.INFO, ProbeUnit.MS,
```
cangencer: let's remove this, as it's just a derived value and obsolete as soon as it's collected

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/StoreSnapshotTasklet.java
cangencer: Mostly because they're updated together and in very low frequency (once every few sec), I don't know if there's something else that's comparable? There's still a very slim chance you get mismatching values, it might look confusing and won't be updated until next snapshot.

viliam-durina: We never synchronize any metrics. For example, you can observe `totalKeys` zero and `totalWindows` non-zero. Or you can see 1 received item in a downstream processor and 0 emitted at upstream processor (and the downstream processor didn't generate the item). Even using lazySet can cause that.

I'm not bothered by this at all. We're constantly turning metrics into an application logic thing that must be perfectly reliable and consistent. Metrics give you a peep at the engine's entrails. Compare it to car's diagnostics, each value you see is a bit late and each is a different bit late.

Also the all snapshots tend to be similarly large, so seeing two values each from a different snapshot will not be immediately apparent.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobExecutionService.java
```diff
@@ -366,4 +381,21 @@ public void provideDynamicMetrics(MetricDescriptor descriptor, MetricsCollection
...
     }
+
+    private static class Stats {
+
+        @Probe(name = MetricNames.JOB_EXECUTIONS_STARTED)
+        private final Counter executionStarted = MwCounter.newMwCounter();
+        @Probe(name = MetricNames.JOB_EXECUTIONS_TERMINATED)
+        private final Counter executionTerminated = MwCounter.newMwCounter();
+
+        public void executionStarted() {
```
cangencer: it's a little unnecessary encapsulation here, you can just increment the counters directly.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -968,4 +984,27 @@ NodeEngineImpl nodeEngine() {
...
+    private static class Stats {
+
+        @Probe(name = MetricNames.JOBS_SUBMITTED)
+        private final Counter jobSubmitted = MwCounter.newMwCounter();
+        @Probe(name = MetricNames.JOBS_COMPLETED_SUCCESSFULLY)
+        private final Counter jobCompletedSuccessfully = MwCounter.newMwCounter();
+        @Probe(name = MetricNames.JOBS_COMPLETED_WITH_FAILURE)
+        private final Counter jobCompletedWithFailure = MwCounter.newMwCounter();
+
+        void jobSubmitted() {
```
cangencer: same comment as before, unnecessary encapsulation

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MetricNames.java
```diff
@@ -95,26 +92,59 @@
...
+     */
+    public static final String EXECUTION_START_TIME = "executionStartTime";
+
+    /**
+     * Tracks the completion time of a given execution of a specific job.
+     * The execution and the job can be identified based on the
+     * {@link MetricTags#EXECUTION} & {@link MetricTags#JOB} tags of
+     * the metric.
+     */
+    public static final String EXECUTION_COMPLETION_TIME = "executionCompletionTime";
```
viliam-durina: Isn't the execution gone as soon as the job is completed? One can observe a non-zero value in a very short window.
If not, we should document that it's 0 when the job is running.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/OutboxImpl.java
viliam-durina: Hmm, number of snapshots might be interesting, even though one can estimate that from the interval..

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MetricNames.java
cangencer: it should be in the completed jobs metrics, I think?

viliam-durina: I realized it might be useful to have executionCompletionTime if one stores metrics after job completion. I added a test for it and fixed it...

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/metrics/StatisticalMetricsTest.java
jbartok: Yes, indeed, I'll add a generic TODO for it and try to do it next week, this shouldn't hinder us in merging in.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/MetricNames.java
jbartok: Which then brakes other things...

jbartok: Pardon my momentary frustration... Well it's not momentary, stopping it from coming out is.

jbartok:
@cangencer, can I merge this in?

cangencer:
Yes

PR #1721, opened 2019-11-05
Improve user-defined metrics related code & documentation

jbartok:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/ExecutionPlan.java
```diff
@@ -135,7 +135,7 @@
     }
 
     public void initialize(
-            NodeEngine nodeEngine, long jobId, long executionId, SnapshotContext snapshotContext
+        NodeEngine nodeEngine, long jobId, long executionId, SnapshotContext snapshotContext, boolean metricsEnabled
```
viliam-durina: Should be indented by 8 spaces

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsContext.java
```diff
@@ -208,4 +213,51 @@ public long get() {
...
+        public void decrement(long amount) {
+        }
+
+        @Override
+        public void set(long newValue) {
+        }
+
+        @Override
+        public long get() {
+            return 0;
```
viliam-durina: This is wrong...

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsContext.java
```diff
@@ -208,4 +213,51 @@ public long get() {
             return VOLATILE_VALUE_UPDATER.get(this);
         }
     }
+
+    private static final class DummyMetric implements Metric {
```
viliam-durina: Better name might be `NoopMetric`. We can even use singleton instance of this class, it's stateless.

hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/log-debug.adoc
```diff
@@ -271,11 +271,15 @@ distinguishes them from regular job-specific metrics is exactly what
 their name implies: they are not built-in, but defined when processing
 pipelines are written.
 
-Since user metrics are also job metrics, they will have all the tags
+Since user-defined metrics are also job metrics, they will have all the tags
 job metrics have. They also have an extra tag, called `USER` which is
```
viliam-durina: should be lower-case `user`

hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/log-debug.adoc
```diff
@@ -284,16 +288,27 @@ this:
...
+When obtaining `Metric` instances from the `Metrics` utility class it's
+possible to specify the desired `{jet-javadoc}/core/metrics/Unit.html[measurement unit]`
+of the metrics. It's important to note though that this measurement unit
+serves a purely informative purpose only (it's being used to set one of the
+tags of the metrics, for details see the
+`{jet-javadoc}/core/metrics/Unit.html[javadoc]`), no validation is being
+done on it. The `Metric` instance for a certain name can be requested multiple times
+from the utility class with different `Unit` values and this will not generate
+any errors. The first `Unit` value from the first request will take effect,
+subsequent ones will be ignored.
```
viliam-durina: The language is a bit garrulous:
- "It's important to note though that this measurement unit serves a purely informative purpose only" -> "The unit can be used by UI tools to format the value, it's ignored by Jet."
- "no validation is being done on it" -> already covered by "ignored by Jet".
- "The `Metric` instance for a certain ... will be ignored" -> "If you request a metric multiple times with different unit, the unit of the first request will be used to register the metric, others will be ignored"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsContext.java
jbartok: Renamed it, but making it a singleton is not really an option until we have the `name` and `unit` methods `Metric`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsContext.java
jbartok: Will fix by removing the `Metric.get()` method.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsContext.java
viliam-durina: True, I retract

jbartok:
@viliam-durina Pls. take another look at `MetricsContext`, I'm not sure you'll like the abstract class I invented there.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsContext.java
```diff
@@ -80,13 +87,19 @@ private Metric metric(String name, Unit unit, BiFunction<String, Unit, Metric> m
     }
 
     public void collectMetrics(MetricTagger tagger, MetricsCollectionContext context) {
+        if (!enabled) {
```
cangencer: this is not needed as this method is not called when metrics are disabled.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/metrics/MetricsContext.java
```diff
@@ -204,8 +214,40 @@ public void set(long newValue) {
         }
 
         @Override
-        public long get() {
+        protected long get() {
             return VOLATILE_VALUE_UPDATER.get(this);
         }
     }
+
+    private static final class NoopMetric extends AbstractMetric {
```
cangencer: this seems like a micro-optimization to me. We don't disable Jet-level metrics when metrics collection is disabled. I would just remove it.

cangencer: perf impact of metrics when doing an update is pretty much nothing due to use of lazySet. threadSafeMetrics is only useful for async processors. We have many more metrics for example in ProcessorTasklet which is updated more frequently and we don't optimize for those cases.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/metrics/Metric.java
```diff
@@ -71,9 +72,4 @@
      */
     void set(long newValue);
 
-    /**
-     * Returns the current value.
-     */
-    long get();
```
olukas: Actually this means that value of User Metric cannot be obtained in user code. Doesn't it too restrictive? I believe there can be scenarios where it is useful to `get` current value, modify it and `set` back to metric.

jbartok: Doing that would be not safe, just a simple `get` method doesn't allow you to do it atomically. Better to not have it at all.

PR #1737, opened 2019-11-13
Auto import modules in JDK9+ in start script

cangencer:
Automatically add required modules when starting Jet from JDK9+

### DIFF
--- a/hazelcast-jet-distribution/src/bin-regular/common.sh
+++ b/hazelcast-jet-distribution/src/bin-regular/common.sh
@@ -12,10 +12,23 @@ if [ -z "$JAVA" ]; then
     exit 1
 fi
 
-CLASSPATH="$JET_HOME/lib/*:$CLASSPATH:$CLASSPATH_DEFAULT"
+# 1 -> Java 8 or earlier (1.8..)
+# 9, 10, 11 -> JDK9, JDK10, JDK11 etc.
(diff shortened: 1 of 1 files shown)


hazelcast-jet-distribution/src/bin-regular/common.sh
```diff
@@ -12,19 +12,36 @@ if [ -z "$JAVA" ]; then
     exit 1
 fi
 
-CLASSPATH="$JET_HOME/lib/*:$CLASSPATH:$CLASSPATH_DEFAULT"
+# 1 -> Java 8 or earlier (1.8..)
+# 9, 10, 11 -> JDK9, JDK10, JDK11 etc.
+JAVA_VERSION=$(${JAVA} -version 2>&1 | sed -En 's/.* version "([0-9]+).*$/\1/p')
```
viliam-durina: Will this work for other JVM impls? Ibm, Azul...

cangencer: I tested on OpenJDK and HotSpot. These will need additional testing.

mtopolnik: The output of `java -version` is specified: https://www.oracle.com/technetwork/java/javase/versioning-naming-139433.html

PR #1740, opened 2019-11-14
Parallelize tasklet initialization

mtopolnik:
Some tasklets have heavyweight initialization, this helps to speed up job start in such cases.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
@@ -38,12 +38,16 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ArrayBlockingQueue;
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
```diff
@@ -208,6 +203,26 @@ private void submitCooperativeTasklets(
...
+                future.get();
+            } catch (InterruptedException | ExecutionException e) {
+                logger.severe("Tasklet initialization failed", e);
+                firstFailure = firstFailure != null ? firstFailure : peel(e);
+                failureCount++;
+            }
+        }
+        if (firstFailure != null) {
+            throw new JetException(String.format(
+                    "%,d of %,d tasklets failed to initialize. One of the failures is attached as the cause.",
```
viliam-durina: It's better to add the `cause` to the message. Some tools hide the stack trace and causes, then you won't see nothing, maybe mancenter does that too.
```java
"%,d of %,d tasklets failed to initialize. One of the failures: " + firstFailure,
```

PR #1748, opened 2019-11-22
Some minor tweaks to hello-world example

cangencer:
* Make output map name non-random, so you can start multiple clients
* Add a yaml file to use as config when using embedded mode.


cangencer:
verify

PR #1749, opened 2019-11-25
Fix @Nullable name in BatchStage

### LINKED ISSUE #1675: Inconsistent @Nonnull and @Nullable annotations
`BatchStage.setName()` has `@Nullable` parameter, but some subclasses have it `@Nonnull`.

cangencer:
Fixes #1675

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/transform/AbstractTransform.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/transform/AbstractTransform.java
@@ -21,6 +21,7 @@ import com.hazelcast.jet.core.Vertex;
 import javax.annotation.Nonnull;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import static java.util.Collections.singletonList;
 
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/transform/AbstractTransform.java
```diff
@@ -50,6 +51,7 @@ protected AbstractTransform(String name, @Nonnull Transform upstream) {
 
     @Override
     public void setName(@Nonnull String name) {
+        Objects.requireNonNull(name, " name");
```
viliam-durina: You can do 
```
this.name = requireNonNull(name, "name");
```
There's also extra space in the string

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/AbstractStage.java
```diff
@@ -21,9 +21,11 @@
 import com.hazelcast.jet.pipeline.Pipeline;
 import com.hazelcast.jet.pipeline.SinkStage;
 import com.hazelcast.jet.pipeline.Stage;
+import jdk.internal.jline.internal.Preconditions;
```
viliam-durina: Import from `jdk.internal`?

cangencer: This was leftover from when I added the check to AbstractStage, fixed now.

PR #1752, opened 2019-11-25
Set aws plugin version to 3.0.1-SNAPSHOT

gurbuzali:
aws plugin has a specific branch which is up to date with the latest IMDG(4.0) changes


cangencer:
could you add tags?

PR #1754, opened 2019-11-26
Follow changes on client protocol and remove client messages

eminn:
which have counterpart messages on the IMDG MC


hazelcast-jet-core/src/main/resources/client-protocol-definition/Jet.yaml
```diff
@@ -52,7 +50,6 @@ methods:
     doc: ''
     request:
       retryable: true
-      acquiresResource: false
       partitionIdentifier: jobId
```
cangencer: this is not correct - this request always goes to master, the same for other similar requests. I don't know what effect this has though, or why it's part of the protocol definition..

eminn:
verify

PR #1758, opened 2019-11-27
Remove MapEntry and PriorityQeueu hooks since they are on IMDG side

gurbuzali:
see https://github.com/hazelcast/hazelcast/issues/15179#issuecomment-558966682

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/serialization/MapEntryHook.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/serialization/MapEntryHook.java
@@ -23,15 +23,16 @@ import com.hazelcast.nio.serialization.SerializerHook;
 import com.hazelcast.nio.serialization.StreamSerializer;
 
 import java.io.IOException;
+import java.util.AbstractMap;
 import java.util.Map.Entry;
 
 import static com.hazelcast.jet.Util.entry;
 
-public final class MapEntryHook implements SerializerHook<Entry> {
+public final class MapEntryHook implements SerializerHook<AbstractMap.SimpleImmutableEntry> {
 
     @Override
-    public Class<Entry> getSerializationType() {
-        return Entry.class;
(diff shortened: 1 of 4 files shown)


gurbuzali:
```
java.util.concurrent.CompletionException: com.hazelcast.jet.JetException: Exception in SenderTasklet{ordinal=0, destinationAddress=[127.0.0.1]:5701, sourceVertexName='group-and-aggregate-prepare'}: com.hazelcast.nio.serialization.HazelcastSerializationException: Failed to serialize 'java.util.AbstractMap$SimpleImmutableEntry'
	at java.util.concurrent.CompletableFuture.reportJoin(CompletableFuture.java:375)
	at java.util.concurrent.CompletableFuture.join(CompletableFuture.java:1934)
	at com.hazelcast.jet.Job.join(Job.java:137)
```

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/serialization/MapEntryHook.java
```diff
@@ -23,15 +23,16 @@
 import com.hazelcast.nio.serialization.StreamSerializer;
 
 import java.io.IOException;
+import java.util.AbstractMap;
 import java.util.Map.Entry;
 
 import static com.hazelcast.jet.Util.entry;
 
-public final class MapEntryHook implements SerializerHook<Entry> {
+public final class MapEntryHook implements SerializerHook<AbstractMap.SimpleImmutableEntry> {
```
eminn: maybe we can add `AbstractMap.SimpleImmutableEntry` serializer to IMDG side so we don't maintain any serializers on our side ?

gurbuzali: this was a work-around, just to make builds pass till we find a real fix for the issue on imdg side. but even with this change there are some test failures
`Failed to serialize 'com.hazelcast.query.impl.QueryEntry'`

PR #1764, opened 2019-11-28
Move Data to private API

mmedenjak:
Data was part of the custom-user services SPI which was moved to private
API.

Depends on: https://github.com/hazelcast/hazelcast/pull/16147


mtopolnik:
verify

mmedenjak:
verify

PR #1766, opened 2019-11-29
Add option to use UNICAST for the edge

viliam-durina:


cangencer:
would be nice to add a simple test so it doesn't show up as unused in the API.

PR #1774, opened 2019-12-02
Update default parallelism of map/cache sources and sinks to 1

cangencer:
Also move the lambda to be the last parameter of `ProcessorMetaSupplier.of()` method


hazelcast-jet-avro/src/main/java/com/hazelcast/jet/avro/AvroProcessors.java
```diff
@@ -77,13 +77,13 @@ private AvroProcessors() {
             @Nonnull SupplierEx<DatumWriter<D>> datumWriterSupplier
     ) {
         return ProcessorMetaSupplier.of(
-                WriteBufferedP.<DataFileWriter<D>, D>supplier(
+                1, WriteBufferedP.<DataFileWriter<D>, D>supplier(
```
viliam-durina: There's `PMS.preferLocalParallelismOne`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/HazelcastWriters.java
```diff
@@ -62,7 +62,7 @@ private HazelcastWriters() {
             @Nonnull String name,
             @Nullable ClientConfig clientConfig
     ) {
-        return ProcessorMetaSupplier.of(new WriteMapP.Supplier<>(asXmlString(clientConfig), name));
+        return ProcessorMetaSupplier.of(1, new WriteMapP.Supplier<>(asXmlString(clientConfig), name));
```
viliam-durina: Javadoc needs updating: https://github.com/hazelcast/hazelcast-jet/blob/ab58c5085964a66538344f0287dbf1b59cb760e6/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/Sources.java#L164-L165

cangencer: I wonder if we should even document parallelism here.

PR #1775, opened 2019-12-03
Add Blue/Green Documentation section to reference manual

gurbuzali:


hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/configure-fault-tolerance.adoc
```diff
@@ -239,3 +239,40 @@ members must be present after the restart for Jet to be able to reload
...
+<hazelcast-client-failover>
+    <try-count>10</try-count>
+    <clients>
+        <client>hazelcast-client-A.xml</client>
+        <client>hazelcast-client-B.xml</client>
+    </clients>
+</hazelcast-client-failover>
+----
+
+After switching to the cluster-B, any running job on cluster-A will be
```
olukas: `After switching to the cluster-B, any running job on cluster-A will be canceled` - this is not correct - connection to cluster-A is lost from client but it does not mean that running jobs on cluster-A will be canceled. 

Moreover it can be problematic for users - original job can be still running in cluster-A then resubmit them to another cluster can cause unexpected data duplication in remote sink. Maybe we should specify that Blue/Green in Jet context is mostly applicable for short living batch jobs (i.e. queries expressed as Jet Jobs).

olukas:
I would also appreciate information that clients in Jet context are used mostly for submitting jobs to cluster.

PR #1776, opened 2019-12-03
Fix merging of differently-timestamped streams

viliam-durina:
We only improve the reported error, it failed before too, but depending on the order it failed either at runtime with a ClassCastException or when adding a window it complained that the pipeline doesn't have timestamps.

### DIFF
--- a/hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/StreamStageTest.java
+++ b/hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/StreamStageTest.java
@@ -926,7 +926,24 @@ public class StreamStageTest extends PipelineStreamTestSupport {
         exception.expectMessage("both have or both not have timestamp definitions");
 
         // When
-        nonTimestamped.merge(timestamped);
+        timestamped.merge(nonTimestamped);
+    }
+
+    @Test
+    public void when_mergeToItself_then_doubleOutput() {
+        List<Integer> input = sequence(itemCount);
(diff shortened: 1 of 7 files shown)


hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/StreamStageTest.java
```diff
@@ -898,6 +905,30 @@ public void merge() {
...
+    @Test
+    public void when_mergeTimestampedToNonTimestamped_then_error() {
+        StreamStage<SimpleEvent> timestamped = p.readFrom(TestSources.itemStream(1)).withIngestionTimestamps();
+        StreamStage<SimpleEvent> nonTimestamped = p.readFrom(TestSources.itemStream(1)).withoutTimestamps();
+
+        // Then
+        exception.expectMessage("both have or both not have timestamp definitions");
+
+        // When
+        nonTimestamped.merge(timestamped);
```
cangencer: I think this is meant to be `timestamped.merge(nonTimestamped)`

viliam-durina: It's disputable, test name is `mergeTimestampedToNonTimestamped`, in `A.merge(B)` I consider that I'm merging `B` into `A`. Most importantly, there are two methods and they merge both ways.

viliam-durina: ah, no, the other method is same 😒

PR #1778, opened 2019-12-04
Improve the default config to add additional discovery options

cangencer:


hazelcast-jet-distribution/src/root/config/hazelcast.yaml
```diff
@@ -1,30 +1,78 @@
...
+        label: application=hazelcast-jet
+        zones: us-east1-b,us-east1-c
+    # Allows access to cluster via the REST API
+    rest-api:
+      enabled: true
+      endpoint-groups:
+        CLUSTER_READ:
+          enabled: true
+        CLUSTER_WRITE:
+          enabled: true
```
gurbuzali: this enables cluster shutdown and it is `disabled` by default for IMDG, is it a good idea to enable it by default ?

cangencer: It is useful for doing shutdown from MC and using the cluster.sh scripts. I thought we enable it by default so those scripts work.

cangencer:
verify

cangencer:
verify

PR #1783, opened 2019-12-05
Add daemon mode and rolling file logger

cangencer:


hazelcast-jet-distribution/src/bin-filemode-755/jet-start
```diff
@@ -2,14 +2,30 @@
...
 echo "# JAVA_OPTS=${JAVA_OPTS_ARRAY[*]}"
 echo "# CLASSPATH=$CLASSPATH"
 echo "########################################"
-echo "Starting Hazelcast Jet"
 
-set -x
-$JAVA "${JAVA_OPTS_ARRAY[@]}" -cp "$CLASSPATH" com.hazelcast.jet.server.JetMemberStarter
+if [[ "$DAEMON" = "true" ]]; then
+    mkdir -p $JET_HOME/logs
+    JET_LOG="$JET_HOME/logs/hazelcast-jet.out"
```
gurbuzali: we create `hazelcast-jet.out` if daemon true and `hazelcast-jet.log` otherwise

hazelcast-jet-distribution/src/bin-filemode-755/jet-start
```diff
@@ -2,14 +2,30 @@
...
+    echo "Starting Hazelcast Jet in daemon mode. Standard out and error will be written to $JET_LOG"
+    nohup $JAVA "${JAVA_OPTS_ARRAY[@]}" -cp "$CLASSPATH" com.hazelcast.jet.server.JetMemberStarter > "$JET_LOG" 2>&1 &
```
gurbuzali: we can get rid off specific log forwarding
`nohup $JAVA "${JAVA_OPTS_ARRAY[@]}" -cp "$CLASSPATH" com.hazelcast.jet.server.JetMemberStarter &`
should be enough thanks to log4j configuration. This eliminates above naming issue too

hazelcast-jet-distribution/src/bin-filemode-755/jet-cluster-admin
```diff
@@ -20,14 +21,14 @@ if [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
     exit 0
 fi
 
-command -v curl >/dev/null 2>&1 || { \
```
gurbuzali: this change makes each call to print `/usr/bin/curl` to output

cangencer: do we need `-v` there? I think that's the part that prints the text?

hazelcast-jet-distribution/src/bin-filemode-755/jet-start
cangencer: hazelcast-jet.log is always created (by log4j logger) and it's appended. the .out file the sysout, and deleted when you restart the node.

hazelcast-jet-distribution/src/bin-filemode-755/jet-start
cangencer: I think if you turn this off and there's some error on startup i.e. node fails to start, then it won't be logged. I didn't find a good solution to this though, Kafka does something similar.

hazelcast-jet-distribution/src/bin-filemode-755/jet-start
gurbuzali: ah so it was intentional

hazelcast-jet-distribution/src/bin-filemode-755/jet-cluster-admin
gurbuzali: yes without `-v` it does not work at all. `command -v` outputs where the binary for that command is

cangencer: I reverted the change, I got confused that this call was actually invoking curl

hazelcast-jet-distribution/src/root/config/log4j.properties
```diff
@@ -13,14 +13,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+log4j.rootLogger=INFO, stdout, rollingFile
```
mtopolnik: Logging to `stdout` in daemon mode is bad. We should address this in future work.

Issue #1731, opened 2019-11-09
Unify Source Context and Sink Writer


jerrinot:
`SourceBuilder` works with user [`Context`](https://github.com/hazelcast/hazelcast-jet/blob/3721541fb419792b2d5b00dc39468b8803c047d5/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/SourceBuilder.java#L180) while the `SinkBuilder` works with a ["Writer"](https://github.com/hazelcast/hazelcast-jet/blob/3721541fb419792b2d5b00dc39468b8803c047d5/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/SinkBuilder.java#L92). 

These objects are essentially the same: Both represent a state of a Source/Source. The terminology should be unified.

PR #1785, opened 2019-12-06
Use context object instead of writer for sink-builder

gurbuzali:
fixes https://github.com/hazelcast/hazelcast-jet/issues/1731


hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/SinkBuilder.java
```diff
@@ -34,23 +34,23 @@
 /**
  * See {@link SinkBuilder#sinkBuilder(String, FunctionEx)}.
  *
- * @param <W> type of the writer object
+ * @param <C> type of the writer object
```
mtopolnik: writer -> context

PR #1787, opened 2019-12-06
ServiceFactory provides a custom ServiceContext to createFn

cangencer:
We use `Context` in `SourceBuilder` and `SinkBuilder` but not in `ServiceFactory`. It should follow the same pattern. Since the service can be both owned by a processor and shared among all local processors, neither of `Processor.Context` and `ProcessorSupplier.Context` is a fit for all cases.


mtopolnik:
I'm making the same changes in my Python branch, but my realization was that `ProcessorSupplier.Context` isn't enough when you instantiate a service. You still don't know the index of the service instance. In my branch I introduced a dedicated `ServiceContext` instead, which also makes sense because it doesn't mix in Core API.

cangencer:
> I'm making the same changes in my Python branch, but my realization was that `ProcessorSupplier.Context` isn't enough when you instantiate a service. You still don't know the index of the service instance. In my branch I introduced a dedicated `ServiceContext` instead, which also makes sense because it doesn't mix in Core API.

Do you mind porting it to this PR? So we decouple that from the python work?

viliam-durina:
We could use `ProcessorSupplier.Context`, but we can document that if local sharing is off, you can safely downcast to `Processor.Context`, wdyt?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/ServiceFactory.java
```diff
@@ -107,7 +107,7 @@ private ServiceFactory(
      */
     @Nonnull
     public static <S> ServiceFactory<S> withCreateFn(
-            @Nonnull FunctionEx<JetInstance, ? extends S> createServiceFn
+            @Nonnull FunctionEx<Context, ? extends S> createServiceFn
```
viliam-durina: We should use `? super Context` here. `? super ProcessorSupplier.Contex` would be even better. Same in `SinkBuilder.sinkBuilder()`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/ServiceFactory.java
```diff
@@ -279,4 +282,21 @@ public int maxPendingCallsPerProcessor() {
     public boolean hasOrderedAsyncResponses() {
         return orderedAsyncResponses;
     }
+
+    public interface ServiceContext<S> {
+        int localIndex();
```
cangencer: we're also missing vertexName perhaps?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/ServiceFactory.java
```diff
@@ -279,4 +282,21 @@ public int maxPendingCallsPerProcessor() {
...
+
+        int globalIndex();
+
+        int jetMemberIndex();
```
cangencer: we use memberIndex in PS.Context

mtopolnik: In this higher-level API I thought this would be clearer.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/ServiceContextImpl.java
```diff
@@ -0,0 +1,101 @@
...
+    private final ILogger logger;
+    private final JetInstance jetInstance;
+
+    public ServiceContextImpl(
+            @Nonnull ServiceFactory<?> serviceFactory,
+            @Nonnull ProcessorSupplier.Context context
+    ) {
+        this.memberCount = context.memberCount();
+        this.memberIndex = context.memberIndex();
+        this.localIndex = context instanceof Processor.Context
```
mtopolnik: It's a bit awkward that we use two styles of checking the same thing (that the service is shared). I'd prefer to first establish `hasLocalSharing` and then assert the type of the context, which must be `Processor.Context` if there is no local sharing and vice versa.

PR #1800, opened 2019-12-12
Introduced an option to pick main class from the JAR file

### LINKED ISSUE #1796: Select an entry class when submitting job from the CLI
The entry point to a Jet JAR must be declared as a Main-Class in MANIFEST.MF.  Keep this as a default but allow the user to override the entry class with a CLI parameter, e.g.

`jet submit --class com.hazelcast.jet.demos.HelloWorld helloWorld.jar`

eminn:
Fixes #1796

### DIFF
--- a/hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetBootstrap.java
+++ b/hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetBootstrap.java
@@ -19,6 +19,7 @@ package com.hazelcast.jet.server;
 import com.hazelcast.cluster.Cluster;
 import com.hazelcast.collection.IList;
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.internal.util.StringUtil;
 import com.hazelcast.jet.Jet;
 import com.hazelcast.jet.JetCacheManager;
 import com.hazelcast.jet.JetException;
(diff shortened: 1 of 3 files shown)


hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetBootstrap.java
```diff
@@ -108,20 +109,22 @@ private JetBootstrap() {
...
         try (JarFile jarFile = new JarFile(jar)) {
-            if (jarFile.getManifest() == null) {
-                error("No manifest file in " + jar);
-            }
-            String mainClass = jarFile.getManifest().getMainAttributes().getValue("Main-Class");
-            if (mainClass == null) {
-                error("No Main-Class found in manifest");
+            if (StringUtil.isNullOrEmpty(mainClass)) {
+                if (jarFile.getManifest() == null) {
+                    error("No manifest file in " + jar);
```
viliam-durina: We can hint here that one can use `-c` option. Also on line 126.

eminn: After I gave it a thought, if user is using the JetBootstrap directly, then it might not be meaningful to say `use -c option`. WDYT ?

cangencer: How do you mean using JetBootstrap directly?

eminn: Never mind, I've missed the fact that the executeJar method is package private.

PR #1803, opened 2019-12-12
Add mapUsingServiceBatched

cangencer:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/Processors.java
```diff
@@ -1014,13 +1015,33 @@ private Processors() {
...
+     * @param <T>
+     * @param <R>
+     * @return
+     */
+    @Nonnull
+    public static <S, T, R> ProcessorSupplier flatMapUsingServiceAsyncBatchedP(
+            @Nonnull ServiceFactory<?, S> serviceFactory,
+            int maxBatchSize,
+            @Nonnull BiFunctionEx<? super S, ? super List<T>,
+                    ? extends CompletableFuture<List<Traverser<R>>>> flatMapAsyncFn
```
viliam-durina: We should take a `Collection` here so that a `Set` can be used also. Expecting to return a collection also makes sense, as long it's of the same size.

viliam-durina: @mtopolnik I think it's clear enough from the API what you're doing. The collections are supposed to have equal number of items. And every collection has an _iteration order_, you just can't get the items by index. The processor preserves the order of batches, but if you want to reorder the results for one batch, even a `List` won't prevent you from doing that.

But I realized the first argument can actually be a `List`, because that's what we give. But we're fine with a `Collection` in return:

```java
@Nonnull BiFunctionEx<? super S, ? super List<T>,
                    ? extends CompletableFuture<Collection<Traverser<R>>>> flatMapAsyncFn
```

I think it should actually be:
```java
@Nonnull BiFunctionEx<? super S, ? super List<T>,
                    ? extends CompletableFuture<? extends Collection<? extends Traverser<? extends R>>>> flatMapAsyncFn
```

mtopolnik: Returning a `Set` really makes no sense here, but we can generalize just for good measure. Nested wildcards don't have the effect that you'd expect because there's no wildcard capture for them.

viliam-durina: We should use the lowest interface that works for us. That's `Collection`.

cangencer:
verify

PR #1807, opened 2019-12-13
Remove too much flexibility from the JMS sink & improvements

viliam-durina:
They would prevent implementing ex-once in the future without breaking changes.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/SinkProcessors.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/SinkProcessors.java
@@ -295,11 +295,11 @@ public final class SinkProcessors {
      */
     @Nonnull
     public static <T> ProcessorMetaSupplier writeJmsQueueP(
+            @Nonnull String queueName,
             @Nonnull SupplierEx<? extends Connection> newConnectionFn,
-            @Nonnull BiFunctionEx<? super Session, ? super T, ? extends Message> messageFn,
-            @Nonnull String name
+            @Nonnull BiFunctionEx<? super Session, ? super T, ? extends Message> messageFn
     ) {
-        return WriteJmsP.supplier(newConnectionFn, messageFn, name, false);
+        return WriteJmsP.supplier(queueName, newConnectionFn, messageFn, false);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/SinkProcessors.java
```diff
@@ -312,13 +308,10 @@ public static ProcessorMetaSupplier writeRemoteListP(@Nonnull String listName, @
     @Nonnull
     public static <T> ProcessorMetaSupplier writeJmsTopicP(
             @Nonnull SupplierEx<? extends Connection> newConnectionFn,
-            @Nonnull FunctionEx<? super Connection, ? extends Session> newSessionFn,
             @Nonnull BiFunctionEx<? super Session, ? super T, ? extends Message> messageFn,
```
cangencer: should we still expose `Session` here?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/SinkProcessors.java
```diff
@@ -297,13 +296,10 @@ public static ProcessorMetaSupplier writeRemoteListP(@Nonnull String listName, @
     @Nonnull
     public static <T> ProcessorMetaSupplier writeJmsQueueP(
             @Nonnull SupplierEx<? extends Connection> newConnectionFn,
-            @Nonnull FunctionEx<? super Connection, ? extends Session> newSessionFn,
             @Nonnull BiFunctionEx<? super Session, ? super T, ? extends Message> messageFn,
-            @Nonnull BiConsumerEx<? super MessageProducer, ? super Message> sendFn,
-            @Nonnull ConsumerEx<? super Session> flushFn,
             @Nonnull String name
```
cangencer: lambda should be last param here also.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/SinkProcessors.java
viliam-durina: Yes, you need it to create the message, e.g. `session.createTextMessage("foo")`. 
The change is that Jet creates the session with parameters as needed, but user still uses it.

PR #1819, opened 2019-12-20
Remove hN tags from method javadocs

viliam-durina:
They look bad in generated javadocs, replaced with plain bold text. We keep the
usages in class or package javadocs, there they look ok.

Before:
javadoc-before

After:
javadoc-after


cangencer:
I think using <h4> or <h5> is fine, have we tried how it looks?

viliam-durina:
subheadings

So only `h3` and `h4` have the bands, these are those I used the most often. `h1` and `h2` are quite big and `h5` and `h6` are smaller than the regular text. I like the bold text the best.

PR #1824, opened 2019-12-23
Introduce non-classpath resource upload mechanism

eminn:
Provides convenience for users to attach their files and directories in the Jet pipelines.  

Checklist
- [x] Tags Set
- [x] Milestone Set

Breaking Changes:
- Removed `REGULAR_FILE` type and added `CLASSPATH_RESOURCE` ,`FILE` and `DIRECTORY` to `ResourceType` enum 
- Renamed `JobConfig.addResource` to `addClasspathResource`

Co-authored-by: Marko Topolnik <marko@hazelcast.com>


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/ProcessorMetaSupplier.java
```diff
@@ -393,5 +407,35 @@ default boolean snapshottingEnabled() {
...
+        /**
+         * Returns a temporary directory which contains the attached files to the
+         * job with the given id.
+         *
+         * @param id identifier defined on the {@link ResourceConfig} to be
+         *           used retrieve files from storage.
+         * @return {@link File} handle to a temporary directory which contains the
+         * files attached to the job with provided identifier.
+         */
+        default File getJobFileStorageById(String id) {
```
cangencer: shouldn't this be part of the context ? It seems strange to add it to PMS? Is it meant to be overriden?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/ResourceConfig.java
```diff
@@ -40,8 +41,8 @@
      * @param id            id of the resource
      * @param resourceType  type of the resource
      */
-    ResourceConfig(@Nonnull URL url, String id, ResourceType resourceType) {
-        checkTrue((resourceType != ResourceType.REGULAR_FILE) ^ id != null,
+    ResourceConfig(@Nonnull URL url, @Nullable String id, @Nonnull ResourceType resourceType) {
```
cangencer: would it be better to always assign id here?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/ProcessorMetaSupplier.java
```diff
@@ -282,7 +284,8 @@ public void init(@Nonnull Context context) {
                                       .getPartition(key).getOwner().getAddress();
             }
 
-            @Nonnull @Override
+            @Nonnull
+            @Override
```
cangencer: this change can be reverted

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/ProcessorMetaSupplier.java
```diff
@@ -393,5 +396,28 @@ default boolean snapshottingEnabled() {
...
+        /**
+         * Returns the attached directory to the job with the given id.
+         *
+         * @param id identifier defined on the {@link ResourceConfig} to be
+         *           used retrieve files from storage.
+         * @return {@link File} handle to the attached directory to the job with provided identifier.
+         * @since 4.0
+         */
+        @Nonnull
+        File getAttachedDirectory(@Nonnull String id);
```
cangencer: you can skip the `get` (so just have `attachedDirectory` as it's the convention on this interface

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -351,6 +373,19 @@ void deleteJob(long jobId) {
         // delete the job record and related records
         jobExecutionRecords.remove(jobId);
         jobRecords.remove(jobId);
+        deleteTempFiles("jet-" + instance.getName() + "-" + idToString(jobId));
+    }
+
+    private void deleteTempFiles(String filter) {
```
cangencer: This seems rather expensive - it's going to iterate all temp files each time a job completes? We made a lot of optimizations for jobs that take just a few ms (small query jobs). It could be done as part of the Job GC process. Furthermore this is only called on master node, not on other nodes, how are those files on other nodes cleaned up?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -182,15 +192,25 @@ long uploadJobResources(JobConfig jobConfig) {
...
+                            readStreamAndPutCompressedToMap(rc.getId(), tmpMap, in);
+                        }
+                        break;
+                    case FILE:
+                        IMapOutputStream os = new IMapOutputStream(getJobFileStorage(jobId).get(),
+                                idToString(jobId) + rc.getId());
+                        zipFileToOutputStream(Paths.get(rc.getUrl().getFile()), os);
+                        break;
+                    case DIRECTORY:
+                        IMapOutputStream os2 = new IMapOutputStream(getJobFileStorage(jobId).get(),
```
cangencer: is this stream closed somewhere?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -246,6 +257,83 @@ public static void checkSerializable(Object object, String objectName) {
...
+        try (ZipOutputStream zipOut = new ZipOutputStream(outputStream)) {
+            Path fileName = filePath.getFileName();
+            if (fileName != null) {
+                zipOut.putNextEntry(new ZipEntry(fileName.toString()));
+                try (InputStream in = Files.newInputStream(filePath)) {
+                    copyStream(in, zipOut);
+                }
+                zipOut.closeEntry();
+            }
+        } catch (IOException e) {
```
cangencer: you can make the methods throws IOException , as it's already handled above.

cangencer:
We need unit tests for `IMapOutputStream`, testing edge cases.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -418,6 +558,12 @@ public JobConfig addResource(@Nonnull String path, @Nonnull String id) {
     }
 
     private JobConfig add(URL url, String id, ResourceType resourceType) {
+        boolean duplicate = resourceConfigs.stream()
+                                           .map(ResourceConfig::getId)
+                                           .anyMatch(existingId -> existingId.equals(id));
+        if (duplicate) {
+            throw new IllegalArgumentException("Resource with id:" + id + " already exists!");
```
cangencer: can we remove the `!` marks from exceptions? There's several in many places in the PR

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -418,6 +558,12 @@ public JobConfig addResource(@Nonnull String path, @Nonnull String id) {
     }
 
     private JobConfig add(URL url, String id, ResourceType resourceType) {
+        boolean duplicate = resourceConfigs.stream()
+                                           .map(ResourceConfig::getId)
```
cangencer: can be simplified like `.anyMatch(c -> id.equals(c.getId())`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/Contexts.java
```diff
@@ -108,26 +122,69 @@ public ILogger logger() {
...
         }
+
+        @Nonnull @Override
+        public File attachedDirectory(@Nonnull String id) {
+            Preconditions.checkHasText(id, "id cannot be null or empty!");
+            findResourceConfigOrThrowException(id);
+            JetInstance instance = jetInstance();
+            String jobId = idToString(jobId());
+            IMap<String, byte[]> map = instance.getMap(FILE_STORAGE_MAP_NAME_PREFIX + jobId);
+            try (IMapInputStream inputStream = new IMapInputStream(map, jobId, id)) {
```
cangencer: wouldn't this create the file everytime it's asked?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ExecutionContext.java
```diff
@@ -63,6 +66,7 @@
     private final Object executionLock = new Object();
     private final ILogger logger;
     private String jobName;
+    private List<File> localFiles = new ArrayList<>();
```
cangencer: this list is accessed concurrently, but is not thread-safe

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/Contexts.java
```diff
@@ -108,26 +122,69 @@ public ILogger logger() {
         public ProcessingGuarantee processingGuarantee() {
             return processingGuarantee;
         }
-    }
+
+           }
```
cangencer: wrong indentation

viliam-durina:
Is any of the listed breaking changes actually a breaking change? They are either an addition or are in private API.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -417,7 +570,14 @@ public JobConfig addResource(@Nonnull String path, @Nonnull String id) {
         return resourceConfigs;
     }
 
-    private JobConfig add(URL url, String id, ResourceType resourceType) {
+    private JobConfig add(@Nonnull URL url, @Nullable String id, @Nonnull ResourceType resourceType) {
```
viliam-durina: the `id` should be `@Nonnull` too, I checked all callers and they ensure it. Even if it's null, we pass it to `ResourceConfig` constructor where it's non-null.

cangencer:
@viliam-durina `JobConfig.addResource` was not private. But yes the description mixes breaking changes with non-breaking ones.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/test/TestProcessorSupplierContext.java
```diff
@@ -71,9 +75,24 @@ public int memberIndex() {
         return memberIndex;
     }
 
-    /**
-     * Set the member index.
-     */
+    @Nonnull @Override
+    public File attachedDirectory(@Nonnull String id) {
+        return attachedDirectory;
```
viliam-durina: We should store the files in a map, keyed by the `id`. This might be enough fur current cases, but it's an incorrect implementation.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/IMapOutputStream.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public void flush() throws IOException {
+        if (isClosed()) {
+            return;
+        }
+        boolean incompleteChunk = currentChunk.remaining() > 0;
+        byte[] value = currentChunk.array();
+        if (incompleteChunk) {
```
viliam-durina: Why this test? If there's no incomplete chunk, we'll store the entire backing array into the imap but we won't increment the chunk index and we'll overwrite it next time (if there is a _next time_)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/IMapOutputStream.java
```diff
@@ -0,0 +1,106 @@
...
+
+public class IMapOutputStream extends OutputStream {
+    private static final int CHUNK_SIZE = 1 << 17;
+
+    private final String prefix;
+    private final IMap<String, byte[]> map;
+    private final ByteBuffer currentChunk = ByteBuffer.allocate(CHUNK_SIZE);
+    private final byte[] singleByteBuffer = new byte[1];
+
+    private int currentChunkIndex = 1;
```
viliam-durina: If you started from 0, you'll have less plus-ones and minus-ones in the code.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -182,15 +192,23 @@ long uploadJobResources(JobConfig jobConfig) {
...
-                        readStreamAndPutCompressedToMap(rc.getId(), tmpMap, in);
-                        in.close();
+                    case CLASS:
+                        try (InputStream in = rc.getUrl().openStream()) {
+                            readStreamAndPutCompressedToMap(rc.getId(), tmpMap, in);
+                        }
+                        break;
+                    case FILE:
+                        IMapOutputStream os = new IMapOutputStream(getJobFileStorage(jobId).get(), rc.getId());
+                        zipFileToOutputStream(Paths.get(rc.getUrl().getFile()), os);
```
viliam-durina: What if the URL isn't a local file? It will break. We should use `rc.getUrl().openStream()`. 

Also the `zipFileToOutputStream` is named incorrectly. What it does is that it puts a file as a sole entry in a zip file.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/Contexts.java
```diff
@@ -108,26 +131,78 @@ public ILogger logger() {
...
         }
 
         @Override
         public int memberIndex() {
             return memberIndex;
         }
+
+        @Nonnull @Override
+        public File attachedDirectory(@Nonnull String id) {
+            Preconditions.checkHasText(id, "id cannot be null or empty");
```
viliam-durina: We don't require non-empty ID when adding files. Also again 7 lines below.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/Contexts.java
```diff
@@ -108,26 +131,78 @@ public ILogger logger() {
...
+            findResourceConfigOrThrowException(id);
```
viliam-durina: Why don't we just rely on the presence of the key in the file IMap? This is superfluous. I was even thinking whether we shouldn't make the `resourceConfigs` transient in `JobConfig`, the URLs might be invalid on another member. The ID is enough and the user is supposed to know the IDs, we don't need to have them on members.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -246,6 +257,89 @@ public static void checkSerializable(Object object, String objectName) {
...
+        }
+    }
+
+    /**
+     * Creates a ZIP-file stream from the directory tree rooted at the supplied
+     * {@code baseDir}. Pipes the stream into the provided output stream, closing
+     * it when done.
+     * <p>
+     * <strong>Note:</strong> files and directories starting with either {@code _}
+     * or {@code .} are ignored.
```
viliam-durina: - This should be documented at `JobConfig.attachDirectory`, not here. 
- It is a non-portable way to detect hidden files, it's valid on macOS only i guess (never heard about the `_` prefix). You should use `Files.isHidden(path)`.
- You should also ignore hidden files, not just hidden directories.
- It might be easier to use `Files.find`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -182,15 +192,23 @@ long uploadJobResources(JobConfig jobConfig) {
...
-                        InputStream in = rc.getUrl().openStream();
-                        readStreamAndPutCompressedToMap(rc.getId(), tmpMap, in);
-                        in.close();
+                    case CLASS:
+                        try (InputStream in = rc.getUrl().openStream()) {
+                            readStreamAndPutCompressedToMap(rc.getId(), tmpMap, in);
+                        }
+                        break;
+                    case FILE:
+                        IMapOutputStream os = new IMapOutputStream(getJobFileStorage(jobId).get(), rc.getId());
```
viliam-durina: This is broken, we get a lazy concurrent supplier each time and immediately do a `get()` on it. The `getJobFileStorage(jobId)` should be stored outside of the loop. And it doesn't have to be concurrent.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -458,6 +481,13 @@ public JobExecutionRecord getJobExecutionRecord(long jobId) {
         return Util.memoizeConcurrent(() -> instance.getMap(RESOURCES_MAP_NAME_PREFIX + idToString(jobId)));
     }
 
+    /**
+     * Gets the job files storage map, lazily evaluated to avoid creating the map if it won't be needed
+     */
+    Supplier<IMap<String, byte[]>> getJobFileStorage(long jobId) {
+        return Util.memoizeConcurrent(() -> instance.getMap(jobFileStorageMapName(jobId)));
```
viliam-durina: We should make it lazy at caller, not here. Here we can get the map unconditionally.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/JetClassLoader.java
```diff
@@ -76,28 +66,6 @@ public JetClassLoader(@Nonnull NodeEngine nodeEngine,
         return defineClass(name, classBytes, 0, classBytes.length);
     }
 
-    @Override
-    protected URL findResource(String name) {
```
viliam-durina: Why was it necessary to remove this method? What if there's a jar on classpath and it wants to read some xml file on its classpath?

hazelcast-jet-core/src/test/java/com/hazelcast/jet/config/ResourceConfigTest.java
```diff
@@ -0,0 +1,327 @@
...
+        config.attachFile(path, resourceName);
+        ResourceConfig resourceConfig = config.getResourceConfigs().iterator().next();
+
+        assertEquals(resourceName, resourceConfig.getId());
+        assertEquals(ResourceType.FILE, resourceConfig.getResourceType());
+        assertEquals(new File(path).toURI().toURL(), resourceConfig.getUrl());
+    }
+
+    @Test
+    public void testAttachFile_with_File() throws Exception {
```
viliam-durina: This test is exactly the same as `JobConfigTest.when_addResourceWithFile_thenReturnsResourceConfig`. Others probably too.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
viliam-durina: It's worse, it actually fails on Windows with:
```
java.nio.file.InvalidPathException: Illegal char <:> at index 2: /C:/work/hazelcast/hazelcast-jet/hazelcast-jet-core/target/test-classes/deployment/resource.txt

	at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
	at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:229)
	at java.base/java.nio.file.Path.of(Path.java:147)
	at java.base/java.nio.file.Paths.get(Paths.java:69)
	at com.hazelcast.jet.impl.JobRepository.uploadJobResources(JobRepository.java:202)
	at com.hazelcast.jet.impl.AbstractJetInstance.uploadResourcesAndAssignId(AbstractJetInstance.java:149)
	at com.hazelcast.jet.impl.AbstractJetInstance.newJob(AbstractJetInstance.java:57)
```
We must use `URL.openStream()`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
cangencer: I've partially fixed this. Javadoc still needs updating.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -97,6 +102,11 @@
      */
     public static final String RESOURCES_MAP_NAME_PREFIX = INTERNAL_JET_OBJECTS_PREFIX + "resources.";
 
+    /**
+     * Name of internal IMap which stores attached job files.
+     */
+    public static final String FILE_STORAGE_MAP_NAME_PREFIX = INTERNAL_JET_OBJECTS_PREFIX + "files.";
```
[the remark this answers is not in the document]
cangencer: we can do this using a namespace as part of the key (i.e. class.<id> or file.<id>) so there's no interference between CL and file resources. WDYT?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/deployment/IMapOutputStream.java
viliam-durina: Got it, I was wrong. It writes an entry and then overwrites that entry with more data. Hopefully, nobody flushes mid-way of writing. But if somebody flushes just before closing, the final entry will be `put` twice.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/Contexts.java
```diff
@@ -108,26 +131,76 @@ public ILogger logger() {
...
 
         @Override
         public int memberIndex() {
             return memberIndex;
         }
+
+        @Nonnull @Override
+        public File attachedDirectory(@Nonnull String id) {
+            Preconditions.checkHasText(id, "id cannot be null or empty");
+            return tempDirectories.computeIfAbsent(id, x -> extractFileToDisk(id));
```
eminn: `computIfAbsent` method doesn't guarantee that the value object generated by the mapping function will end up as a value in the map. It uses `putIfAbsent` internally and if that fails it fetches the actual value from the map. The value object created with the mapping function will be left on the disk untracked.

cangencer: That's not true, CHM has a special implementation of `computeIfAbsent` which is atomic.

eminn: I've missed that, I was looking at the default implementation on the ConcurrentMap. Good to know that.

viliam-durina: We should then use CHM for the field, not just CM.

PR #1825, opened 2019-12-23
Make job scanning delay adaptable

cangencer:
There was a 5 seconds delay after starting a node
that the submitted jobs would be running on it. By having
a shorter delay when partitions are not yet initialized, we
can start the job quicker on initial submission.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
@@ -22,6 +22,7 @@ import com.hazelcast.core.HazelcastInstanceNotActiveException;
 import com.hazelcast.instance.impl.Node;
 import com.hazelcast.internal.cluster.ClusterService;
 import com.hazelcast.internal.partition.impl.InternalPartitionServiceImpl;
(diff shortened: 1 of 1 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -868,28 +880,40 @@ private InternalPartitionServiceImpl getInternalPartitionService() {
...
+    }
+
+    private void doScanJobs() {
+        Collection<JobRecord> jobs = jobRepository.getJobRecords();
+        for (JobRecord jobRecord : jobs) {
+            JobExecutionRecord jobExecutionRecord = ensureExecutionRecord(jobRecord.getJobId(),
+                    jobRepository.getJobExecutionRecord(jobRecord.getJobId()));
+            startJobIfNotStartedOrCompleted(jobRecord, jobExecutionRecord, "discovered by scanning of JobRecords");
+        }
+        jobRepository.cleanup(nodeEngine);
```
cangencer: I will make a note that we should perhaps move job cleanup to another scheduled task. If some exception would happen above, jobs would not be cleaned up.

PR #1830, opened 2019-12-25
Enhance Hadoop JARs and Classpath chapter in reference manual

gurbuzali:
Describes which JARs to put to classpath in order to use Hadoop module


cangencer:
I remember also using this command:

https://hadoop.apache.org/docs/r2.10.0/hadoop-project-dist/hadoop-common/CommandsManual.html#classpath

We should mention this in the manual, perhaps?

PR #1831, opened 2019-12-26
Add random group.id if not set explicitly

### LINKED ISSUE #1705: Kafka Source without group.id throws exception
When using Kafka Source and no processing guarantees, the source attempts to commit offsets to the broker after each batch. In this case you get the following exception in newer Kafka versions:

```
Caused by: com.hazelcast.jet.JetException: Exception in
ProcessorTasklet{Query1/streamKafka(trades)#1}: 
org.apache.kafka.common.errors.InvalidGroupIdException: 
To use the group management or offset commit APIs, you must provide a valid group.id in the consumer configuration.
```

We should either generate a group id if it's missing with this setup or explain inside the Javadoc

gurbuzali:
`group.id` is used to commit offsets if processing guarantee is none, ignored otherwise. We set it to a random UUID unless explicitly set by user.

Fixes #1705


viliam-durina:
Comitting to random `group.id` is a pure waste I think. Shouldn't we rather avoid committing offests in that case?

PR #1833, opened 2019-12-26
Initialize the context object created by createContextFn of ServiceFactory with configured ManagedContext

### LINKED ISSUE #1642: ContextFactory.createFn should support `@SpringAware`
The context created by ContextFactory.createFn should be initialized using `ManagedContext`,  similar to how processors are initialized with it.

gurbuzali:
Fixes #1642


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/ProcessorSupplierWithService.java
```diff
@@ -50,7 +52,12 @@ private ProcessorSupplierWithService(
 
     @Override
     public void init(@Nonnull ProcessorSupplier.Context context) {
-        this.serviceContext = serviceFactory.createContextFn().apply(context);
+        HazelcastInstanceImpl hazelcastInstance = (HazelcastInstanceImpl) context.jetInstance().getHazelcastInstance();
```
jerrinot: I think it's better to cast to `SerializationServiceSupport`. this will work with both `HazelcastInstanceImpl` and `HazelcastInstanceProxy`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/ProcessorSupplierWithService.java
```diff
@@ -50,7 +52,12 @@ private ProcessorSupplierWithService(
...
+        ManagedContext managedContext = hazelcastInstance.getSerializationService().getManagedContext();
+        serviceContext = serviceFactory.createContextFn().apply(context);
+        if (managedContext != null) {
```
jerrinot: is this allowed to ever return null? I would expect a properly constructed instance to always return `HazelcastManagedContext`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/ProcessorSupplierWithService.java
jerrinot: alternatively the processor could implement the `SerializationServiceAware` interface. I played with that here: https://github.com/hazelcast/hazelcast-jet/pull/1834/files

it does not require casting at all, but I am not sure whether it's really any better.

gurbuzali: For the processor it is too late. we create a single `serviceContext` per member here and pass it to the processors so that they create a service using this context. So we want the `managedContext` to initialize the `serviceContext` and do it once, after that we can create services using the injected values in the `serviceContext`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/ProcessorSupplierWithService.java
jerrinot: I am thinking whether we should also apply the context to the function itself. 

it sound strange at first. but it allows you to provide e.g. `SpringServiceFactory` which will give you an arbitrary Spring bean.

jerrinot:
I played with this PR and it's pretty good. I implemented some Spring convenience on top of this: https://github.com/jerrinot/hazelcast-jet/commit/d1a9ac93098f35420f3f82d892eee581d0266980#diff-9850df76615637ff4f7fb5501717777eR42-R66

Feel free to include it in your PR.

hazelcast-jet-spring/src/main/java/com/hazelcast/jet/spring/JetSpringServices.java
```diff
@@ -0,0 +1,288 @@
...
+
+import javax.annotation.Nonnull;
+
+/**
+ * Utility class with methods that create several useful Spring Bean
+ * {@link ServiceFactory service factories} and transform functions.
+ *
+ * @since 4.0
+ */
+public final class JetSpringServices {
```
jerrinot: This would deserve an example. as the method signature can be intimidating, but actual usage is simple.

hazelcast-jet-spring/src/main/java/com/hazelcast/jet/spring/JetSpringServices.java
```diff
@@ -0,0 +1,288 @@
...
+     * @param beanName the name of the bean
+     * @param <T>      the type of the bean
+     */
+    public static <T> ServiceFactory<?, T> beanServiceFactory(@Nonnull String beanName) {
+        return ServiceFactory.withCreateContextFn(ctx -> new BeanExtractor())
+                             .withCreateServiceFn((c, g) -> g.getBean(beanName));
+    }
+
+    @SpringAware
+    private static final class BeanExtractor {
```
jerrinot: It just occurred to me all the `getBean()` methods should check that `context` is not null. 
If it's null then it means `SpringManagedContext` was not injected into Hazelcast `Config` and we should explain how to do it / point to the right documentation. Otherwise it's hard to figure this out on your own. 

Injected of the `SpringManagedContext` works out of the box only when you are using the Spring XML namespace support and "nobody" does that anymore. However it's more complex when you are using Spring Java Config.

gurbuzali:
verify

cangencer:
verify

hazelcast-jet-spring/src/main/java/com/hazelcast/jet/spring/JetSpringServices.java
```diff
@@ -0,0 +1,297 @@
...
+     * Returns a transform function for the {@link BatchStage#apply(FunctionEx)}
+     * which adds a mapping stage using the specified bean.
+     *
+     * @param beanName     the name of the bean
+     * @param requiredType the class of the bean
+     * @param mapper       the mapping function
+     * @param <T>          the type of the bean
+     * @param <R>          the type of the returned stage
+     */
+    public static <I, R, T> FunctionEx<BatchStage<I>, BatchStage<R>> mapBatchUsingSpringBean(
```
cangencer: I don't think we need these variants. They don't bring much convenience, and there's too many variations. The plain one using ServiceFactory is good enough and quite convenient. What do you think @jerrinot ?

hazelcast-jet-spring/src/test/java/com/hazelcast/jet/spring/RiskCalculator.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.jet.spring;
+
+public interface RiskCalculator {
```
cangencer: let's just call this Calculator and have methods like `add` or `multiply` because the sample used here makes no sense whatsoever

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/ProcessorSupplierWithService.java
```diff
@@ -50,7 +52,10 @@ private ProcessorSupplierWithService(
 
     @Override
     public void init(@Nonnull ProcessorSupplier.Context context) {
-        this.serviceContext = serviceFactory.createContextFn().apply(context);
+        HazelcastInstanceImpl hazelcastInstance = (HazelcastInstanceImpl) context.jetInstance().getHazelcastInstance();
+        ManagedContext managedContext = hazelcastInstance.getSerializationService().getManagedContext();
+        serviceContext = serviceFactory.createContextFn().apply(context);
+        serviceContext = (C) managedContext.initialize(serviceContext);
```
cangencer: do we have any guarantee that always same type is returned? should we throw a more meaningful exception if it's not the case?

hazelcast-jet-spring/src/main/java/com/hazelcast/jet/spring/JetSpringServices.java
```diff
@@ -0,0 +1,297 @@
...
+
+import javax.annotation.Nonnull;
+
+/**
+ * Utility class with methods that create several useful Spring Bean
+ * {@link ServiceFactory service factories} and transform functions.
+ *
+ * @since 4.0
+ */
+public final class JetSpringServices {
```
cangencer: `JetSpringServiceFactories` would be more inline with the other naming (`ServiceFactories` )

hazelcast-jet-spring/src/main/java/com/hazelcast/jet/spring/JetSpringServices.java
```diff
@@ -0,0 +1,297 @@
...
+                             .withCreateServiceFn((c, g) -> g.getBean(beanName, requiredType));
+    }
+
+    /**
+     * Returns a factory that provides the specified bean as the service object.
+     *
+     * @param requiredType the class of the bean
+     * @param <T>          the type of the bean
+     */
+    public static <T> ServiceFactory<?, T> beanServiceFactory(@Nonnull Class<T> requiredType) {
```
cangencer: javadoc is too basic. needs some sample and more detailed documentation on how this works

hazelcast-jet-spring/src/main/java/com/hazelcast/jet/spring/JetSpringServices.java
jerrinot: I really like to be able to write: `stage.apply(mapStreamUsingSpringBean("risk-calculator", RiskCalculator::calculateRisk))`

It reads nicely.

examples/spring/src/main/java/com/hazelcast/jet/examples/spring/BeanAsAServiceSample.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public static void main(String[] args) throws Exception {
+        ApplicationContext context = new AnnotationConfigApplicationContext(AppConfig.class);
+        JetInstance jet = context.getBean(JetInstance.class);
+
+        fillList(jet);
+
+        Pipeline pipeline = Pipeline.create();
+        pipeline.<String>readFrom(Sources.list(LIST_NAME))
+                .mapUsingService(JetSpringServiceFactories.beanServiceFactory(UserDao.class),
```
cangencer: how about just naming the method `bean` ? A bean is already somewhat equivalent to a service

hazelcast-jet-spring/src/main/java/com/hazelcast/jet/spring/JetSpringServiceFactories.java
```diff
@@ -0,0 +1,145 @@
...
+     * context object which autowires the {@link ApplicationContext}. The
+     * context object obtains the specified bean from {@link ApplicationContext}
+     * and returns it as a service.
+     * <p>
+     * Below is a sample usage which reads the names from a list and maps those
+     * names to {@code User} objects using the spring bean {@code userDao}.
+     * <pre>{@code
+     * Pipeline pipeline = Pipeline.create();
+     * pipeline.<String>readFrom(Sources.list(LIST_NAME))
+     *         .mapUsingService(JetSpringServiceFactories.beanServiceFactory(UserDao.class),
```
cangencer: these still refer to old name

hazelcast-jet-spring/src/test/resources/com/hazelcast/jet/spring/application-context-processor.xml
```diff
@@ -44,4 +44,5 @@
 
     <jet:map instance-ref="jet-instance" name="my-map" id="my-map-bean"/>
 
+    <bean name="risk-calculator" class="com.hazelcast.jet.spring.SillyCalculator" />
```
cangencer: this name doesn't look like it's referred anywhere?

gurbuzali:
verify

PR #1841, opened 2020-01-08
Make JobConfig.resourceConfigs private

viliam-durina:
The resourceConfigs are only usable and used on the JVM that submits the job to upload the resources. The URLs are often to local files. They are also never used afterwards in our code. Since their size can be significant and we store it also in JobResult, it's better to remove them.

List of breaking changes:
- marked `JobConfig.getResourceConfigs()` method and `ResourceConfig` and `ResourceType` classes private API

Checklist
- [x] Tags Set
- [x] Milestone Set


cangencer:
verify

PR #1842, opened 2020-01-08
Adapt to IMDG using UUID instead of Address for Invocations

jbartok:
Add description here

Checklist
- [ ] Tags Set
- [ ] Milestone Set
- [ ] Any breaking changes are documented

Fixes #NNNN

List of breaking changes:

* ..

### DIFF
--- a/hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetCommandLine.java
+++ b/hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetCommandLine.java
@@ -61,8 +61,6 @@ import java.time.LocalDateTime;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
-import java.util.Optional;
-import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.LockSupport;
 import java.util.function.Consumer;
@@ -408,7 +406,7 @@ public class JetCommandLine implements Runnable {
             JetClientInstanceImpl client = (JetClientInstanceImpl) jet;
             HazelcastClientInstanceImpl hazelcastClient = client.getHazelcastClient();
             ClientClusterService clientClusterService = hazelcastClient.getClientClusterService();
-            Member masterMember = clientClusterService.getMember(masterUuid(clientClusterService));
+            Member masterMember = clientClusterService.getMember(clientClusterService.getMasterMember().getUuid());
             MCClusterMetadata clusterMetadata = FutureUtil.getValue(hazelcastClient.getManagementCenterService()
                                                                                    .getClusterMetadata(masterMember));
             Cluster cluster = client.getCluster();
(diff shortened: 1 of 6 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetClientInstanceImpl.java
```diff
@@ -151,9 +151,9 @@ public ILogger getLogger() {
         }
     }
 
-    private static Address masterAddress(Cluster cluster) {
+    private static UUID masterUUID(Cluster cluster) {
```
cangencer: can you make this a Util method? it's repeated several times.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobProxy.java
```diff
@@ -142,6 +145,11 @@ protected JobConfig doGetJobConfig() {
     }
 
     @Override
+    protected UUID masterUuid() {
+        Optional<Member> first = container().getClusterService().getMembers().stream().findFirst();
```
sancar: We have a getMasterMember on the ClusterService method.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/ClientJobProxy.java
```diff
@@ -156,16 +155,15 @@ protected long doGetJobSubmissionTime() {
...
             return serializationService().toObject(data);
         });
     }
 
     @Override
-    protected Address masterAddress() {
-        Optional<Member> first = container().getCluster().getMembers().stream().findFirst();
-        return first.orElseThrow(() -> new IllegalStateException("No members found in cluster")).getAddress();
+    protected UUID masterUuid() {
+        return Util.getMaster(container().getCluster()).getUuid();
```
cangencer: you can use use clientClusterService.getMaster() here and inline the Util method

jbartok: Ok, but we should be aware that we are changing behaviour, not just adapting (ie. no more exception). Is that ok?

hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetCommandLine.java
```diff
@@ -406,7 +406,7 @@ public void cluster(
             JetClientInstanceImpl client = (JetClientInstanceImpl) jet;
             HazelcastClientInstanceImpl hazelcastClient = client.getHazelcastClient();
             ClientClusterService clientClusterService = hazelcastClient.getClientClusterService();
-            Member masterMember = clientClusterService.getMember(clientClusterService.getMasterAddress());
+            Member masterMember = clientClusterService.getMember(clientClusterService.getMasterMember().getUuid());
```
sancar: To get master member, we get the master member and get its uuid and then get the master member again ?
@jbartok

PR #1852, opened 2020-01-10
Remove cloning of iteration pointers

cangencer:
It's now cloned as part of the Operation. see also: https://github.com/hazelcast/hazelcast/pull/16464


cangencer:
verify

PR #1857, opened 2020-01-16
[ProjectX] Do not depend on HZ snapshot

olukas:
This PR adds tooling for not to depend on latest HZ IMDG snapshot but rather on some particular version.

Checklist
- [ ] Tags Set
- [ ] Milestone Set
- [ ] Any breaking changes are documented

### DIFF
--- a/update-snapshot-version.sh
+++ b/update-snapshot-version.sh
@@ -1,0 +1,101 @@
+#!/bin/bash
+
+JET_EE_LOCAL_PATH=$1
+HAZELCAST_SNAPSHOT_VERSION=4.0-SNAPSHOT
+USER_GIT_REMOTE_REPO_NAME=origin
+MVN_REPOSITORY=~/.m2/repository
+ORIGINAL_PATH=`pwd`
+
+# get latest version of Hazelcast snapshot
+echo "Starting get latest version of Hazelcast snapshot section"
+mvn dependency:get -DgroupId=com.hazelcast -DartifactId=hazelcast -Dversion=${HAZELCAST_SNAPSHOT_VERSION} -Dpackaging=jar -Dmaven.repo.local=${MVN_REPOSITORY}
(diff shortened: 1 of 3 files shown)


update-snapshot-version.sh
```diff
@@ -0,0 +1,101 @@
...
+git checkout -b ${BRANCH_NAME}
+
+# change version
+echo "Starting change version section"
+sed -i "s#<hazelcast.version>.*</hazelcast.version>#<hazelcast.version>${HAZELCAST_LATEST_VERSION}</hazelcast.version>#" pom.xml
+sed -i "s#<hazelcast.ee.version>.*</hazelcast.ee.version>#<hazelcast.ee.version>${HAZELCAST_EE_LATEST_VERSION}</hazelcast.ee.version>#" pom.xml
+
+# try to compile
+echo "Starting try to compile section"
+mvn clean install -DskipTests -Dmaven.repo.local=${MVN_REPOSITORY}
```
viliam-durina: I would leave this for the PR builder, no need to wait for it locally...

update-snapshot-version.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
+
+JET_EE_LOCAL_PATH=$1
```
viliam-durina: We should fail if parameter is missing

update-snapshot-version.sh
```diff
@@ -0,0 +1,101 @@
...
+HAZELCAST_SNAPSHOT_VERSION=4.0-SNAPSHOT
+USER_GIT_REMOTE_REPO_NAME=origin
+MVN_REPOSITORY=~/.m2/repository
+ORIGINAL_PATH=`pwd`
```
viliam-durina: `dirname $0` works also if we're not running the script from its directory

olukas:
@viliam-durina it's still WIP version. There are a lot of things which need to be improved etc. Since 4.0 is almost done I plan to improve it after release and start to use it for 4.1. (It will also cause some changes in release scripts etc.)

viliam-durina:
We'll depend on the released `4.0` for as long as we can...

Issue #1851, opened 2020-01-10
IOUtilTest.test_zipAndUnzipNestedFolder_then_contentsShouldBeSame


gurbuzali:
failed on _master_ PR builder

Failed on http://jenkins.hazelcast.com/job/jet-oss-pr-builder/1846/testReport/junit/com.hazelcast.jet.impl.util/IOUtilTest/test_zipAndUnzipNestedFolder_then_contentsShouldBeSame/

Stacktrace:
```
java.lang.AssertionError: expected:<7> but was:<6>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.hazelcast.test.HazelcastTestSupport.assertCollection(HazelcastTestSupport.java:180)
	at com.hazelcast.jet.impl.util.IOUtilTest.test(IOUtilTest.java:71)
	at com.hazelcast.jet.impl.util.IOUtilTest.test_zipAndUnzipNestedFolder_then_contentsShouldBeSame(IOUtilTest.java:48)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:114)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:106)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.lang.Thread.run(Thread.java:748)
```

Standard output:
```
[...]

PR #1858, opened 2020-01-20
Print collection contents in case of failure

gurbuzali:
Checklist
- [x] Tags Set
- [x] Milestone Set

related to #1851

### DIFF
--- a/hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/util/IOUtilTest.java
+++ b/hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/util/IOUtilTest.java
@@ -35,7 +35,6 @@ import java.util.stream.Collectors;
 
 import static com.hazelcast.jet.impl.util.IOUtil.packDirectoryIntoZip;
 import static com.hazelcast.jet.impl.util.IOUtil.unzip;
-import static org.junit.Assert.fail;
 
 @RunWith(HazelcastParallelClassRunner.class)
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/util/IOUtilTest.java
```diff
@@ -68,6 +69,9 @@ public void test(Path originalPath) throws IOException {
         Set<Path> originalSet = getChildrenFromRoot(originalPath);
         Set<Path> unzippedSet = getChildrenFromRoot(unzippedPath);
 
+        if (originalSet.size() != unzippedSet.size()) {
```
eminn: Can we add this to assertCollection method ?

gurbuzali: you mean to have it on IMDG side or overwrite it on our side ?

eminn: I thought it comes from JetAssert but still we can improve the one in IMDG to print contents when the size is not equal

gurbuzali: I've put it to `JetTestSupport` for now, after the change on IMDG side we can remove it

Issue #1855, opened 2020-01-14
com.hazelcast.jet.pipeline.test.TestSourcesTest.test_itemStream_withWindowing[1: mode=client]


olukas:
_3.2-maintenance_ (commit 31e3f9d51d1279665c3e94a7ba9ea608d9a41282)

Failed on Zulu JDK 11: http://jenkins.hazelcast.com/job/jet-oss-maintenance-zulu-jdk11/87/testReport/com.hazelcast.jet.pipeline.test/TestSourcesTest/test_itemStream_withWindowing_1__mode_client_/

Stacktrace:
```
java.lang.AssertionError: 

Expected: exception with message a string containing "com.hazelcast.jet.pipeline.test.AssertionCompletedException"
     but: message was "Exception in ProcessorTasklet{assertCollectedEventually#0}: java.lang.AssertionError: java.lang.AssertionError: Did not find any window with 10 items: [WindowResult{start=23:03:51.000, end=23:03:52.000, value='8', isEarly=false}, WindowResult{start=23:03:52.000, end=23:03:53.000, value='11', isEarly=false}, WindowResult{start=23:03:53.000, end=23:03:54.000, value='9', isEarly=false}, WindowResult{start=23:03:54.000, end=23:03:55.000, value='11', isEarly=false}, WindowResult{start=23:03:55.000, end=23:03:56.000, value='9', isEarly=false}, WindowResult{start=23:03:56.000, end=23:03:57.000, value='11', isEarly=false}, WindowResult{start=23:03:57.000, end=23:03:58.000, value='9', isEarly=false}, WindowResult{start=23:03:58.000, end=23:03:59.000, value='11', isEarly=false}, WindowResult{start=23:03:59.000, end=23:04:00.000, value='9', isEarly=false}, WindowResult{start=23:04:00.000, end=23:04:01.000, value='11', isEarly=false}]"
[...]

PR #1859, opened 2020-01-20
Make TestSourceTest more tolerant to hiccups

jbartok:
Fixes issue https://github.com/hazelcast/hazelcast-jet/issues/1855

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/test/TestSourcesTest.java
+++ b/hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/test/TestSourcesTest.java
@@ -75,8 +75,8 @@ public class TestSourcesTest extends PipelineTestSupport {
          .window(WindowDefinition.tumbling(1000))
          .aggregate(AggregateOperations.counting())
          .apply(assertCollectedEventually(60, windowResults -> {
-             //look at last 5 windows at most
-             int windowsToConsider = Math.min(5, windowResults.size());
+             //look at last 5 windows at most, always ignore first
+             int windowsToConsider = Math.min(5, Math.max(windowResults.size() - 1, 0));
 
              //count the total no. of items emitted in those windows
              int totalItems = windowResults.stream()
--- a/hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/test/TestSourcesTest.java
+++ b/hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/test/TestSourcesTest.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.jet.pipeline.test;
 
+import com.google.common.collect.Range;
 import com.hazelcast.jet.aggregate.AggregateOperations;
 import com.hazelcast.jet.pipeline.PipelineTestSupport;
 import com.hazelcast.jet.pipeline.WindowDefinition;
@@ -75,9 +76,11 @@ public class TestSourcesTest extends PipelineTestSupport {
          .window(WindowDefinition.tumbling(1000))
          .aggregate(AggregateOperations.counting())
          .apply(assertCollectedEventually(10, windowResults -> {
-             // find any window that has 10 items, some may be incomplete due to hiccups
-             boolean matched = windowResults.stream().anyMatch(r -> r.result() == itemsPerSecond);
-             assertTrue("Did not find any window with 10 items: " + windowResults, matched);
+             Range<Integer> range = Range.closed(itemsPerSecond - 1, itemsPerSecond + 1);
+             int total = windowResults.size();
+             int matched = (int) windowResults.stream().filter(r -> range.contains(r.result().intValue())).count();
+             assertTrue(String.format("Did not find enough good windows, only %d out of %d: %s",
+                     matched, total, windowResults), matched >= 0.5d * total);
          }));
 
         expectedException.expectMessage(AssertionCompletedException.class.getName());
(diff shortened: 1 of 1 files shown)


jbartok:
verify

hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/test/TestSourcesTest.java
```diff
@@ -75,9 +76,11 @@ public void test_itemStream_withWindowing() throws Throwable {
          .window(WindowDefinition.tumbling(1000))
          .aggregate(AggregateOperations.counting())
          .apply(assertCollectedEventually(10, windowResults -> {
-             // find any window that has 10 items, some may be incomplete due to hiccups
-             boolean matched = windowResults.stream().anyMatch(r -> r.result() == itemsPerSecond);
-             assertTrue("Did not find any window with 10 items: " + windowResults, matched);
+             Range<Integer> range = Range.closed(itemsPerSecond - 1, itemsPerSecond + 1);
```
cangencer: I'd prefer not to depend on Guava for tests, you can rewrite this easily using `IntStream.range()`

jbartok: Ok

jbartok:
verify

hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/test/TestSourcesTest.java
```diff
@@ -74,10 +74,24 @@ public void test_itemStream_withWindowing() throws Throwable {
          .withNativeTimestamps(0)
          .window(WindowDefinition.tumbling(1000))
          .aggregate(AggregateOperations.counting())
-         .apply(assertCollectedEventually(10, windowResults -> {
-             // find any window that has 10 items, some may be incomplete due to hiccups
-             boolean matched = windowResults.stream().anyMatch(r -> r.result() == itemsPerSecond);
-             assertTrue("Did not find any window with 10 items: " + windowResults, matched);
+         .apply(assertCollectedEventually(60, windowResults -> {
+             //look at last 5 windows at most
+             int windowsToConsider = Math.min(5, windowResults.size());
```
cangencer: you can probably look at fewer windows so that the test doesn't take too long.

cangencer: should be Math.max

hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/test/TestSourcesTest.java
```diff
@@ -74,10 +74,24 @@ public void test_itemStream_withWindowing() throws Throwable {
...
+
+             //count the total no. of items emitted in those windows
+             int totalItems = windowResults.stream()
+                     .skip(windowResults.size() - windowsToConsider)
```
cangencer: this will throw `IllegalArgumentException` because you can get a negative number here

hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/test/TestSourcesTest.java
cangencer: now if just 1 window has 9 items, the test will pass. That's not a very strong assertion.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/test/TestSourcesTest.java
jbartok: I don't see how I could get a negative number there, until `windowsToConsider = Math.min(5, windowResults.size())`.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/test/TestSourcesTest.java
jbartok: If there is 1 window with 9 items, then `deviationFromTarget = 1`, which is not smaller than the allowed tolerance of `0.01`, so it won't pass, but wait for more windows.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/test/TestSourcesTest.java
cangencer: yes, you are right, my bad

hazelcast-jet-core/src/test/java/com/hazelcast/jet/pipeline/test/TestSourcesTest.java
cangencer: ok, again I read it thinking it was a percentage, but it's actually an absolute value.

cangencer: how long does the test take to pass?

jbartok: On my machine it always takes 5 seconds to pass, but unfortunately it's not the test that's bad, but the first window always has way less values than it should, usually at most 7, instead of 10. 

And that usually only in the "member" mode, the subsequent "client" mode is better, usually starts with a 9. 

Feels like a warmup, but I'm not sure what it is. Looking at it.

cangencer: how about we always skip the first window?

jbartok: Good idea, pushed that version, much better.

PR #1861, opened 2020-01-20
Execute evictFn after completion

### LINKED ISSUE #1801: StreamStageTest.mapStateful_withEvictFnReturningNull failure
Happened in a [PR build](http://jenkins.hazelcast.com/job/jet-oss-pr-builder/1584/testReport/junit/com.hazelcast.jet.pipeline/StreamStageTest/mapStateful_withEvictFnReturningNull_0__mode_member_/), but it reproduces on master too. The test is racy. It uses 2 members and a non-distributed source. But such kind of source uses a no-op processor on one member, but the no-op processor still has to run and return `true` from `complete`. And if the non-no-op processor completes before the no-op one, thanks to WM coalescing the evict action doesn't execute.

We can consider running the onEvict function for all items in `complete`, that would make it deterministic for batch jobs.

Stacktrace:
```
org.junit.ComparisonFailure: expected:<[1 evicted]> but was:<[]>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at com.hazelcast.jet.pipeline.StreamStageTest.mapStateful_withEvictFnReturningNull(StreamStageTest.java:700)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:114)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:106)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
```

You can reproduce it by inserting a `complete` method that sleeps for 1 second before returning true to the processor in `ProcessorMetaSupplier.forceTotalParallelismOne`

viliam-durina:
The aim is to make batch jobs more deterministic: all items are evicted at the
end, regardless of the WM.

Fixes #1801

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/TransformStatefulP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/TransformStatefulP.java
@@ -168,6 +168,8 @@ public class TransformStatefulP<T, K, S, R> extends AbstractProcessor {
     @Override
     public boolean saveToSnapshot() {
         if (inComplete) {
+            // If we are in completing phase, we can have a half-emitted item. Instead of finishing it and
+            // writing a snapshot, we finish the final items and save no state.
             return complete();
         }
         if (snapshotTraverser == null) {
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/TransformStatefulP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/TransformStatefulP.java
@@ -48,6 +48,7 @@ import static java.lang.Math.min;
 public class TransformStatefulP<T, K, S, R> extends AbstractProcessor {
     private static final int HASH_MAP_INITIAL_CAPACITY = 16;
     private static final float HASH_MAP_LOAD_FACTOR = 0.75f;
+    private static final Watermark FLUSHING_WATERMARK = new Watermark(Long.MAX_VALUE);
 
     @Probe(name = "lateEventsDropped")
     private final Counter lateEventsDropped = SwCounter.newSwCounter();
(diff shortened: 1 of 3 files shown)


viliam-durina:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/TransformStatefulP.java
```diff
@@ -154,6 +167,9 @@ void reset(Watermark wm) {
 
     @Override
     public boolean saveToSnapshot() {
+        if (inComplete) {
```
cangencer: why do we need this? we don't this for other stateful processors (i.e. windowing)

cangencer: this needs more explanation, like in SessionWindowP. Do we perhaps need better support for this in ProcessorTasklet?

PR #1864, opened 2020-01-20
Don't call Processor.close on cooperative thread

### LINKED ISSUE #1793: Don't call Processor.close on cooperative threads

viliam-durina:
Fixes #1793

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
@@ -51,6 +51,8 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Queue;
 import java.util.TreeMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicLongArray;
 import java.util.function.Predicate;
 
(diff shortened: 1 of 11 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
```diff
@@ -343,11 +344,29 @@ private void stateMachineStep() {
...
+                    progTracker.madeProgress();
+                    state = CLOSE;
+                    stateMachineStep();
                 }
                 return;
 
+            case CLOSE:
+                if (isCooperative()) {
+                    if (closeFuture == null) {
+                        closeFuture = executionService.submit(this::closeProcessor);
```
cangencer: I wonder if this will slow down job completion, since most processors will have a noop close implementation?

mtopolnik: Maybe we can have a no-op constant that we recognize and short-circuit this?

viliam-durina: @cangencer Using JMH submitting 100 no-op tasks to ForkJoinPool.commonPool and waiting for their completion by checking future.isDone takes on average 300μs. So it's not that much pain.

@mtopolnik It's not a lambda, but a method we call. We can check if the `declaringClass` of `close` method is the `Processor` class:
```java
if (p.getClass().getMethod("close").getDeclaringClass() == Processor.class)
    // ignore the close method
```
I benchmarked this too. This check took 1.3μs, while directly executing a no-op runnable took 1.7μs. I think it's not worth the complexity, we're not on a hot path.

PR #1865, opened 2020-01-20
Add a type check for the initilized object by ManagedContext

gurbuzali:
The initialized object can have a different type than the original object passed to the `ManagedContext`. We check and throw exception if so.

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
@@ -208,7 +208,7 @@ public class ProcessorTasklet implements Tasklet {
                 toInit = (Processor) initialized;
             } catch (ClassCastException e) {
                 throw new IllegalArgumentException(String.format(
-                        "The initialized object(%s) should be an instance of %s", initialized, Processor.class));
+                        "The initialized object(%s) should be an instance of %s", initialized, Processor.class), e);
             }
             if (processor instanceof ProcessorWrapper) {
                 ((ProcessorWrapper) processor).setWrapped(toInit);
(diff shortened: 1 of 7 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
```diff
@@ -197,10 +198,11 @@ private OutboxImpl createOutbox(@Nonnull OutboundCollector ssCollector) {
 
     @Override
     public void init() {
-        if (serializationService.getManagedContext() != null) {
+        ManagedContext managedContext = serializationService.getManagedContext();
+        if (managedContext != null) {
             Processor toInit = processor instanceof ProcessorWrapper
                     ? ((ProcessorWrapper) processor).getWrapped() : processor;
-            Object initialized = serializationService.getManagedContext().initialize(toInit);
+            Object initialized = managedContext.initialize(toInit);
```
cangencer: shouldn't use the Util method here too?

gurbuzali: I guess we want to have object equality here, because we may be using the processor elsewhere ?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ProcessorTasklet.java
```diff
@@ -197,11 +198,23 @@ private OutboxImpl createOutbox(@Nonnull OutboundCollector ssCollector) {
...
             Processor toInit = processor instanceof ProcessorWrapper
                     ? ((ProcessorWrapper) processor).getWrapped() : processor;
-            Object initialized = serializationService.getManagedContext().initialize(toInit);
-            assert initialized == toInit : "different object returned";
+            Object initialized = null;
+            try {
+                initialized = managedContext.initialize(toInit);
+                toInit = (Processor) initialized;
+            } catch (ClassCastException e) {
+                throw new IllegalArgumentException(String.format(
```
cangencer: do we want to pass the cause here, just in case?

PR #1868, opened 2020-01-21
Improve return-results-to-caller related sections of RefMan

jbartok:
Improve return-results-to-caller related sections of RefMan

Checklist
- [x] Tags Set
- [x] Milestone Set


cangencer:
why is it tagged as #breaking-changes-core?

jbartok:
> why is it tagged as #breaking-change-core?

By accident, fixed.

PR #1870, opened 2020-01-21
Call a Python Function from Jet Pipeline

mtopolnik:
This PR adds a service factory that takes a directory with Python code and calls a function in that code to transform the data from the Jet pipeline. In the first iteration, we support only string data.

The `mapUsingPython` stage launches one or more Python processes (number determined by the stage's local parallelism), each of which starts a gRPC server, and for each of them there is a matching Jet processor that connects to it. Communication happens within a single gRPC request with bidirectional streaming. Each gRPC message is a list of strings and the list is formed through natural batching within the Jet processor, by picking up all the items available in the `Inbox`.

You can supply either a single Python file that declares the handling function, or a complete directory with arbitrary files, some of which have special meaning to Jet. You must explicitly name the Python module and function that Jet will call to transform the data. In addition, Jet looks for these optional files:

* `requirements.txt` is assumed to list the [dependencies of your Python code](https://pip.pypa.io/en/stable/user_guide/#requirements-files). Jet will automatically install them. You can also install the modules to the Jet servers' global Python environment in order to speed up job initialization. Jet reuses the global modules and adds the missing ones to the job-local virtual environment.
* `init.sh`  is assumed to be a bash script that Jet will run before the job starts
* `cleanup.sh` is assumed to be a bash script that Jet will run when the job is completing

----

Adds a `GeneralStage.mapUsingServiceAsyncBatched` transform that is to be used with the new service factory.

Adds the ability to attach files to `ServiceFactory`. These will be picked up at job submission time and added to the `JobConfig`. This avoids the burden on the user to both declare file IDs in the service factory and add the actual files to `JobConfig`.

Adds the (internal) ability to attach files to `PipelineImpl`. This is needed as an intermediate step to let the `ServiceFactory` propagate its attachments to `JobConfig`.

Changes behavior of `JobConfig.attachFile` to use the actual filename instead of the supplied ID, which may not be a filename-compatible string.

Adds fail-fast checks to `JobConfig.add/attachX` methods for the existence of files/directories being referred to. URLs are not checked, but are validate to include at least one path segment (it is used as the filename).

These items are still missing to complete this work:

- reference manual additions
- a code sample

They will be added in a followup PR.

Checklist
- [x] Tags Set
- [x] Milestone Set

Breaking changes:

* `Processors.flatMapUsingServiceAsyncP`: added missing `? extends` to `flatMapAsyncFn`


cangencer:
@viliam-durina could you test this on windows?

mtopolnik:
No need to test on windows, it will definitely not work :) Unless you can test it within a Unix-like environment on windows, one that understands Bash.

cangencer:
right, I forgot about the init scripts. Maybe for the next version, we could add windows support, or rewrite those scripts as python scripts so it's cross platform.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/ServiceFactory.java
```diff
@@ -102,37 +106,31 @@
...
-            @Nonnull ConsumerEx<? super S> destroyServiceFn,
-            @Nonnull ConsumerEx<? super C> destroyContextFn,
-            boolean isCooperative,
-            int maxPendingCallsPerProcessor,
-            boolean orderedAsyncResponses
-    ) {
+    private ConsumerEx<? super C> destroyContextFn = ConsumerEx.noop();
+
+    @Nonnull
+    private Set<File> filesToAttach = new HashSet<>();
```
cangencer: this would get serialized

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/ComputeStageImplBase.java
```diff
@@ -223,6 +223,7 @@
             @Nonnull BiFunctionEx<? super S, ? super T, ? extends R> mapFn
     ) {
         checkSerializable(mapFn, "mapFn");
+        pipelineImpl.attachFiles(serviceFactory.attachedFiles());
```
cangencer: we should do this another way - it's easy to miss in one place to add the files, and you don't have coverage for all the options. Maybe Pipeline.attachFiles() or stage.addFiles() ?

hazelcast-jet-python/src/main/java/com/hazelcast/jet/python/PythonServiceConfig.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.List;
+
+/**
+ * Configuration object for the Python service factory.
+ */
+public class PythonServiceConfig implements Serializable {
+    private File baseDir;
+    private File handlerFile;
+    private String handlerModule;
+    private String handlerFunction;
```
cangencer: I wonder if it's better to represent these internally as `Properties` class or similar, so we don't have to add a method for every single option.

hazelcast-jet-python/src/main/java/com/hazelcast/jet/python/PythonServiceConfig.java
```diff
@@ -0,0 +1,197 @@
...
+        return handlerFunction;
+    }
+
+    /**
+     * Sets the name of the Python function that transforms Jet pipeline data,
+     * defined in the module you configured with {@link #setHandlerModule}.
+     * The function must take a single argument that is a list of strings, and
+     * return another list of strings which has the results of transforming each
+     * item in the input list. There must be a strict one-to-one match between
+     * the input and output lists.
```
cangencer: we should include an example python function here

hazelcast-jet-python/src/test/java/com/hazelcast/jet/python/PythonServiceTest.java
```diff
@@ -0,0 +1,147 @@
...
+    private File baseDir;
+
+    @BeforeClass
+    public static void beforeClass() throws Exception {
+        initialize(2, null);
+    }
+
+    @Before
+    public void before() throws Exception {
+        baseDir = createTempDirectory();
```
cangencer: we should delete this after the tests

hazelcast-jet-python/src/main/java/com/hazelcast/jet/python/PythonServiceConfig.java
mtopolnik: There is also important Javadoc attached to each setter, and this way the available options are easily discoverable.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/ServiceFactory.java
mtopolnik: Fixed by nulling out in `attachedFiles` call.

hazelcast-jet-python/src/main/java/com/hazelcast/jet/python/PythonServiceConfig.java
mtopolnik: You mean a field? Methods should stay as they have important Javadoc on them and make all the options easily discoverable.

hazelcast-jet-python/src/main/java/com/hazelcast/jet/python/PythonService.java
```diff
@@ -0,0 +1,189 @@
...
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * The service object used by the "map using Python" pipeline stage. As a
+ * user you don't have to deal with this class directly. It is used when
+ * you write {@link #mapUsingPython
+ * stage.apply(PythonService.mapUsingPython(pyConfig))}
+ */
+class PythonService {
```
cangencer: I mistakenly made this package-private, reverted now, also Nonnull / Nullable is missing for methods as well as the since tags

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/ServiceFactory.java
```diff
@@ -315,10 +310,65 @@ private ServiceFactory(
...
+     * ProcessorSupplier.Context#attachedFile
+     * procSupplierContext.attachedFile(file.toString())}.
+     *
+     * @return a copy of this factory with the file attached
+     *
+     * @since 4.0
+     */
+    @Nonnull
+    public ServiceFactory<C, S> withAttachedFile(@Nonnull File file) {
+        if (!file.isFile()) {
```
cangencer: we don't support ID here which is contradictory with how it works when doing it through JobConfig. Wouldn't it make sense to always require ID here, and then pythonService can add its own special id for these files?

hazelcast-jet-python/src/main/java/com/hazelcast/jet/python/PythonService.java
```diff
@@ -0,0 +1,199 @@
...
+                .mapUsingServiceAsyncBatched(factory(cfg), Integer.MAX_VALUE, PythonService::sendRequest)
+                .setName("mapUsingPython");
+    }
+
+    /**
+     * A stage-transforming method that adds a "map using Python" pipeline stage.
+     * Use it with {@code stage.apply(PythonService.mapUsingPythonBatch(pyConfig))}.
+     */
+    @Nonnull
+    public static FunctionEx<BatchStage<String>, BatchStage<String>> mapUsingPythonBatch(
```
eminn: Isn't this method implementation is 1:1 copy of the one above ?

cangencer: It's not possible to make them have same name, as the type params are different

eminn: The batch in the method name confused me, I've missed the signature

cangencer:
verify

eminn:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
```diff
@@ -211,11 +212,12 @@ long uploadJobResources(JobConfig jobConfig) {
...
-                        try (
-                                InputStream in = rc.getUrl().openStream();
-                                IMapOutputStream os = new IMapOutputStream(jobFileStorage.get(), fileKeyName(rc.getId()))
+                        Path fnamePath = Paths.get(Objects.requireNonNull(rc.getUrl().getPath())).getFileName();
+                        assert fnamePath != null; // needed to silence SpotBugs
+                        try (InputStream in = rc.getUrl().openStream();
+                             IMapOutputStream os = new IMapOutputStream(jobFileStorage.get(), fileKeyName(rc.getId()))
                         ) {
-                            packStreamIntoZip(in, os, rc.getId());
+                            packStreamIntoZip(in, os, fnamePath.toString());
```
viliam-durina: This is rather confusing. If the user doesn't use the `attachFile` variant with `id`, then the file name and ID are always equal. If he does, then he does so for some reason. We should use the ID he asked for and not the one from the URL. The URL might even not contain a file name, but we require it here.

If we keep it like this, we should document at `attachFile(URL, id)` that we require the URL to contain a file name and that that file name will be the name of the file, the ID is just used to look up that file.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/Contexts.java
```diff
@@ -185,13 +188,17 @@ private File extractFileToDisk(String id) {
...
+                Path directory = Files.createTempDirectory(sanitizeFilename(prefix));
                 unzip(inputStream, directory);
                 return directory.toFile();
             } catch (IOException e) {
                 throw ExceptionUtil.rethrow(e);
             }
         }
+
+        private static String sanitizeFilename(String proposed) {
+            return proposed.replace('/', '_');
```
viliam-durina: This is a good idea. The `id` is user-supplied, it can be whatever. We should whitelist allowed characters and replace all others. Also maybe trim it to up to 20 chars or so.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/ServiceFactory.java
```diff
@@ -390,4 +439,31 @@ public int maxPendingCallsPerProcessor() {
...
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    protected ServiceFactory<C, S> clone() {
+        try {
+            ServiceFactory<C, S> copy = (ServiceFactory<C, S>) super.clone();
+            copy.attachedFiles = new HashMap<>(attachedFiles);
+            return copy;
+        } catch (CloneNotSupportedException e) {
```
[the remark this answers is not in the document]
mtopolnik: The point is to not have to deal with this exception when using the specific type. The exception will never occur.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
mtopolnik: ID must be unique, and if we use it as the filename then no two resources can have the same filename. On the other hand, a given filename may be required and there is no uniqueness requirement.

viliam-durina: Then we should document that the URL must contain a file name and that the file will actually have that name and ID is used to query that file. We can also mention that multiple files can have the same name and different ID, it will make that option obvious.

mtopolnik: The URL has in fact always had an even stronger requirement: to be resolvable into a local file. We should document that explicitly.

hazelcast-jet-python/src/main/java/com/hazelcast/jet/python/JetToPythonServer.java
```diff
@@ -0,0 +1,99 @@
...
+            logger.info("Started Python process: " + pythonProcessPid);
+            serverSocket.setSoTimeout((int) SECONDS.toMillis(1));
+            while (true) {
+                try (Socket clientSocket = serverSocket.accept()) {
+                    int serverPort = Integer.parseInt(new BufferedReader(
+                            new InputStreamReader(clientSocket.getInputStream(), UTF_8)).readLine());
+                    logger.info("Python process " + pythonProcessPid + " listening on port " + serverPort);
+                    return serverPort;
+                } catch (SocketTimeoutException e) {
+                    if (pythonProcess.waitFor(1, SECONDS)) {
```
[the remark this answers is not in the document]
mtopolnik: We make sure there are no process leaks. If a job gets stuck, that's visible and actionable more than orphan Python processes piling up.

The loop is there to keep resending the termination signal.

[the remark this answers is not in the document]
mtopolnik: We configure the socket timeout to 1 second, and if it elapses, we wait for one second before going to another `accept()` call, to see if the Python process will die. We could also just check `process.isAlive()` and go to another loop iteration. It's pretty arbitrary.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
viliam-durina: Maybe in the past, now we only use `url.openStream()`, so it has to be any URL that can be read.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/JetInstance.java
```diff
@@ -138,6 +141,17 @@ default Job newJob(@Nonnull Pipeline pipeline) {
      */
     @Nonnull
     default Job newJob(@Nonnull Pipeline pipeline, @Nonnull JobConfig config) {
+        PipelineImpl impl = (PipelineImpl) pipeline;
+        for (Entry<String, File> e : impl.attachedFiles().entrySet()) {
+            File file = e.getValue();
+            if (file.isDirectory()) {
+                config.attachDirectory(file, e.getKey());
+            } else if (file.isFile()) {
```
viliam-durina: `isFile` returns false if the file is in a directory which lacks the `x` attribute, see [here](https://stackoverflow.com/questions/25039156/java-what-constitutes-a-normal-file-in-the-context-of-file-isfile). It's better to use `canRead` or to not check at all and rely when we try to read it.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
viliam-durina: Maybe in the past, but now we only use `url.openStream()`

hazelcast-jet-python/src/main/java/com/hazelcast/jet/python/JetToPythonServer.java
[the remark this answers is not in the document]
viliam-durina: that was my point that it's pretty arbitrary. Why not going to accept again immediately?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
mtopolnik: The complete story is in `JobConfig`, where we convert URLs to files. I'm adding more explicit checking there.

viliam-durina: We convert the URLs to file names, not to files. A http URL works just fine, if it contains a file name.

Nevertheless, I was thinking about it more. Maybe we can leave it as it is. It's very unlikely to encounter a URL like `http://hazelcast.com/` that has no file name. The user will receive a `NPE` or `AssertionError` in this case, which is very easy to fix by looking at the code. I'm marking this as resolved.

cangencer:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobRepository.java
mtopolnik: In the meantime i cleaned up `JobConfig` to make the checks where needed. `attachFile/Directory` must get a `file` URL and checks for the existence of the file. The `add` variants can get an opaque URL that they don't check, but they still check it has at least one path segment to be used as the ID.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/JetInstance.java
mtopolnik: `isFile` is used ubiquitously to check that a file exists as a regular file. I can only add one more check `file.canRead`

mtopolnik: You can't open a file in a directory without the `x` permission:

```
$ cat test/test.txt 
test
$ chmod -x test
$ cat test/test.txt 
cat: test/test.txt: Permission denied
```

PR #1872, opened 2020-01-22
Remove noisy logging from JobCoordinationService

cangencer:
It continuously logs that partitions are not initialized when this is just business as usual


cangencer:
Closing this as the log message is still useful sometimes for debugging, and it's already only at DEBUG level.

PR #1874, opened 2020-01-23
Update README.md

pivovarit:
Corrected some typos as I was going through the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,13 +14,13 @@ Jet is simple to set up. The nodes you start discover each other and
 form a cluster automatically. You can do the same locally, even on the
 same machine (your laptop, for example). This is great for quick testing.
 
-With Jet it's easy to build fault-tolerant and elastic data processing
+With Jet, it's easy to build fault-tolerant and elastic data processing
 pipelines. Jet keeps processing data without loss even if a node fails,
 and you can add more nodes that immediately start sharing the
 computation load.
 
-You can embed Jet as a part of your application, it's just a single JAR
-without dependencies. You can also deploy it standalone, as a
+You can embed Jet as a part of your application. It's just a single JAR
+without dependencies. You can also deploy it standalone as a
 stream-processing cluster.
 
 Jet also provides a highly available, distributed in-memory data store.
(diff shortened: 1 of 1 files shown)


cangencer:
verify

README.md
```diff
@@ -14,13 +14,13 @@ Jet is simple to set up. The nodes you start discover each other and
 form a cluster automatically. You can do the same locally, even on the
 same machine (your laptop, for example). This is great for quick testing.
 
-With Jet it's easy to build fault-tolerant and elastic data processing
+With Jet, it's easy to build fault-tolerant and elastic data processing
```
mtopolnik: This comma is not needed.

README.md
```diff
@@ -70,16 +70,16 @@ jet.newJob(p).join();
 
 When you run this, point the `path` variable to some directory with text
 files in it. Jet will analyze all the files and give you the word
-frequency distribution in the log output (for each word it will say how
+frequency distribution in the log output (for each word, it will say how
```
mtopolnik: This comma is not needed.

README.md
```diff
@@ -70,16 +70,16 @@ jet.newJob(p).join();
...
 
 The above was an example of processing data at rest (i.e., _batch
-processing_). It's conceptually simpler than stream processing so we
+processing_). It's conceptually simpler than stream processing, so we
 used it as our first example.
 
 ### Stream Processing with Jet
 
-For stream processing you need a streaming data source. A simple example
+For stream processing, you need a streaming data source. A simple example
```
mtopolnik: This comma is not needed.

README.md
```diff
@@ -103,8 +103,8 @@ p.readFrom(Sources.fileWatcher(path))
 jet.newJob(p).join();
 ```
 
-Before running this make an empty directory and point the `path`
-variable to it. While the job is running copy some text files into it
+Before running this, make an empty directory and point the `path`
```
mtopolnik: This comma is not needed.

pivovarit: Did you verify this with a native?
I'm quite convinced that if you reverse the order of `Make an empty directory and point the `path` variable to it before running this`, it deserves a comma (similar with above examples)

mtopolnik: The difference is whether it's a noun phrase or a subordinate clause. `Before running this` is not a subordinate clause, but `before you run this` would be.

PR #1876, opened 2020-01-23
Add missing metrics to RefMan

jbartok:
Add description here

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/log-debug.adoc
+++ b/hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/log-debug.adoc
@@ -182,12 +182,12 @@ Additionally, if the vertex sourcing them is a data source or data sink, then th
 **executionCompletionTime**: Start time of the current execution of the job. +
 
 Both contain epoch time in milliseconds.
-|** * only ones without `vertex` tag** +
+|** no extra tags, they are even missing `vertex`** +
 There will be a single instance of these metrics for each job execution.
 
 |**snapshotBytes**: Total number of bytes written out in the last snapshot. +
 **snapshotKeys**: Total number of keys written out in the last snapshot.
-|**none** +
+|**no extra tags** +
 There will be a single instance of these metrics for each vertex.
 
 |**distributedBytesIn**: Total number of bytes received from remote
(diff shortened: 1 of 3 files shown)


hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/log-debug.adoc
```diff
@@ -178,6 +178,13 @@ Additionally, if the vertex sourcing them is a data source or data sink, then th
 |===
 |Names|Extra tags
 
+|**executionStartTime**: Start time of the current execution of a job. +
+**executionCompletionTime**: Start time of the current execution of a job. +
+
+Both contain epoch time in milliseconds.
+|** * only ones without `vertex` tag** +
```
cangencer: what does this mean?

jbartok: See the first paragraph of the "Jet Job metrics" section, stating that almost all job metrics have the 'vertex' tag set (ie. are per vertex). So I wanted to point out that these two metrics are the exceptions.

jbartok: Pushed a more straightforward version.

hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/log-debug.adoc
```diff
@@ -178,9 +178,16 @@ Additionally, if the vertex sourcing them is a data source or data sink, then th
 |===
 |Names|Extra tags
 
+|**executionStartTime**: Start time of the current execution of the job. +
+**executionCompletionTime**: Start time of the current execution of the job. +
```
viliam-durina: copy-paste error

hazelcast-jet-reference-manual/src/main/asciidoc/work-with-jet/log-debug.adoc
```diff
@@ -178,9 +178,16 @@ Additionally, if the vertex sourcing them is a data source or data sink, then th
...
+
+Both contain epoch time in milliseconds.
+|** no extra tags, they are even missing `vertex`** +
```
viliam-durina: `they are even missing 'vertex'` - no surprise here. I personally would just skip this part, the next sentence is enough,

jbartok:
> I wonder what the value is when the job is running.

There is no completion time metric value when the job is running (see `ExecutionContext.provideDynamicMetrics`, the metric gets set by dynamic collection only if there is a non-zero value).

Actually the only way to get/see a value for this execution completion metric is to set `JobConfig.storeMetricsAfterJobCompletion` and then use the Job Metrics API (after completion).

PR #1879, opened 2020-01-23
Fix emission of older watermark

viliam-durina:
This could happen:
- in `currentTraverser` there was a WM pending, but there was no
unfinished item
- the initial check in `tryProcessWatermark` didn't cause to return
immediately nor attempt finishing of the current traverser
- the `watermarkCounts` collection was empty, so we added the new WM
directly to outbox
- later we emitted the `currentTraverser` and an older WM was added to
the outbox


viliam-durina:
verify

PR #1880, opened 2020-01-23
Fix ThrottleWrappedP

viliam-durina:
Didn't cause any actual test failure, but it could :)

- didn't override all methods
- didn't throttle when offer(int[], Object) was used


viliam-durina:
verify

PR #1884, opened 2020-01-24
Use Util.mappedList in obvious places

viliam-durina:


viliam-durina:
@cangencer Just revert those occurrences you don't like. If you don't like all of them, close this PR. I won't dispute the style.

I personally find it useful, there are surprisingly many places where we map one collection into another. This removes the `stream().map` and `collect(Collectors.toList())` parts. But I agree some more convoluted instances might be more readable without it.

PR #1886, opened 2020-01-27
Add RabbitMQ tests and fix issues with it

viliam-durina:
Add tests for one more JMS broker. RabbitMQ throws very misleading error when `cf.createConnection(null, null)` is used, ActiveMQ is fine with that...

RabbitMQ tests are run nightly.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/JmsSourceBuilder.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/JmsSourceBuilder.java
@@ -86,8 +86,8 @@ public final class JmsSourceBuilder {
      * If not provided, this function is used:
      * <pre>
      *     connectionFn = factory -> username != null || password != null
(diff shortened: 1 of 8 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/JmsSourceBuilder.java
```diff
@@ -85,11 +85,10 @@ public JmsSourceBuilder connectionParams(@Nullable String username, @Nullable St
      * <p>
      * If not provided, this function is used:
      * <pre>
-     *     connectionFn = factory -> factory instanceof XAConnectionFactory
-     *            ? ((XAConnectionFactory) factory).createXAConnection(username, password)
-     *            : factory.createConnection(username, password);
+     *     connectionFn = factory -> username != null || password != null
+     *                     ? factory.createConnection(usernameLocal, passwordLocal)
```
olukas: Is this alignment intended?

PR #1916, opened 2020-01-31
Update Kubernetes plugin version

leszko:
Update Kubernetes plugin version to 2.0.1


pom.xml
```diff
@@ -74,7 +74,7 @@
         <!-- Dependencies on IMDG modules: must be the same versions IMDG depends on! -->
         <hazelcast.protocol.version>1.8.0-5</hazelcast.protocol.version>
         <hazelcast.aws.version>3.0.1-SNAPSHOT</hazelcast.aws.version>
```
cangencer: could you update this as well?

leszko: Updated

PR #1917, opened 2020-01-31
Python Code Sample and Reference Manual, Plus Some Improvements

mtopolnik:
Fixes https://github.com/hazelcast/hazelcast-jet/issues/1905

Checklist
- [x] Tags Set
- [x] Milestone Set


examples/python/src/main/java/com/hazelcast/jet/examples/python/Python.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Objects;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.stream.Stream;
+
+import static com.hazelcast.jet.impl.util.Util.uncheckRun;
+import static com.hazelcast.jet.python.PythonTransforms.mapUsingPython;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * This example shows you how to invoke a Python function to process the
```
cangencer: I think we should say what the pipeline does, rather than just "process data". It's better to lead by describing the actual processing that is done

examples/python/src/main/java/com/hazelcast/jet/examples/python/Python.java
```diff
@@ -0,0 +1,109 @@
...
+            observable.addObserver(Observer.of(System.out::println));
+            JobConfig config = new JobConfig().setName("python-mapping");
+            jet.newJobIfAbsent(p, config).join();
+        } finally {
+            IOUtil.delete(baseDir);
+            Jet.shutdownAll();
+        }
+    }
+
+    private static void copyClasspathResourcesToDirectory(String sourcePath, Path destPath) throws IOException {
```
cangencer: This seem rather like boilerplate and a lot of code.. could we move it somewhere else perhaps? as a util method on Python?

examples/python/src/main/resources/python/take_sqrt.py
```diff
@@ -0,0 +1,7 @@
+import numpy as np
+
+
+def transform_list(input_list):
```
cangencer: I would add some python style docs here describing the input parameters etc as standard in python

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -688,6 +689,32 @@ public JobConfig attachDirectory(@Nonnull File file, @Nonnull String id) {
...
+    /**
+     * Attaches all the files/directories in the supplied map, as if by calling
+     * {@link #attachDirectory(File, String) attachDirectory(dir, id)} for every
+     * entry that resolves to a directory and {@link #attachFile(File, String)
+     * attachFile(file, id)} for every entry that resolves to a regular file.
+     *
+     * @return {@code this} instance for fluent API
+     */
+    @Nonnull
+    public JobConfig attachAll(@Nonnull Map<String, File> idToFile) {
```
cangencer: this one is missing a test

PR #1919, opened 2020-01-31
Service async cleanup

### LINKED ISSUE #1909: Cleanup of *usingService(Async) methods

jbartok:
Add description here

Checklist
- [x] Tags Set
- [x] Milestone Set

Fixes mandatory parts of #1909 (optional parts will be in separate PR).

List of breaking changes:
* `filterUsingServiceAsync` operation removed from all stages and processors
* `flatMapUsingServiceAsync` operation removed from all stages and processors
* asynchronous processing specific parameters moved from `ServiceFactory` to `mapUsingServiceAsync` operation of stages & processors

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/Processors.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/Processors.java
@@ -936,37 +936,6 @@ public final class Processors {
     }
 
     /**
-     * Asynchronous version of {@link #mapUsingServiceP}: the {@code
-     * filterAsyncFn} returns a {@code CompletableFuture<Boolean>} instead of
-     * just a {@code boolean}.
-     * <p>
-     * The function can return a null future, but the future must not return
-     * null {@code Boolean}.
-     * <p>
-     * The {@code extractKeyFn} is used to extract keys under which to save
-     * in-flight items to the snapshot. If the input to this processor is over
-     * a partitioned edge, you should use the same key. If it's a round-robin
-     * edge, you can use any key, for example {@code Object::hashCode}.
-     *
-     * @param serviceFactory the service factory
-     * @param extractKeyFn a function to extract snapshot keys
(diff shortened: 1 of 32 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/Processors.java
```diff
@@ -901,11 +902,16 @@ private Processors() {
     @Nonnull
     public static <C, S, T, K, R> ProcessorSupplier mapUsingServiceAsyncP(
             @Nonnull ServiceFactory<C, S> serviceFactory,
+            int maxAsyncOps,
+            boolean orderedAsyncResponses,
```
cangencer: how about `preserveOrder` ?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/Processors.java
```diff
@@ -901,11 +902,16 @@ private Processors() {
     @Nonnull
     public static <C, S, T, K, R> ProcessorSupplier mapUsingServiceAsyncP(
             @Nonnull ServiceFactory<C, S> serviceFactory,
+            int maxAsyncOps,
```
cangencer: `maxConcurrentOps` perhaps?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/ComputeStageImplBase.java
```diff
@@ -298,7 +302,7 @@
 
         return (RET) attach(
                 flatMapUsingServiceAsyncBatchedTransform(
-                        transform, operationName, serviceFactory, maxBatchSize, flattenedFn),
+                        transform, operationName, serviceFactory, 2, maxBatchSize, flattenedFn),
```
mtopolnik: This `2` is burried to deeply, should extract a descriptive constant

PR #1925, opened 2020-02-04
Fix refman formatting & grammar

gierlachg:
Fixed reference manual formatting in couple of places.

### DIFF
--- a/reference-manual/src/main/asciidoc/expert-zone/dag.adoc
+++ b/reference-manual/src/main/asciidoc/expert-zone/dag.adoc
@@ -263,7 +263,7 @@ is often significantly faster and it's safe to use on a local edge.
 On a distributed edge it is not a safe strategy in general because
 `hashCode()` 's contract does not require repeatable results across
 JVMs or even different instances of the same JVM version. If a given
-class' Javadoc explicitly specifies the hashing function it uses, then
+class's Javadoc explicitly specifies the hashing function it uses, then
 its instances are safe to partition with `hashCode()`.
 
 You can provide your own implementation of `Partitioner` to gain full
--- a/reference-manual/src/main/asciidoc/expert-zone/dag.adoc
+++ b/reference-manual/src/main/asciidoc/expert-zone/dag.adoc
@@ -215,7 +215,7 @@ the redundancy it creates, this kind of edge isn't appropriate for the
 main data stream. Its purpose is dispatching a small amount of data to
 all the processors of a vertex, typically as a part of setting up before
 processing the data stream. For this reason a broadcast edge is often
-high-priority as well. For example, in a hash-join one vertex creates
+high-priority as well. For example, in a hash join one vertex creates
 the lookup hashtable and sends the same instance to all the processors
 of the next vertex, the one that processes the main data stream.
 
@@ -321,7 +321,7 @@ include::{javasource}/expertzone/Dag.java[tag=s9]
 By default the processor receives items from all inbound edges as they
 arrive. However, there are important cases where an edge must be
 consumed in full to make the processor ready to accept data from other
-edges. A major example is a `hash-join` which enriches the data stream
+edges. A major example is a hash join which enriches the data stream
 with data from a lookup table. This can be modeled as a join of two data
 streams where the _enriching_ stream contains the data for the lookup
 table and must be consumed in full before consuming the stream to be
--- a/reference-manual/src/main/asciidoc/expert-zone/dag.adoc
+++ b/reference-manual/src/main/asciidoc/expert-zone/dag.adoc
@@ -263,7 +263,7 @@ is often significantly faster and it's safe to use on a local edge.
 On a distributed edge it is not a safe strategy in general because
 `hashCode()` 's contract does not require repeatable results across
 JVMs or even different instances of the same JVM version. If a given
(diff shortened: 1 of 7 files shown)


reference-manual/src/main/asciidoc/expert-zone/dag.adoc
```diff
@@ -321,7 +321,7 @@ include::{javasource}/expertzone/Dag.java[tag=s9]
 By default the processor receives items from all inbound edges as they
 arrive. However, there are important cases where an edge must be
 consumed in full to make the processor ready to accept data from other
-edges. A major example is a "`hash join`" which enriches the data stream
+edges. A major example is a `hash-join` which enriches the data stream
```
viliam-durina: "hash join" isn't a piece of code, it shouldn't use back-ticks. Also shouldn't have the dash: a dash is used when you make one word form multiple, but in this case it's still two words, try to google for hash-join with the dash, i found no example. It could maybe be in italics.

reference-manual/src/main/asciidoc/jet-concepts.adoc
```diff
@@ -576,7 +576,7 @@ maintain over a long term.
 
 ==== Caveats When Changing the Grouping Key
 
-You can change the key-extracting function for a `groupingKey`'s, but
+You can change the key-extracting function for a `groupingKey` 's, but
```
viliam-durina: I guess this doesn't look great rendered. Also, plural shouldn't use apostrophe. And plural shouldn't use "a". I think we can go with a singular and also add `()` to indicate it's a method:
```
for a `groupingKey()`.
```

reference-manual/src/main/asciidoc/pipeline-api/pipeline-api.adoc
```diff
@@ -278,7 +278,7 @@ time setting a different product on it. This will create a data race
 with the downstream stage, which may observe the same product several
 times while missing the other ones.
 
-The safe way to perform a many-to-many hash join is to either package
+The safe way to perform a many-to-many `hash-join` is to either package
```
viliam-durina: again, no dash, no code.

reference-manual/src/main/asciidoc/expert-zone/dag.adoc
mtopolnik: I just ran into this while re-reading the refman. As a verb, it should be "hash-join" but as a noun, it should be "hash join".

reference-manual/src/main/asciidoc/expert-zone/dag.adoc
```diff
@@ -263,7 +263,7 @@ is often significantly faster and it's safe to use on a local edge.
 On a distributed edge it is not a safe strategy in general because
 `hashCode()` 's contract does not require repeatable results across
 JVMs or even different instances of the same JVM version. If a given
-class's Javadoc explicitly specifies the hashing function it uses, then
+class' Javadoc explicitly specifies the hashing function it uses, then
```
mtopolnik: It should actually stay as "class's". Technically both are acceptable, no point in changing from one to the other and frankly `class'` doesn't looks so familiar to me.

reference-manual/src/main/asciidoc/expert-zone/dag.adoc
gierlachg: :thinking: there are 22 occurrences of `hash-join`s in the reference manual while only 3 `hash join`s... should I edit all of them?

reference-manual/src/main/asciidoc/expert-zone/dag.adoc
gierlachg: Actually I wanted it to be consistent with other spelling `members'`, `items'`...

reference-manual/src/main/asciidoc/expert-zone/dag.adoc
gierlachg: I just saw your comment @mtopolnik, let me go through the usages then

gierlachg: Went through all of them, does it make more sense now?

reference-manual/src/main/asciidoc/pipeline-api/pipeline-api.adoc
```diff
@@ -196,7 +196,7 @@ enrichment_.
 
 Jet offers two basic techniques to enrich your data stream:
 
-. <<hash-join, Hash-join>> the stream with one or more datasets that you
+. <<hash join, Hash-join>> the stream with one or more datasets that you
```
viliam-durina: This should be the other way: 
```
<<hash-join, Hash join>>
```

reference-manual/src/main/asciidoc/expert-zone/dag.adoc
viliam-durina: "members" is plural, "class" is singular. See https://english.stackexchange.com/q/1073/232121

reference-manual/src/main/asciidoc/pipeline-api/pipeline-api.adoc
gierlachg: Yes, I would broke the link... but i have left `Hash-join` as a verb.

PR #1928, opened 2020-02-05
Freeze gRPC versions installed via pip

mtopolnik:
Add description here

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/extensions/python/src/main/resources/jet_to_python_init.sh
+++ b/extensions/python/src/main/resources/jet_to_python_init.sh
@@ -5,7 +5,7 @@ set -e -x
 
 python3 -m venv jet_to_python_env --system-site-packages
 source jet_to_python_env/bin/activate
-python3 -m pip install protobuf grpcio
+python3 -m pip install protobuf==3.11.3 grpcio==1.26.0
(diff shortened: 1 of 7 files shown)


extensions/python/src/main/resources/jet_to_python_init.sh
```diff
@@ -5,7 +5,7 @@ set -e -x
 
 python3 -m venv jet_to_python_env --system-site-packages
 source jet_to_python_env/bin/activate
-python3 -m pip install protobuf grpcio
+python3 -m pip install protobuf==3.11.3 grpcio==1.26.0
```
cangencer: I think it should be possible to use maven properties here and add a resource transformation step

cangencer: see for example `jet-runtime.properties`

cangencer: I see for example in python module we use proto 3.9 and 3.11 here and also different grpc versions

mtopolnik: That's updated as well.

PR #1929, opened 2020-02-05
Add keyed version of `mapUsingServiceAsyncBatched`

### LINKED ISSUE #1909: Cleanup of *usingService(Async) methods

jbartok:
Add description here

Checklist
- [x] Tags Set
- [x] Milestone Set

Fixes #1909


jbartok:
I'm missing the stream part (that's why it's a draft PR for now), but the batch part should be ok, pls. take a look.

mtopolnik:
verify

Issue #1689, opened 2019-10-12
Improve the way classes are added to a Jet job


mtopolnik:
There are two issues with packaging dependency classes with a Jet job when submitting from a Jet client:

1. There is no way to add anonymous classes. In particular, Kotlin compiles lambdas to anonymous classes that can't be packaged into a Jet job.
2. You can only add class by class, making the process cumbersome. Adding a whole package would be much simpler for the user.

Suggestions:

1. `JobConfig.addClass()` should add the class's nested classes as well.
2. introduce a new method `JobConfig.addPackage()`.

PR #1932, opened 2020-02-06
Add nested classes & packages to JobConfig

gierlachg:
`JobConfig.addClass()` adds nested classes to resources as well.

Checklist
- [x] Tags Set
- [x] Milestone Set

[#1689](https://github.com/hazelcast/hazelcast-jet/issues/1689)


hazelcast-jet-core/src/test/java/com/hazelcast/jet/config/ResourceConfigTest.java
```diff
@@ -1024,4 +1033,9 @@ private File createDirectory(String path) {
         assertTrue("Failed to create directory " + dirFile, dirFile.mkdirs());
         return dirFile;
     }
+
+    private static class BaseClass {
+        private static class NestedClass {
```
gurbuzali: can we add a test with anonymous class too

gierlachg: Actually, anonymous class case is not covered yet, I'm working on it right now - will do.

cangencer: Could you also test with Kotlin - since Kotlin lambdas are always created as anonymous classes.

gierlachg: Do you mean actually adding kotlin as test dependency or just make sure it works with it? Adding kotlin tests might further slow down the build?

cangencer: Maybe just test it for now - I don't know how complicated it would be to add it as a full dependency and some tests. It's a nice to have.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -232,10 +241,26 @@ public JobConfig setSnapshotIntervalMillis(long snapshotInterval) {
...
             ResourceConfig cfg = new ResourceConfig(clazz);
             resourceConfigs.put(cfg.getId(), cfg);
+
+            ClassLoader classLoader = clazz.getClassLoader();
+            String packagePrefixPath = Optional.ofNullable(clazz.getPackage())
+                                               .map(pakage -> pakage.getName().replace('.', '/') + "/")
+                                               .orElse("");
+            Pattern nestedPattern = Pattern.compile(clazz.getName()
+                                                         .replace(".", "\\/")
+                                                         .replace("$", "\\$") + "\\$.+" + "\\.class");
```
mtopolnik: You can simplify this by first building a literal class string (no escaping) and then calling `Pattern.quote()` on it.

hazelcast-jet-core/src/test/kotlin/com/hazelcast/jet/config/KotlinResourceTest.kt
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Test
+    fun shouldAddNestedAndAnonymousClasses() {
+        // When
+        config!!.addClass(OuterClass::class.java)
+
+        // Then
+        val resourceConfigs: Collection<ResourceConfig> = config!!.resourceConfigs.values
+        Assert.assertThat(resourceConfigs, Matchers.hasSize(4))
```
mtopolnik: Would you mind using static imports here? It would remove quite a lot of noise.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -219,8 +228,8 @@ public JobConfig setSnapshotIntervalMillis(long snapshotInterval) {
     }
 
     /**
-     * Adds the given classes to the Jet job's classpath. They will be
-     * accessible to all the code attached to the underlying pipeline or DAG,
+     * Adds the given classes and recursively all their nested classes to the Jet job's classpath.
```
cangencer: minor, but we wrap Javadoc at 72 chars (net)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -232,10 +241,26 @@ public JobConfig setSnapshotIntervalMillis(long snapshotInterval) {
...
             resourceConfigs.put(cfg.getId(), cfg);
+
+            ClassLoader classLoader = clazz.getClassLoader();
+            String packagePrefixPath = Optional.ofNullable(clazz.getPackage())
+                                               .map(pakage -> pakage.getName().replace('.', '/') + "/")
+                                               .orElse("");
+            Pattern nestedPattern = Pattern.compile(
+                    Pattern.quote(clazz.getName().replace('.', '/')) + "\\$.+" + "\\.class"
+            );
+            resources(classLoader, packagePrefixPath)
```
gurbuzali: this does not work if the class is coming from a dependency. I've created a jar which contains the `OuterClass` and added that jar as a dependency to my project. When I called `addClass`, the returned resource list is empty.

eminn:
We'll need to include a notice in the `NOTICE` file similar to [this](https://github.com/hazelcast/hazelcast/blob/master/NOTICE#L24) for shaded jars. I also noticed that it is missing for the `picocli`. Could you please fix it for both of them ?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -238,11 +240,40 @@ public JobConfig setSnapshotIntervalMillis(long snapshotInterval) {
     @Nonnull
     @SuppressWarnings("rawtypes")
     public JobConfig addClass(@Nonnull Class... classes) {
-        checkNotNull(classes, "Classes can not be null");
-        for (Class<?> clazz : classes) {
+        checkNotNull(classes, "Classes cannot be null");
```
mtopolnik: Do we really have to double-check a `@Nonnull` parameter?

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/util/ReflectionUtilsTest.java
```diff
@@ -50,9 +63,49 @@ public void readStaticFieldOrNull_readFromPublicField() {
         assertEquals("staticPublicFieldContent", field);
     }
 
+    @Test
+    public void shouldDiscoverAllNestedClasses() throws ClassNotFoundException {
```
mtopolnik: The `when_<PRECONDITION>_then_<POSTCONDITION>` name pattern is more useful. Here you could add `when_nestedClassesOf_then_`.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/util/ReflectionUtilsTest.java
```diff
@@ -50,9 +63,49 @@ public void readStaticFieldOrNull_readFromPublicField() {
...
+        assertThat(classes, hasSize(3));
+        assertThat(classes, containsInAnyOrder(
+                OuterClass.class,
+                OuterClass.NestedClass.class,
+                Class.forName("com.hazelcast.jet.impl.util.ReflectionUtilsTest$OuterClass$1")
+        ));
+    }
+
+    @Test
+    public void shouldDiscoverAllClassesAndResourcesInAPackage() {
```
mtopolnik: This should assert that the nested class is also there.

gierlachg: Then we would actually test whether JVM creates nested classes in the same package as the root class which is not really the goal here.

mtopolnik: It definitely is in the same package, and if it's not included, then `addPackages` is broken, going back to the place we were before this PR.

gierlachg: What I mean is - there are many more classes in this package, the nested class is just one of them, in the context of this test nested class is no different from the _regular_ class, I guess to properly test it I should create a separate package with known classes in it and ensure all of them are listed, I was thinking about it but then didn't like the idea of creating yet another package... WDYT?

mtopolnik: My thinking was explicitly look for the nested classes that you already look for in `addClass`.

PR #1934, opened 2020-02-06
Remove offset committing from Kafka source

### LINKED ISSUE #1924: Fix working with offsets in Kafka Sink
When not snapshotting, we call `commitSync` after each batch of records. This is:
- slow: `commitAsync` would give exactly same guarantee, but not block
- forces the user to specify `group.id`. In samples we use random UUID as a group id, which is nonsense, committed offsets only take time to write them and space to store them.

We could remove this logic altogether. If the user runs the job without snapshots, it's questionable whether he needs to have some approximation of exactly-once for the source while the rest of the job loses any state. And if he really does, he can use `enable.auto.commit`.

viliam-durina:
Fixes #1924


viliam-durina:
verify

extensions/kafka/src/main/java/com/hazelcast/jet/kafka/KafkaSources.java
```diff
@@ -91,12 +90,13 @@ private KafkaSources() {
...
-     *     offsets to Kafka using {@link KafkaConsumer#commitSync()}. But the
-     *     offsets are committed before or after the event is fully processed.
-     *     Therefore some events can be processed twice or not at all. You can
-     *     configure {@code group.id} in this case. If not configured a random
-     *     UUID will be set.
+     *     <li>if processing guarantee is disabled, the source will start
+     *     reading from default offsets (based on the {@code auto.offset.reset}
+     *     property). You can enable offset committing by assigning a {@code
+     *     group.id}, enabling auto offset committing using {@code
+     *     enable.auto.commit} and configuring {@code auto.commit.interval.ms}
```
cangencer: this is normally on by default, do we explicitly disable it? if so this should be mentioned.

viliam-durina: We don't disable it, now we do nothing to the properties. It's actually good for us: just by assigning the `group.id` we rely on kafka committing the offsets.

extensions/kafka/src/main/java/com/hazelcast/jet/kafka/KafkaSources.java
```diff
@@ -110,7 +110,7 @@ private KafkaSources() {
      * {@code poll(timeout)} if the cluster is down. If {@link
      * JobConfig#setSnapshotIntervalMillis(long) snapshotting is enabled},
      * entire job might be blocked. This is a known issue of Kafka
-     * (KAFKA-1894). Refer to Kafka documentation for details.
+     * (KAFKA-1894, now fixed). Refer to Kafka documentation for details.
```
cangencer: should we remove this? since it's been fixed?

PR #1937, opened 2020-02-06
Fix File sink with lp>1 or shared dir

### LINKED ISSUE #1935: Files sink with exactly-once is not able to work with more nodes on one machine
When exactly-once is configured for job which runs pipeline with `files` sink and is running on 2 or more members on the same machine then following Excaption is thrown after first snapshot:
```
WARNING: [192.168.0.102]:5701 [jet] [4.0-SNAPSHOT] Exception in ProcessorTasklet{filesSink(/tmp/filesink)#0}
java.nio.file.NoSuchFileException: /tmp/filesink/0-0.tmp -> /tmp/filesink/0-0
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
	at sun.nio.fs.UnixCopyFile.move(UnixCopyFile.java:396)
	at sun.nio.fs.UnixFileSystemProvider.move(UnixFileSystemProvider.java:262)
	at java.nio.file.Files.move(Files.java:1395)
	at com.hazelcast.jet.impl.connector.WriteFileP$FileWithTransaction.commit(WriteFileP.java:365)
	at com.hazelcast.jet.impl.processor.TwoPhaseSnapshotCommitUtility$LoggingNonThrowingResource.commit(TwoPhaseSnapshotCommitUtility.java:406)
	at com.hazelcast.jet.impl.processor.UnboundedTransactionsProcessorUtility.commitPendingTransactions(UnboundedTransactionsProcessorUtility.java:173)
	at com.hazelcast.jet.impl.processor.UnboundedTransactionsProcessorUtility.snapshotCommitFinish(UnboundedTransactionsProcessorUtility.java:166)
	at com.hazelcast.jet.impl.connector.WriteFileP.snapshotCommitFinish(WriteFileP.java:182)
	at com.hazelcast.jet.impl.processor.ProcessorWrapper.snapshotCommitFinish(ProcessorWrapper.java:128)
	at com.hazelcast.jet.impl.execution.ProcessorTasklet.stateMachineStep(ProcessorTasklet.java:335)
	at com.hazelcast.jet.impl.execution.ProcessorTasklet.call(ProcessorTasklet.java:240)
	at com.hazelcast.jet.impl.execution.TaskletExecutionService$CooperativeWorker.runTasklet(TaskletExecutionService.java:366)
	at java.util.concurrent.CopyOnWriteArrayList.forEach(CopyOnWriteArrayList.java:891)
	at com.hazelcast.jet.impl.execution.TaskletExecutionService$CooperativeWorker.run(TaskletExecutionService.java:346)
```
Only file `1-0` is created in `/tmp/filesink` and this file includes items emitted only by one of sink processors. Also before snapshot should be created only file `1-0.tmp` occurs in `/tmp/filesink`

This does not happen in exactly once is disabled.

**How to reproduce:**
Run job with pipeline like this on cluster with 2 members on the same machine (e.g. locally):
```
StreamSource<Long> source = SourceBuilder
        .stream("srcForFileSink", procCtx -> new long[1])
        .<Long>fillBufferFn((ctx, buf) -> {
            buf.add(ctx[0]++);
            Thread.sleep(100);
        })
        .createSnapshotFn(ctx -> ctx[0])
        .restoreSnapshotFn((ctx, state) -> ctx[0] = state.get(0))
        .build();
Sink<Object> sink = Sinks.files("/tmp/filesink");
p.readFrom(source)
        .withoutTimestamps()
        .groupingKey(identity())
        .filterUsingService(sharedService(ctx -> null), (s, k, v) -> true)
        .writeTo(sink);
```

viliam-durina:
Fixes #1935

In the original state all processors tried to delete all .tmp files in the
directory. However when P1 got to do it, P2 might already started creating
files which caused that P1 deleted working files of P2. On Windows, the delete
failed, on Linux the processor failed later when it was trying to commit the
file.


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteFileP.java
```diff
@@ -236,6 +248,33 @@ private void abortUnfinishedTransactions() {
...
+                            context.logger().warning(
+                                    "file with unknown name structure found in the directory: " + file, e);
+                            return false;
+                        }
+                        int localIndexLow = context.memberIndex() * context.localParallelism();
+                        int localIndexHigh = localIndexLow + context.localParallelism();
+                        boolean isLocalIndex = index >= localIndexLow && index < localIndexHigh;
+                        if (sharedFileSystem || isLocalIndex) {
+                            // If the file index belongs to one of the local processors (when directory is not shred)
+                            // or when the directory is shared, we must leave that processor to do the cleanup
```
gurbuzali: why don't we always leave the cleanup to the processor, looks like this will simplify the logic ?

viliam-durina: Because there might be no local processor with that index. For example, if LP=1 and you have 2 members. Member 0 will fail and be restarted. After restart it will have index 1, no member on local machine will have index 0. Therefore the files with index 0 will not be handled.

I agree it's quite complex and I was thinking about just ignoring it: the issue would be that some `.tmp` files will be left behind, but that's it.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteFileP.java
```diff
@@ -236,6 +248,33 @@ private void abortUnfinishedTransactions() {
...
+                        } catch (NumberFormatException e) {
+                            context.logger().warning(
+                                    "file with unknown name structure found in the directory: " + file, e);
+                            return false;
+                        }
+                        int localIndexLow = context.memberIndex() * context.localParallelism();
+                        int localIndexHigh = localIndexLow + context.localParallelism();
+                        boolean isLocalIndex = index >= localIndexLow && index < localIndexHigh;
+                        if (sharedFileSystem || isLocalIndex) {
+                            // If the file index belongs to one of the local processors (when directory is not shred)
```
cangencer: typo, "shared"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/WriteFileP.java
```diff
@@ -236,6 +248,33 @@ private void abortUnfinishedTransactions() {
...
+                        }
+                        int index;
+                        try {
+                            index = Integer.parseInt(m.group(1));
+                        } catch (NumberFormatException e) {
+                            context.logger().warning(
+                                    "file with unknown name structure found in the directory: " + file, e);
+                            return false;
+                        }
+                        int localIndexLow = context.memberIndex() * context.localParallelism();
```
cangencer: can you add some comments here explaining what are these?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/FileSinkBuilder.java
```diff
@@ -86,6 +87,8 @@
      * The rolling is based on system time, not on event time. By default no
      * rolling by date is done. If the system clock goes back, the outcome is
      * unspecified and possibly corrupt.
+     *
+     * @since 4.0
      */
     public FileSinkBuilder<T> rollByDate(@Nullable String datePattern) {
```
cangencer: missing nonnull annotation (also check the rest)

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/FileSinkBuilder.java
```diff
@@ -96,6 +99,8 @@
      * Enables rolling by file size. If the size after writing a batch of items
      * exceeds the limit, a new file will be started. From this follows that
      * the file will typically be larger than the given maximum.
+     *
+     * @since 4.0
      */
     public FileSinkBuilder<T> rollByFileSize(@Nullable Long maxFileSize) {
```
cangencer: wouldn't it be better to take primitive `long`? what good does it to do pass null? You could use Long.MAX_VALUE if you don't want rolling by size.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/FileSinkBuilder.java
```diff
@@ -120,17 +125,41 @@
...
+     * If you start all the members on a single machine (such as for
+     * development), set this property to true. If you have multiple machines
+     * with multiple members each and the directory is not a shared storage,
+     * it's not possible to configure the file reader correctly - use only one
+     * member per machine.
+     *
+     * @since 4.0
+     */
+    @Nonnull
+    public FileSinkBuilder<T> sharedFileSystem(boolean sharedFileSystem) {
```
cangencer: can't we always assume it's shared, why require the user to specify it or not? It makes sense for source, but not for sink. Even if it's not shared, you want distinct files you can merge later.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/FileSinkBuilder.java
viliam-durina: You should then provide a constant for the user, the user has to know it or has to read the javadoc. 

Also we add a suffix to the file name when we use rolling, Long.MAX_VALUE says "roll when you reach this impossible to reach size", and it's not the same as "don't use rolline".

I'm not strictly against, but seems clearer to me in current shape.

cangencer: I just find it a bit strange to pass `null`, since you're not required to call the method?

PR #1938, opened 2020-02-06
Improve `mapUsingServiceAsyncBatched` javadoc

### LINKED ISSUE #1909: Cleanup of *usingService(Async) methods

jbartok:
Checklist
- [x] Tags Set
- [x] Milestone Set

Fixes #1909


hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/GeneralStage.java
```diff
@@ -298,6 +300,12 @@
      * Asynchronous version of {@link #mapUsingService}: the {@code mapAsyncFn}
      * returns a {@code CompletableFuture<R>} instead of just {@code R}.
      * <p>
+     * Uses default values for some extra parameters, so the maximum number
+     * of concurrent async operations per processor will be limited to
+     * {@value AbstractAsyncTransformUsingServiceP#MAX_CONCURRENT_OPS} and
```
viliam-durina: The constant should be `DEFAULT_MAX_CONCURRENT_OPS`, I thought it's a maximum configurable number of parallel ops and 256 seemed low...

PR #1945, opened 2020-02-07
Add test for durable subscriber

### LINKED ISSUE #1930: Add stress test for JMS durable topic consumer and exactly-once
We already have stress test for JMS queue and exactly-once. We should also have similar stress test for durable topic.

viliam-durina:
Fixes #1930

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamJmsP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamJmsP.java
@@ -230,8 +230,12 @@ public class StreamJmsP<T> extends AbstractProcessor {
 
     @Override
     public void close() throws Exception {
-        consumer.close();
-        session.close();
+        if (consumer != null) {
+            consumer.close();
+        }
+        if (session != null) {
+            session.close();
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamJmsP.java
```diff
@@ -230,8 +230,12 @@ protected void restoreFromSnapshot(@Nonnull Object key, @Nonnull Object value) {
 
     @Override
     public void close() throws Exception {
-        consumer.close();
-        session.close();
+        if (consumer != null) {
+            consumer.close();
```
cangencer: can this throw?

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/JmsIntegrationTestBase.java
```diff
@@ -383,33 +384,50 @@ public void sinkTopic_whenBuilder_withParameters() throws JMSException {
...
+        stressTest(false, NONE, false);
     }
 
-    private void stressTest(boolean graceful, ProcessingGuarantee maxGuarantee)
+    @Test
+    public void stressTest_exactlyOnce_forceful_durableTopic() throws Exception {
+        stressTest(false, EXACTLY_ONCE, true);
+    }
+
+    private void stressTest(boolean graceful, ProcessingGuarantee maxGuarantee, boolean useTopic)
```
cangencer: I frequently see that we always keep adding parameters to test in this way and it makes it harder to read the test with a lot of conditional logic inside. It would be better if they were refactored in a way that would avoid conditional test logic.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/StreamJmsP.java
viliam-durina: yes, "due to some internal error", as the javadoc says. If it throws, we won't throw the session, but we'll throw the connection in the PSupplier, which should close everything...

PR #1947, opened 2020-02-08
Add migration guide to RefMan

jbartok:
Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/reference-manual/src/main/asciidoc/appendix.adoc
+++ b/reference-manual/src/main/asciidoc/appendix.adoc
@@ -303,32 +303,43 @@ link:https://docs.hazelcast.org/docs/latest/manual/html-single/#migration-guides
 
 The most important changes that have affected Jet are as follows:
 
-* Serializable variants of `java.util.function` interfaces (such as `FunctionEx`, `PredicateEx`) have been moved to IMDG
-and removed from Jet.
-* Generic parameters on several methods and classes were updated/fixed (i.e. EntryProcessor) (see
+* Many packages have been renamed, or classes moved from one package to another and these changes have been followed in
+Jet. For details see link:https://docs.hazelcast.org/docs/rn/index.html#4-0[IMDG Release Notes]. The change most obvious
+in Jet is that many classes have been moved from the generic `com.hazelcast.core` package to more specific packages,
+like `com.hazelcast.map` or `com.hazelcast.collection`.
+* Serializable variants of `java.util.function` interfaces have been moved to IMDG and removed from Jet: `BiConsumerEx`,
+`BiFunctionEx`, `BinaryOperatorEx`, `BiPredicateEx`, `ComparatorEx`, `ComparatorsEx`, `ConsumerEx`, `FunctionEx`,
+`Functions`, `PredicateEx`, `SupplierEx`, `ToDoubleFunctionEx`, `ToIntFunctionEx`, `ToLongFunctionEx`.
+* `Functions` utility class has been moved to IMDG and removed from Jet (from `com.hazelcast.jet.function`
+to `com.hazelcast.function`).
+* Generic parameters on several methods and classes have been updated/fixed (i.e. `EntryProcessor`) (see
 link:https://docs.hazelcast.org/docs/latest/manual/html-single/#introducing-lambda-friendly-interfaces[relevant section]
-in IMDG Migration Guide) and these changes have been adapted to Jet.
-* Some config options are renamed i.e. `group name` is now called `cluster name`. These changes have been adapted,
-also to the Jet Command Line.
-* `EventJournalConfig` is moved from top level config to data-structure specific config.
-* Async methods in IMDG return `CompletionStage` rather than the old `ICompletableFuture` (see
+in IMDG Migration Guide). These changes have been adopted in Jet.
+* `group name` config option is now called `cluster name`. These changes have been adapted to the Jet Command Line too
+(`-g/--groupName` replaced by `-n/--cluster-name` option).
+* `EventJournalConfig` has been moved from top level `Config` class to data-structure specific configs (`MapConfig`,
+`CacheConfig`).
+* Async methods in IMDG return `CompletionStage` now rather than the old `ICompletableFuture` (see
 link:https://docs.hazelcast.org/docs/latest/manual/html-single/#removal-of-icompletablefuture[relevant section] in
-IMDG Migration Guide) and these have been adapted to Jet.
+IMDG Migration Guide) and these have been adopted in Jet.
 
 === Jet API Changes
 Jet 4.0 makes multiple breaking changes in its own APIs:
 
-* `IMapJet`, `ICacheJet` and `IListJet` interfaces are removed, and replaced with IMDG's `IMap` and `IList` instead.
-* `Pipeline.drawFrom` is renamed to `Pipeline.readFrom`.
-* `GeneralStage.drainTo` is renamed to `GeneralStage.writeTo`.
-* `map/filter/flatMapUsingContext` is renamed to `map/filter/flatMapUsingService`.
-* ``ContextFactory`` is renamed to `ServiceFactory`.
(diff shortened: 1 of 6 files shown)


reference-manual/src/main/asciidoc/appendix.adoc
```diff
@@ -283,6 +283,64 @@ The URL used for phone home requests is
...
+* `Pipeline.drawFrom` has been renamed to `Pipeline.readFrom`.
+* `GeneralStage.drainTo` has been renamed to `GeneralStage.writeTo`.
+* `map/filter/flatMapUsingContext` has been renamed to `map/filter/flatMapUsingService`.
+* ``ContextFactory`` has been renamed to `ServiceFactory`.
+* ``mapUsingService``'s `ServiceFactory` now takes `ProcessorSupplier.Context` instead of just `JetInstance`.
+* Signature of `com.hazelcast.jet.pipeline.Sinks#mapWithEntryProcessor` method has changed.
+* HDFS sink and source have been renamed to `Hadoop.inputFormat` and `Hadoop.outputFormat`.
+* `MetricsConfig` is no longer part of `JetConfig`, but resides in the IMDG `Config` class.
+* Generic wildcard parameters in `Traverser` class have been tweaked slightly.
+* Lambdas have been moved to become last parameter, where applicable, for sources and sinks.
```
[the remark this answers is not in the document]
jbartok: Nope. I agree on principle with your "illustrate to make it clear approach" but this one is too simple and it would be too much of an overkill.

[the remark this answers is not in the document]
olukas: ok, agree - as I said `It is probably not needed`.

reference-manual/src/main/asciidoc/appendix.adoc
```diff
@@ -283,6 +283,111 @@ The URL used for phone home requests is
...
+link:https://docs.hazelcast.org/docs/latest/manual/html-single/#migration-guides[IMDG Migration Guides].
+
+The most important changes that have affected Jet are as follows:
+
+* Many packages have been renamed, or classes moved from one package to another and these changes have been followed in
+Jet. For details see link:https://docs.hazelcast.org/docs/rn/index.html#4-0[IMDG Release Notes]. The change most obvious
+in Jet is that many classes have been moved from the generic `com.hazelcast.core` package to more specific packages,
+like `com.hazelcast.map` or `com.hazelcast.collection`.
+
+* Serializable variants of `java.util.function` interfaces have been moved to IMDG and removed from Jet: `BiConsumerEx`,
```
olukas: As I wrote previously I suggest to have notes like `from com.hazelcast.jet.function to com.hazelcast.function` mentioned in this point.

reference-manual/src/main/asciidoc/appendix.adoc
```diff
@@ -283,6 +283,133 @@ The URL used for phone home requests is
 http://phonehome.hazelcast.com/ping
 ----
 
+[appendix]
```
mtopolnik: You should apply line wrapping at 72 throughout.

jbartok: Ok, will do. But curious: why?

cangencer: It's the standard convention in text files, makes it a lot more readable if you don't have word wrapping on. I use rewrap plugin for VSCode.

reference-manual/src/main/asciidoc/appendix.adoc
```diff
@@ -283,6 +284,158 @@ The URL used for phone home requests is
...
+----
+include::{javasource}/migration/Migration.java[tag=serviceFactory2]
+----
+
+* We changed the signature of
+`com.hazelcast.jet.pipeline.Sinks#mapWithEntryProcessor` in order to
+accommodate the improved `EntryProcessor`, which has become more
+lambda-friendly in IMDG (see the
+https://docs.hazelcast.org/docs/latest/manual/html-single/#introducing-lambda-friendly-interfaces[relevant section]
+in the IMDG Migration Guide). The return type of the `EntryProcessor` is
```
Serdaro: no need "the" before `EntryProcessor`

jbartok:
@mtopolnik, just for the record, I do not like your "active voice" rewrite, for following reason (I tried it out and then reverted it for the same reason). As I had it before you could identify a list item most often just from the first 1-2 words of it. Now those first words are always "we moved/rewrote/updated the", so you always have to read at least half a line to start getting some actual information. It was just much more "browsable" as it was.

jbartok:
Have addressed Serdar's comments.

reference-manual/src/main/asciidoc/appendix.adoc
```diff
@@ -283,6 +284,158 @@ The URL used for phone home requests is
...
+include::{javasource}/migration/Migration.java[tag=pipeline2]
+----
+
+* We removed the methods `withMaxPendingCallsPerProcessor(int)` and
+`withUnorderedAsyncResponses()` from `ServiceFactory`. These properties
+are relevant only in the context of asynchronous operations and were used
+in conjunction with `GeneralStage.mapUsingServiceAsync(...)`. In Jet 4.0
+the `GeneralStage.mapUsingServiceAsync(...)` method has a new variant with
+explicit parameters for the above settings.
++
```
cangencer: what about removal of filterUsingServiceAsync and flatMapUsingServiceAsync?

reference-manual/src/main/asciidoc/appendix.adoc
```diff
@@ -283,6 +284,158 @@ The URL used for phone home requests is
...
+
+=== Jet API Changes
+
+We made multiple breaking changes in Jet's own APIs too:
+
+* We removed the Jet-specific interfaces `IMapJet`, `ICacheJet` and
+`IListJet` and now use IMDG's `IMap`, `ICache` and `IList` directly.
+
+* We renamed `Pipeline.drawFrom` to `Pipeline.readFrom` and
+`GeneralStage.drainTo` to `GeneralStage.writeTo`.
```
cangencer: `Pipeline.drawFrom` -> `Pipeline.readTo` and `GeneralStage.drainTo` -> `GeneralStage.writeTo`

reference-manual/src/main/asciidoc/appendix.adoc
jbartok: I was under the impression that those weren't present before 4.0. Will include them.

reference-manual/src/main/asciidoc/appendix.adoc
cangencer: I think I read this wrong, my mistake..

reference-manual/src/main/asciidoc/appendix.adoc
mtopolnik: @jbartok I reworded to move the key words to the beginning, you can check it.

jbartok: @mtopolnik rewording looks good now

jbartok: @cangencer I've added info about removal of async filter & flatmap; pls. take a look, especially the code sample I've added for replacing flatMap, because I'm not sure that what I came up with is the right/best way to do it.

reference-manual/src/main/java/migration/Migration.java
```diff
@@ -0,0 +1,192 @@
...
+
+        StreamStage<Long> stage = p.readFrom(TestSources.itemStream(3))
+                .withoutTimestamps()
+                .map(SimpleEvent::sequence);
+
+        //tag::async1[]
+        stage.mapUsingServiceAsync(serviceFactory,
+                (executor, item) -> {
+                    CompletableFuture<Traverser<String>> f = new CompletableFuture<>();
+                    executor.submit(() -> f.complete(traverseItems(item + "-1", item + "-2", item + "-3")));
```
cangencer: it would be better to return a List<> and then use Traversers.traverseIterable() in the following step

PR #1955, opened 2020-02-10
Shade/relocate picocli library.

gierlachg:
Properly shaded/relocated `picocli` library

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/hazelcast-jet-all/pom.xml
+++ b/hazelcast-jet-all/pom.xml
@@ -70,10 +70,7 @@
                     <artifactSet>
                         <includes>
                             <include>com.hazelcast.jet:*</include>
-                            <include>com.hazelcast:hazelcast</include>
-                            <include>com.hazelcast:hazelcast-aws</include>
-                            <include>com.hazelcast:hazelcast-gcp</include>
-                            <include>com.hazelcast:hazelcast-kubernetes</include>
+                            <include>com.hazelcast:*</include>
                             <include>info.picocli:picocli</include>
                         </includes>
                     </artifactSet>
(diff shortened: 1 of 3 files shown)


hazelcast-jet-all/pom.xml
```diff
@@ -67,6 +67,25 @@
...
                     <createDependencyReducedPom>true</createDependencyReducedPom>
+                    <artifactSet>
+                        <includes>
+                            <include>com.hazelcast.jet:hazelcast-jet-core</include>
+                            <include>com.hazelcast:hazelcast</include>
+                            <include>com.hazelcast:hazelcast-aws</include>
+                            <include>com.hazelcast:hazelcast-gcp</include>
+                            <include>com.hazelcast:hazelcast-kubernetes</include>
+                            <include>org.apache.geronimo.specs:geronimo-jms_1.1_spec</include>
+                            <include>com.rabbitmq.jms:rabbitmq-jms</include>
```
cangencer: why are these here as shaded?

gierlachg: These were already shaded implicitly, I have made it explicit.

cangencer: I realized they're there because of a mistake in the core pom. Should be removed from there as well

hazelcast-jet-all/pom.xml
```diff
@@ -67,6 +67,25 @@
                 <configuration>
                     <createSourcesJar>true</createSourcesJar>
                     <createDependencyReducedPom>true</createDependencyReducedPom>
+                    <artifactSet>
+                        <includes>
+                            <include>com.hazelcast.jet:hazelcast-jet-core</include>
```
cangencer: is it possible to use wildcards here?

PR #1956, opened 2020-02-10
Add error message for invalid class

### LINKED ISSUE #1951: Error message is not descriptive enough when submitting an invalid class
When verbose option is not used the error message is confusing. We should also print the name of the class that's going to be executed.
```
$ bin/jet submit -c invalidclass ~/src/hazelcast-jet-hello/build/libs/hazelcast-jet-hello.jar 
Submitting JAR '/Users/can/src/hazelcast-jet-hello/build/libs/hazelcast-jet-hello.jar' with arguments []
ERROR: invalidclass
```

gurbuzali:
Checklist
- [x] Tags Set
- [x] Milestone Set

Fixes #1951

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetBootstrap.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetBootstrap.java
@@ -110,7 +110,8 @@ public final class JetBootstrap {
                             new URLClassLoader(new URL[]{jarUrl}, JetBootstrap.class.getClassLoader())
             );
 
-            Class<?> clazz = classLoader.loadClass(mainClass);
+            Class<?> clazz = loadMainClass(classLoader, mainClass);
+
             Method main = clazz.getDeclaredMethod("main", String[].class);
             int mods = main.getModifiers();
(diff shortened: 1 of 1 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetBootstrap.java
```diff
@@ -129,6 +130,15 @@ public static synchronized void executeJar(@Nonnull Supplier<JetInstance> suppli
         }
     }
 
+    private static Class<?> loadMainClass(ClassLoader classLoader, String mainClass) throws ClassNotFoundException {
+        try {
+            return classLoader.loadClass(mainClass);
+        } catch (ClassNotFoundException e) {
+            System.err.println("Cannot load main class: " + mainClass);
```
cangencer: should say cannot find or load main class, I think?

nfrankel: Open question: what's the point of writing in the standard err if it's thrown anyway? Won't you get the unfound class name in the exception?

gurbuzali: we have a `verbose` option which is off by default and it was not very descriptive enough in that case

PR #1966, opened 2020-02-12
Fix handling JET_LICENSE_KEY for jet-start.

### LINKED ISSUE #1963: JET_LICENSE_KEY does not work for jet-start
License key for Jet should be able to be configured by `export JET_LICENSE_KEY=JET_ENTERPRISE#<the_rest_of_your_key>` which is impossible now because of issue in https://github.com/hazelcast/hazelcast-jet/blob/b524a7c868b4180dbae1ce9c38229dd774c1c30e/hazelcast-jet-distribution/src/bin-filemode-755/jet-start#L16. Currently it rewrites `JAVA_OPTS_ARRAY` instead of adjusting it. 

However fix is not straight-forward since license should not be printed to server log.

gierlachg:
Fixed handling `JET_LICENSE_KEY` for Unix `jet-start`.

Checklist
- [x] Tags Set
- [x] Milestone Set

Fixes #1963

### DIFF
--- a/hazelcast-jet-distribution/src/bin-regular/common.sh
+++ b/hazelcast-jet-distribution/src/bin-regular/common.sh
@@ -27,8 +27,4 @@ if [ "$JAVA_VERSION" -ge "9" ]; then
     "
 fi
 
-if [ "$JET_LICENSE_KEY" ]; then
-    LICENSING_OPTS="-Dhazelcast.enterprise.license.key=${JET_LICENSE_KEY}"
-fi
-
 CLASSPATH="$JET_HOME/lib/*:$CLASSPATH"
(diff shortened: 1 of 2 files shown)


hazelcast-jet-distribution/src/bin-regular/common.sh
```diff
@@ -28,7 +28,7 @@ if [ "$JAVA_VERSION" -ge "9" ]; then
 fi
 
 if [ "$JET_LICENSE_KEY" ]; then
```
cangencer: I think this actually doesn't need to be in common, since license key is only used on server side, it has no effect for the client. @olukas could you confirm?

olukas: Yes, that's right. Some time ago it was used also on client side but currently license is configured only for server side.

PR #1967, opened 2020-02-12
Added a section regarding ML model prediction

eminn:
Add description here

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/reference-manual/src/main/asciidoc/pipeline-api/pipeline-api.adoc
+++ b/reference-manual/src/main/asciidoc/pipeline-api/pipeline-api.adoc
@@ -424,7 +424,7 @@ Jet pipeline.
 
 == Machine Learning Model Prediction
 Usage of pre-trained Machine Learning models are common to see in the
-pipelines as an external scoring/classification mechanisms. 
+pipelines as external scoring/classification mechanisms. 
 Hazelcast Jet comes with the facility that enables uploading files or 
 directories along with the job for later retrieval in the 
 service objects. The transformation step can take advantage of having a 
@@ -442,7 +442,7 @@ include::{javasource}/MLModelPrediction.java[tag=s1]
 After configuring the job to include the MOJO file, it can be accessed
 in the context object passed in to the service factory via the {jet-javadoc}/core/ProcessorSupplier.Context.html#attachedFile-java.lang.String-[`attachedFile()`]
 or {jet-javadoc}/core/ProcessorSupplier.Context.html#attachedDirectory-java.lang.String-[`attachedDirectory()`]
-method. That method returns the file or directory locally which has been 
+method. That methods returns the file or directory locally which has been 
 uploaded on the job submission. Then, depending on the use case and
 selection of the Machine Learning library, user needs to build the 
 service required for serving the models. 
@@ -454,7 +454,7 @@ include::{javasource}/MLModelPrediction.java[tag=s2]
 
 The service will be passed in the next stage of the transformation along 
 with the stream where you can actually invoke methods on the service to 
-implement scoring/classification. In the example above, the binomial 
+execute scoring/classification. In the example above, the binomial 
 prediction has been done for each item on the stream and the predicted 
 labels are emitted to the downstream.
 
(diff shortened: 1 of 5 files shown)


reference-manual/src/main/java/MLModelPrediction.java
```diff
@@ -0,0 +1,118 @@
...
+        //end::s1[]
+    }
+
+    static void s2() {
+        //tag::s2[]
+        StreamStage<ItemToBeClassified> sourceStage = sourceStage();
+        sourceStage.mapUsingService(ServiceFactory.withCreateContextFn(context -> {
+            String model = context.attachedFile("gbm_mojo").toString();
+            return new EasyPredictModelWrapper(MojoModel.load(model));
+        }).withCreateServiceFn((context, easyPredictModelWrapper) -> easyPredictModelWrapper),
```
viliam-durina: You can use `ServiceFactories.sharedService` and you won't need the `withCreateServiceFn`

reference-manual/src/main/java/MLModelPrediction.java
```diff
@@ -0,0 +1,118 @@
...
+        jobConfig.attachFile(modelFileLocation, "gbm_mojo");
+        Job job = jet.newJob(buildPipeline(), jobConfig);
+        //end::s1[]
+    }
+
+    static void s2() {
+        //tag::s2[]
+        StreamStage<ItemToBeClassified> sourceStage = sourceStage();
+        sourceStage.mapUsingService(ServiceFactory.withCreateContextFn(context -> {
+            String model = context.attachedFile("gbm_mojo").toString();
```
viliam-durina: Might be clearer to assign the result of `attachedFile` to a `File` and do the tostring later when using it.

reference-manual/src/main/asciidoc/pipeline-api/pipeline-api.adoc
```diff
@@ -422,6 +422,43 @@ include::{javasource}/BuildComputation.java[tag=s16a]
...
+service required for serving the models. 
+
+[source]
+----
+include::{javasource}/MLModelPrediction.java[tag=s2]
+----
+
+The service will be passed in the next stage of the transformation along 
+with the stream where you can actually invoke methods on the service to 
+implement scoring/classification. In the example above, the binomial
```
viliam-durina: implement -> execute

reference-manual/src/main/asciidoc/pipeline-api/pipeline-api.adoc
```diff
@@ -422,6 +422,43 @@ include::{javasource}/BuildComputation.java[tag=s16a]
 In a similar fashion you can integrate other external systems with a
 Jet pipeline.
 
+== Machine Learning Model Prediction
+Usage of pre-trained Machine Learning models are common to see in the
+pipelines as an external scoring/classification mechanisms.
```
viliam-durina: "an" must not be used with plural

reference-manual/src/main/asciidoc/pipeline-api/pipeline-api.adoc
```diff
@@ -422,6 +422,43 @@ include::{javasource}/BuildComputation.java[tag=s16a]
...
+
+[source]
+----
+include::{javasource}/MLModelPrediction.java[tag=s1]
+----
+
+After configuring the job to include the MOJO file, it can be accessed
+in the context object passed in to the service factory via the {jet-javadoc}/core/ProcessorSupplier.Context.html#attachedFile-java.lang.String-[`attachedFile()`]
+or {jet-javadoc}/core/ProcessorSupplier.Context.html#attachedDirectory-java.lang.String-[`attachedDirectory()`]
+method. That method returns the file or directory locally which has been
```
olukas: We should mention that this file/directory is stored in `/tmp` locally. 

Should not we also mention some information about attaching big files? Even if handling big files can be from user PoV only technical detail (like that they are split into `128k` chunks and store into maps) users still need some information about how many memory is needed for cluster which will be used for it (i.e. - without implementation knowledge it is not intuitive that with default Jet configuration where each map has 1 backup you actually need 3 nodes with `1GB` memory to load `1GB` file; I think that it can be confusing why just not only 2 nodes with `1GB` memory are not enough).

eminn: May be that detail should go into to the Javadoc of the attachFile/Directory methods ? WDYT ?

olukas: Yes, I think they can be described in Javadoc of those methods.

PR #1975, opened 2020-02-13
(docs) Add kubernetes section to operations

gurbuzali:


site/docs/operations/kubernetes.md
```diff
@@ -3,4 +3,74 @@ title: Jet on Kubernetes
 id: kubernetes
 ---
 
-TODO
+The easiest way to install **Hazelcast Jet** to **Kubernetes** is using
```
cangencer: I think we need an intro section to conceptually how Hazelcast works on Kubernetes (how discovery is made) and also concerns like how to access the cluster from outside Kubernetes.

Also talk about auto-scaling, and other concerns related to Kubernetes.

PR #1985, opened 2020-02-17
Add serialization sample implementation.

gierlachg:
Added serialization sample implementation.


eminn:
Do you know why it shows 14 commits ?

gierlachg:
Not really, I might have messed up something with branching/merging... but the change set seems to be correct.

site/docs/api/serialization.md
```diff
@@ -84,11 +77,49 @@ Strategy                                                   Number of Bytes   Ove
...
+        }
+    }
+```
+
+Then the serializer should be registered with Jet on cluster startup:
+```
+JetConfig config = new JetConfig();
+config.getHazelcastConfig().getSerializationConfig()
+    .addSerializerConfig(new SerializerConfig().setTypeClass(Person.class).setClass(PersonSerializer.class));
+JetInstance jet = Jet.newJetInstance(config);
```
cangencer: I think worth mentioning is SerializerHooks, and also how to mention it when using a standalone cluster

PR #1988, opened 2020-02-18
Check init script return code

### LINKED ISSUE #1987: Failure in init.sh/cleanup.sh for Python mapping is ignored
In case when `init.sh`/`cleanup.sh` script for Python mapping finished with error we should fail-fast - job should finish with failure.

In current behavior it seems that any error is ignored and job is started even if bash script failed. It does not work for cases when script is failed expectedly by `exit 1` or even if there is syntax error in script.

cangencer:
Fixes #1987

Checklist
- [x] Tags Set
- [x] Milestone Set

Fixes #1987

### DIFF
--- a/extensions/python/src/main/java/com/hazelcast/jet/python/PythonServiceContext.java
+++ b/extensions/python/src/main/java/com/hazelcast/jet/python/PythonServiceContext.java
@@ -95,7 +95,7 @@ class PythonServiceContext {
                 initProcess.waitFor();
                 if (initProcess.exitValue() != 0) {
                     throw new Exception(
-                            "Initialization script finished with non-zero return code: " + initProcess.exitValue()
+                            "Initialization script finished with non-zero exit code: " + initProcess.exitValue()
                     );
                 }
                 stdoutLoggingThread.join();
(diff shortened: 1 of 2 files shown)


cangencer:
we can't fail the job if cleanup.sh fails, because it's called at the completion stage. We can make it log a warning. Calling cleanup if init.sh is a bit problematic as well, because cleanup may be reliant on something being done in init.

olukas:
Regarding `cleanup.sh` - that's right that it cannot fail the job since it is running after job is finished but we still should log a warning for it.

Regarding running `cleanup.sh` if `init.sh` failed - `cleanup.sh` can be use for clean up some resources which were allocated during init. I think it can cause troubles to users to do not execute cleanup since it can leak some resources etc.

olukas:
If it is really problematic to run `cleanup.sh` after failure in init phase (`init.sh` or `requirements.txt`) we can document it in reference manual. However I think it is not convenient for users.

cangencer:
I've realized if we don't do the destroy part the temp folders won't be deleted. So I've changed that if init.sh we explicitly destroy. If it fails elsewhere it means some internal error in Jet (i.e. NullPointerException or similar), in this case no special action is taken.

PR #2013, opened 2020-02-26
Unify various trade sources, extract in separate module

jbartok:
We want the Trade Generator source to be clearly defined, unique and published on Maven so that we can use in tutorials as a realistic stream source.

Checklist
- [ ] Tags Set
- [ ] Milestone Set


examples/early-window-results/src/main/java/com/hazelcast/jet/examples/earlyresults/TradingVolumeOverTime.java
```diff
@@ -53,17 +53,18 @@
...
         Pipeline p = Pipeline.create();
-        p.readFrom(TradeGenerator.tradeSource(TRADES_PER_SEC, DURATION_SECONDS))
-         .withNativeTimestamps(TradeGenerator.MAX_LAG)
+        p.readFrom(TradeGenerator.tradeSource(1, TRADES_PER_SEC, MAX_LAG, DURATION_SECONDS))
+         .withNativeTimestamps(MAX_LAG)
          .window(tumbling(SECONDS.toMillis(2))
                  // comment out this line to see how the chart behaves without early results:
                  .setEarlyResultsPeriod(20)
          )
-         .aggregate(summingLong(trade -> trade.getQuantity() * trade.getPrice()))
```
eminn: Is the change in the logic intended ?

examples/rolling-aggregation/src/main/java/com/hazelcast/jet/examples/rollingaggregation/TradingVolume.java
```diff
@@ -18,34 +18,36 @@
...
     private static final int DURATION_SECONDS = 60;
 
     private static Pipeline buildPipeline() {
         Pipeline p = Pipeline.create();
-        p.readFrom(tradeSource(NUMBER_OF_TICKERS, TRADES_PER_SEC, DURATION_SECONDS))
+        p.readFrom(tradeSource(NUMBER_OF_TICKERS, TRADES_PER_SEC, MAX_LAG, DURATION_SECONDS))
          .withoutTimestamps()
          .groupingKey(Trade::getTicker)
-         .rollingAggregate(summingLong(Trade::getPrice))
+         .rollingAggregate(summingLong(Trade::getQuantity))
```
eminn: same here ?

examples/early-window-results/src/main/java/com/hazelcast/jet/examples/earlyresults/TradingVolumeOverTime.java
jbartok: Yes. Since the original `TradeGenerators` emitted different kinds of trades I had to modify the examples to show the same TYPE of behaviour as before, even when moved to the new, common generator.

examples/rolling-aggregation/src/main/java/com/hazelcast/jet/examples/rollingaggregation/TradingVolume.java
jbartok: Yes, same. Change done to preserve overall behaviour and GUI looks.

cangencer:
verify

PR #2020, opened 2020-02-26
Support job-level serializers in ReceiverTasklet

gierlachg:
This PR is part of the effort to add job level serialization support.

In current state of serialization API, `BufferObjectDataInput` encapsulates `SerializationService`. Every object is read from `ObjectDataInput` with the help of the _service_ set at the time of creation of the _input_, in particular every nested object .

`ReceiverTasklet` implicitly, through `BufferObjectDataInput` uses _global_ `SerializationService`. To be able to use job level serializers, the _input_ should be created in `ReceiverTasklet` using job scoped `SerializationService`.

`Networking` has been decoupled from `SerializationService` - `MemoryDataInput` & `MemoryDataOutput` have been created to handle packets reading & writing, this part is Jet specific and we don't need to rely on IMDG here.
As a temporary measure, `CustomInputOutputFactory` has been created to give access to package protected IMDG _inputs_ - a [PR](https://github.com/hazelcast/hazelcast/pull/16701) to remediate lack of proper API has been already submitted.

Checklist
- [x] Tags Set
- [x] Milestone Set


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/serialization/ByteArrayDataInput.java
```diff
@@ -0,0 +1,67 @@
...
+    public long readLong() {
+        checkAvailable(Long.BYTES);
+        long value = Bits.readLong(buffer, position, false);
+        position += Long.BYTES;
+        return value;
+    }
+
+    private void checkAvailable(int length) {
+        if (position + length > buffer.length) {
+            throw new RuntimeException("Cannot read " + length + " bytes!");
```
viliam-durina: Don't use exclamation marks in exceptions :). No need to shout at the user. Also no trailing periods...

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/serialization/ByteArrayDataInput.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    public int position() {
+        return position;
+    }
+
+    @Override
+    public BufferObjectDataInput toObjectInput(AbstractSerializationService serializationService) {
+        byte[] bytes = new byte[buffer.length - position];
```
viliam-durina: Can't we reuse the buffer if position is 0? I guess it's common.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/serialization/DataInput.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.jet.impl.serialization;
+
+import com.hazelcast.internal.nio.BufferObjectDataInput;
+import com.hazelcast.internal.serialization.impl.AbstractSerializationService;
+
+public interface DataInput {
```
viliam-durina: This is confusing with the standard `java.io.DataInput`. We should use different name, esp. when the purpose is similar. Maybe `JetDataInput`. I was confused how come you implement just few methods when `DataInput` has many more.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/serialization/ByteArrayDataInputTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+public class ByteArrayDataInputTest {
+
+    @Test
+    public void whenNotEnoughBytesToRead_thenThrowsException() {
```
viliam-durina: We use `when_AAA_then_BBB`, the underscore is after _when_ and _then_.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/Networking.java
```diff
@@ -17,33 +17,35 @@
...
-import static com.hazelcast.jet.impl.util.ImdgUtil.createObjectDataOutput;
 import static com.hazelcast.jet.impl.util.ImdgUtil.getMemberConnection;
 import static com.hazelcast.jet.impl.util.ImdgUtil.getRemoteMembers;
 import static com.hazelcast.jet.impl.util.Util.uncheckRun;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 public class Networking {
+
+    private static final int PACKET_HEADER_SIZE = 16;
+    private static final int FLOW_PACKET_INITIAL_SIZE = 128; // TODO:
```
gierlachg: The default was 1 << 15... which seems a little bit excessive. Is there any sensible average I could default to?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/Networking.java
```diff
@@ -63,33 +63,29 @@ void shutdown() {
...
         handleFlowControlPacket(packet.getConn().getEndPoint(), packet.toByteArray());
     }
 
-    private void handleStreamPacket(Packet packet) throws IOException {
-        BufferObjectDataInput in = createObjectDataInput(nodeEngine, packet.toByteArray());
-        long executionId = in.readLong();
-        int vertexId = in.readInt();
-        int ordinal = in.readInt();
+    private void handleStreamPacket(Packet packet) {
+        MemoryDataInput input = new MemoryDataInput(packet.toByteArray());
```
gierlachg: This is one additional allocation, if we are concerned about it we can use `MemoryWriter` directly in trade of readability/maintainability... WDYT?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetService.java
```diff
@@ -232,11 +230,7 @@ public ClassLoader getClassLoader(long jobId) {
     }
 
     void handlePacket(Packet packet) {
-        try {
-            networking.handle(packet);
-        } catch (IOException e) {
-            throw sneakyThrow(e);
-        }
+        networking.handle(packet);
```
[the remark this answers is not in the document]
gierlachg: The callee might throw `RuntimeException`.

hazelcast-jet-core/src/main/java/com/hazelcast/internal/serialization/impl/CustomInputOutputFactory.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+package com.hazelcast.internal.serialization.impl;
+
+import com.hazelcast.internal.nio.BufferObjectDataInput;
+import com.hazelcast.internal.serialization.InternalSerializationService;
+
+import java.nio.ByteOrder;
+
+// remove when `InternalSerializationService.createObjectDataInput(byte[] data, int offset)` is there...
```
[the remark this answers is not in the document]
cangencer: what about a PR for 4.0.1-SNAPSHOT?

PR #2023, opened 2020-02-27
Fix minor problems in Refman

mtopolnik:
Checklist
- [x] Tags Set
- [x] Milestone Set


cangencer:
this is currently against master, can you put it against 4.0?

PR #2043, opened 2020-03-06
Support job-level serializers

gierlachg:
This PR is part of the effort to add job level serialization support.

_Global_ `SerializationService` does not provide any measure to manage (register, deregister, retrieve by some contextId etc.) scoped serializers.

In order to use job-level serializers, `JetSerializationService` has been introduced. It is basically a copy of `AbstractSerializationService` and hopefully will be removed/refactored when [this](https://github.com/hazelcast/hazelcast/pull/16722) is merged. 
It is created for each job execution, maintains its own collection of serializers (that are registered with the given Job) and delegates to the _global_ `SerializationService` if serializer for a given class is not found.

Apart from that, `addSerializer()` has been added to `JobConfig` to allow job-level serializres registration.

Checklist
- [x] Tags Set
- [x] Milestone Set


hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -927,6 +923,44 @@ private static File urlToFile(@Nonnull URL url) {
...
+     *
+     * @implNote Backing storage for this method is an {@link IMap} with a
+     * default backup count of 1. When adding big files as a resource, size
+     * the cluster accordingly in terms of memory, since each file will have 2
+     * copies inside the cluster(primary + backup replica).
+     *
+     * @return {@code this} instance for fluent API
+     */
+    @Nonnull
+    public <T, S extends StreamSerializer<?>> JobConfig addSerializer(Class<T> clazz, Class<S> serializerClass) {
```
cangencer: should we not offer a more convenient interface here? like Serializer<T> or similar?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ExecutionContext.java
```diff
@@ -129,13 +137,19 @@ public ExecutionContext initialize(ExecutionPlan plan) {
         snapshotContext = new SnapshotContext(nodeEngine.getLogger(SnapshotContext.class), jobNameAndExecutionId(),
                 plan.lastSnapshotId(), jobConfig.getProcessingGuarantee());
 
+        serializationService = new JetSerializationService(
```
viliam-durina: Maybe we can avoid this if there are no serializers configured.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -927,6 +923,44 @@ private static File urlToFile(@Nonnull URL url) {
...
+     * @implNote Backing storage for this method is an {@link IMap} with a
+     * default backup count of 1. When adding big files as a resource, size
+     * the cluster accordingly in terms of memory, since each file will have 2
+     * copies inside the cluster(primary + backup replica).
+     *
+     * @return {@code this} instance for fluent API
+     */
+    @Nonnull
+    public <T, S extends Serializer> JobConfig addSerializer(Class<T> clazz, Class<S> serializerClass) {
+        addClass(clazz);
```
viliam-durina: We should not add classes. We should just register the serializer and the user is responsible for having the classes on classpath. There are other ways to make the class available at runtime, this is the least common one.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/config/ResourceConfigTest.java
```diff
@@ -1064,6 +1070,48 @@ public void when_attachAll() throws Exception {
         assertEquals(dir.toURI().toURL(), dirConfig.getUrl());
     }
 
+    @Test
+    public void when_addSerializer() {
```
viliam-durina: This should be in `JobConfigTest`

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/deployment/JobLevelSerializerIsAvailable.java
```diff
@@ -0,0 +1,45 @@
...
+    @Override
+    protected void init(@Nonnull Context context) {
+        ClassLoader cl = Thread.currentThread().getContextClassLoader();
+        try {
+            Class<?> clazz = cl.loadClass(VALUE_CLASS_NAME);
+            Object value = clazz.getDeclaredConstructor().newInstance();
+            // We serialize an object so the job level serializer is invoked.
+            byte[] bytes = ((ProcCtx) context).serializationService().toBytes(value);
+            assertArrayEquals(bytes, new byte[]{0, 0, 0, 0, 0, 0, 0, 42});
+        } catch (Exception e) {
```
viliam-durina: Remove the try-catch and just declare the exception.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/deployment/AbstractDeploymentTest.java
```diff
@@ -323,6 +328,23 @@ public void testDeployment_whenFileAddedAsResource_thenAvailableOnClassLoader()
         executeAndPeel(getJetInstance().newJob(dag, jobConfig));
     }
 
+    @Test
+    @SuppressWarnings("unchecked")
+    public void testDeployment_whenAddSerializer_thenAvailableOnJobContext() throws Throwable {
```
viliam-durina: We should not test the `addClass` functionality, but that the Jet can actually use the serializer. The test could have 2 vertices connected by a distributed edge. The source will generate a non-serializable object for which a custom serializer is used. The cluster must have at least 2 members, otherwise serialization isn't used.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
gierlachg: agreed with @vbekiaris that `StreamSerializer` should be enough, `ByteArraySerializer` is more of a private thingy

hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/deployment/AbstractDeploymentTest.java
viliam-durina: I'd just remove this test. It only tests job class loading.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ExecutionContext.java
gierlachg: We need to call `destroy()` at job completion, that would require another `if` in `completeExecution`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -927,6 +924,40 @@ private static File urlToFile(@Nonnull URL url) {
...
+     * {@code IMap} data source, which can instantiate only the classes from
+     * the Jet instance's classpath.)
+     * <p>
+     * Serializer must have no-arg constructor.
+     *
+     * @return {@code this} instance for fluent API
+     */
+    @Nonnull
+    @EvolvingApi
+    public <T, S extends StreamSerializer<?>> JobConfig registerStreamSerializer(@Nonnull Class<S> serializerClass,
```
cangencer: can we rename method to `registerSerializerFor` and have the parameters reversed. `JobConfig.registerSerializerFor(Value.class, ValueSerializer.class)`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -65,6 +67,7 @@
     private boolean storeMetricsAfterJobCompletion;
 
     private Map<String, ResourceConfig> resourceConfigs = new LinkedHashMap<>();
+    private Map<String, String> serializerConfigs = new LinkedHashMap<>();
```
viliam-durina: Do we need to preserve the iteration order? `HashMap` is cheaper. For `resourceConfigs` it's probably important because two added JARs can have duplicate classes, and the first one should be used, but here we don't allow duplicates.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/JobConfig.java
```diff
@@ -927,6 +924,45 @@ private static File urlToFile(@Nonnull URL url) {
...
+     * Serializers registered on a job level have precedence over any serializer
+     * registered on a cluster level.
+     * <p>
+     * Serializer must have no-arg constructor.
+     *
+     * @return {@code this} instance for fluent API
+     */
+    @Nonnull
+    @EvolvingApi
+    public <T, S extends StreamSerializer<?>> JobConfig registerSerializerFor(@Nonnull Class<T> clazz,
```
viliam-durina: My opinion: can be named just `registerSerializer`. The "for" isn't needed.

PR #2044, opened 2020-03-06
Improve trade source

jbartok:
Our newly extracted `hazelcast-jet-examples-trade-source` module has some problems which we need to fix before we can release it.

Checklist
- [x] Tags Set
- [x] Milestone Set


jbartok:
> can we add javadoc?

Yes we can, I just didn't get to it yet. Also, proper serialisation, per the comment in `Trade` is planned.

cangencer:
> > can we add javadoc?
> 
> Yes we can, I just didn't get to it yet. Also, proper serialisation, per the comment in `Trade` is planned.

I would leave it as Serializable for simplicity sake.

PR #2047, opened 2020-03-09
Disable logging details by default

cangencer:
Do not write version number, ip and cluster name on every log message by default

This option can be turned on again by  setting `hazelcast.logging.details.enabled` to `true`

Checklist
- [x] Tags Set
- [x] Milestone Set


cangencer:
I think there's a separate issue for that, could we do in a separate PR?

eminn:
It's fine to do it in a separate PR, I've forgot that the issue exists. LGTM

PR #2052, opened 2020-03-10
Exclude classgraph from javadocs

cangencer:
It was leftover in the generated javadoc

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -114,6 +114,7 @@
         <grpc.version>1.26.0</grpc.version>
         <protobuf.version>3.11.3</protobuf.version>
         <picocli.version>3.9.0</picocli.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -114,6 +114,7 @@
         <grpc.version>1.26.0</grpc.version>
         <protobuf.version>3.11.3</protobuf.version>
         <picocli.version>3.9.0</picocli.version>
+        <classgraph.version>4.8.62</classgraph.version>
```
olukas: We should also use this version in `hazelcast-jet-core/pom.xml` for `classgraph` instead of hard-coded version as it is now.

cangencer: yes forgot to change it, will update.

PR #2061, opened 2020-03-13
Improve serialization exceptions

gierlachg:
Improved the way how serialization exceptions are handled - giving a hint to the user about `JobConfig.registerSerializer()`

Checklist
- [x] Tags Set
- [x] Milestone Set

Fixes hazelcast/hazelcast-jet-enterprise#109


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/serialization/DelegatingSerializationService.java
```diff
@@ -108,33 +108,57 @@ public PortableContext getPortableContext() {
...
         if (serializer == null) {
-            serializer = delegate.serializerFor(typeId);
-            if (serializer == null) {
-                throw active ?
-                        newHazelcastSerializationException(typeId) :
-                        new HazelcastInstanceNotActiveException();
+            try {
+                serializer = delegate.serializerFor(typeId);
+            } catch (HazelcastSerializationException hse) {
+                throw handle(typeId, hse);
```
viliam-durina: We should not depend on the exact text from the superclass. We also lose the stack trace of the original exc.

I'd rather just wrap the caught exception, use the original message and append a text to it.
```suggestion
                throw new HazelcastSerializationException(hse.getMessage() + ". You can register a job-specific serializer using JobConfig.registerSerializer()", hse);
```

If we also dispose the delegate (which we should), we wouldn't have to handle the inactive case: the delegate will throw `HazelcastInstanceNotActiveException` directly.

gierlachg: We should not `dispose()` delegate as it's a _global_ serializer. I have addressed the remaining comment.

PR #2069, opened 2020-03-18
Add test for non-existing package for JobConfig

olukas:
Added regression test for non-existing package passed in `JobConfif.addPackage()`.

Checklist
- [x] Tags Set
- [x] Milestone Set
- [N/A] Any breaking changes are documented
- [N/A] New public APIs have `@Nonnull/@Nullable` annotations
- [N/A] New public APIs have `@since` tags in Javadoc
- [N/A] For code samples, code sample main readme is updated

### DIFF
--- a/hazelcast-jet-core/src/test/java/com/hazelcast/jet/config/ResourceConfigTest.java
+++ b/hazelcast-jet-core/src/test/java/com/hazelcast/jet/config/ResourceConfigTest.java
@@ -88,6 +88,16 @@ public class ResourceConfigTest extends JetTestSupport {
     }
 
     @Test
+    public void when_addResourcesWithNonExistingPackage() {
+        // When
+        config.addPackage("thispackage.does.not.exist");
+
+        // Then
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/test/java/com/hazelcast/jet/config/ResourceConfigTest.java
```diff
@@ -87,6 +87,16 @@ public void when_addResourcesWithPackage() {
                 ));
     }
 
+    @Test
+    public void when_addResourcesWithNonExistingPackage() {
+        // When
+        config.addPackage("this.package.is.not.exist");
```
gierlachg: this.package.is.not.exist -> this.package.does.not.exist ?

olukas: make sense, I'll also change it to valid package name, so I'll use `thispackage.does.not.exist`

PR #2070, opened 2020-03-18
Fix code sample in "Streaming Source" section

emre-aydin:
Fix compile error in the code sample due to wrong method name and missing method argument


cangencer:
Thanks @emre-aydin !

PR #2082, opened 2020-03-20
Enable job-level serializer for IMDG Lists

gierlachg:
Allowed job-level serializers to be used with IMDG `List`s.

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadIListP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadIListP.java
@@ -28,14 +28,11 @@ import com.hazelcast.jet.core.AbstractProcessor;
 import com.hazelcast.jet.impl.execution.init.Contexts.ProcCtx;
 
 import javax.annotation.Nonnull;
-import java.util.Iterator;
 import java.util.List;
 import java.util.function.BiFunction;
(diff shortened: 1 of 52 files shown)


gierlachg:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadIListP.java
```diff
@@ -52,18 +64,40 @@ public boolean isCooperative() {
...
+            return createTraverser(size, proxy::dataIterator, proxy::dataSubList);
+        } else {
+            throw new RuntimeException("Unexpected list class: " + list.getClass().getName());
+        }
+    }
+
+    private Traverser<Data> createTraverser(int size,
+                                            Supplier<Iterator<Data>> iteratorSupplier,
+                                            BiFunction<Integer, Integer, List<Data>> subListSupplier) {
+        return size <= FETCH_SIZE ?
```
cangencer: did we really need both data iterator and sublist here? seems unnecessary

gierlachg: I wanted to keep the existing behavior but maybe it could be simplified indeed.... :thinking:

PR #2120, opened 2020-03-31
Add job-level serializer & addClass/addPackage docs

gierlachg:
Added job-level serializer & `addClass()`/`addPackage()` docs.

### DIFF
--- a/site/docs/api/serialization.md
+++ b/site/docs/api/serialization.md
@@ -226,26 +226,26 @@ Below you can find a sample implementation of `StreamSerializer` for
 ```java
 class PersonSerializer implements StreamSerializer<Person> {
 
-        private static final int TYPE_ID = 1;
+    private static final int TYPE_ID = 1;
 
-        @Override
-        public int getTypeId() {
-            return TYPE_ID;
-        }
-
-        @Override
-        public void write(ObjectDataOutput out, Person person) throws IOException {
-            out.writeUTF(person.firstName);
-            out.writeUTF(person.lastName);
-            out.writeInt(person.age);
-            out.writeFloat(person.height);
-        }
-
-        @Override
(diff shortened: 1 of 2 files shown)


site/docs/api/serialization.md
```diff
@@ -294,3 +294,28 @@ hazelcast:
 All the classes - data types, serializers & hooks - should be present
 on the server classpath, ideally in server's `lib` directory packaged as
 a jar file.
+
+### Job-level Serializers
```
cangencer: wouldn't it be better to mention this earlier in the document?

site/docs/api/serialization.md
```diff
@@ -216,44 +214,68 @@ not to mention very wasteful of memory usage.
...
+One way of registering a custom serializer is to do that on a job
+level. Assuming both, value and serializer classes are already added to
+the [classpath](submitting-jobs.md), you can simply:
+
+```java
+new JobConfig()
+    .registerSerializer(Person.class, PersonSerializer.class)
+```
+
+Such serializer is scoped - the type id does not have to be globally
```
cangencer: but it can still conflict with other global serializers, right?

site/docs/api/serialization.md
```diff
@@ -292,5 +314,8 @@ hazelcast:
 ```
 
 All the classes - data types, serializers & hooks - should be present
-on the server classpath, ideally in server's `lib` directory packaged as
-a jar file.
+on the cluster classpath, ideally in server's `lib` directory packaged as
+a jar file. Moreover, used type ids have to be unique across all
+serializers.
+Despite those limitations cluster level serializers offer
+full support for IMDG [sources and sinks](sources-sinks.md).
```
cangencer: we should mention this is required for supporting querying, projections and so on.

PR #2124, opened 2020-03-31
Add Protocol Buffers serialization docs

gierlachg:
Added Protocol Buffers serialization docs.


site/docs/api/serialization.md
```diff
@@ -321,3 +321,56 @@ a jar file. Moreover, used type ids have to be unique across all
...
+```xml
+<dependency>
+    <groupId>com.hazelcast.jet</groupId>
+    <artifactId>hazelcast-jet-protobuf</artifactId>
+    <classifier>jar-with-dependencies</classifier>
+    <version>${hazelcast.jet.version}</version>
+</dependency>
+```
+
+If you want to use it with just a job, implement provided adapter:
```
cangencer, line 346: do you need to add any additional classes ? (Protobuf specific) or just registering serializer is enough, without calling addClass ?

gierlachg: You need to `addJar()` the extension obviously.

cangencer: this should be in the docs

gierlachg: Yes, it shades the protobuf lib.

site/docs/api/serialization.md
```diff
@@ -315,9 +316,58 @@ hazelcast:
...
+we encourage registering a Protobuf-specific serializer. There is a Jet
+extension module that simplifies this for Protobuf version 3.
+
+Add the extension as a dependency to your Jet job's Maven project:
+
+```xml
+<dependency>
+    <groupId>com.hazelcast.jet</groupId>
+    <artifactId>hazelcast-jet-protobuf</artifactId>
+    <classifier>jar-with-dependencies</classifier>
```
cangencer, line 342: since it's reworded, we don't need to depend on the fat JAR here, and we should also include the equivalent gradle snippet as the first option (this is the convention throughout the documentation)

site/docs/api/serialization.md
```diff
@@ -315,9 +316,58 @@ hazelcast:
...
+    <version>${hazelcast.jet.version}</version>
+</dependency>
+```
+
+If you want to use it locally within a Jet job, implement an adapter
+by extending the provided class:
+
+```java
+class PersonSerializer extends ProtobufSerializer<Person> {
```
cangencer, line 351: wouldn't it be easier to use the registerSerializer(ProtobufSerializer.from(..) option?

gierlachg: `registerSerializer()` takes serializer by class right now

cangencer: would it make sense to support instance as well?

site/docs/api/serialization.md
mtopolnik: We actually want to support both approaches, with the extension installed on the Jet cluster and used just inside a job. We should update the docs in that direction.

site/docs/api/serialization.md
gierlachg: I guess we could do that, wanted to keep it simple.

gierlachg: However, in case of Protobuf adapter which isn't `Serializable` (`Parser` is not `Serializable`) that would not be very useful.

cangencer: ok I see. 👍

Issue #2130, opened 2020-04-01
Classgraph library sometimes freezes on addClass


frant-hartm:
This test fails on master since d98b229d commit. There were no related code changes in the commit, but a quite significant change in dependencies on the test classpath.

This is likely an issue with the Classgraph library, which can't handle some of the new dependencies well.

```
    @Test(timeout = 10_000)
    public void shouldFinishSubmittedJob() {
        JetInstance member = createJetMember();

        Pipeline p = Pipeline.create();
        p.readFrom(TestSources.items(1, 2, 4))
         .writeTo(logger());

        JobConfig config = new JobConfig();
        config.addClass(MyClass.class); // ClassGraph library chokes up here
        member.newJob(p, config).join();
    }
```

The stacktrace for main thread
```
"Time-limited test" #13 daemon prio=5 os_prio=0 tid=0x00007f81ecee1800 nid=0x1add waiting on condition [0x00007f818f06e000]
   java.lang.Thread.State: WAITING (parking)
	at sun.misc.Unsafe.park(Native Method)
	- parking to wait for  <0x00000000dc584a18> (a java.util.concurrent.FutureTask)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.FutureTask.awaitDone(FutureTask.java:429)
	at java.util.concurrent.FutureTask.get(FutureTask.java:191)
	at io.github.classgraph.ClassGraph.scan(ClassGraph.java:1247)
	at io.github.classgraph.ClassGraph.scan(ClassGraph.java:1287)
	at io.github.classgraph.ClassGraph.scan(ClassGraph.java:1300)
	at com.hazelcast.jet.impl.util.ReflectionUtils.nestedClassesOf(ReflectionUtils.java:73)
	at com.hazelcast.jet.config.JobConfig.addClass(JobConfig.java:249)
	at com.hazelcast.jet.test.hierarchy.TestAddClass.shouldFinishSubmittedJob(TestAddClass.java:23)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #2131, opened 2020-04-01
Bump Classgraph version to 4.8.66

gierlachg:
Bumped Classgraph version to 4.8.66.

Checklist
- [x] Tags Set
- [x] Milestone Set

Fixes https://github.com/hazelcast/hazelcast-jet/issues/2130 by https://github.com/classgraph/classgraph/issues/400


gierlachg:
I would say that it would be testing Classgraph itself which we don't really want to do.

PR #2135, opened 2020-04-02
Fix race asserting RUNNING status and restarting

### LINKED ISSUE #2132: com.hazelcast.jet.impl.connector.WriteJdbcPTest.test_transactional_withRestarts_graceful_exOnce
_master_ (commit 52cb615b31faae9dea91a34d25ef13f358a9cd9e)

Failed on Zulu JDK 8: http://jenkins.hazelcast.com/job/jet-oss-master-zulu-jdk8/188/testReport/junit/com.hazelcast.jet.impl.connector/WriteJdbcPTest/test_transactional_withRestarts_graceful_exOnce/

Stacktrace:
```
java.lang.IllegalStateException: Cannot RESTART_GRACEFUL, job status is NOT_RUNNING, should be RUNNING
	at com.hazelcast.jet.impl.JobCoordinationService.lambda$terminateJob$11(JobCoordinationService.java:347)
	at com.hazelcast.jet.impl.JobCoordinationService.lambda$toNullFunction$33(JobCoordinationService.java:575)
	at com.hazelcast.jet.impl.JobCoordinationService.lambda$callWithJob$34(JobCoordinationService.java:607)
	at com.hazelcast.jet.impl.JobCoordinationService.lambda$submitToCoordinatorThread$43(JobCoordinationService.java:997)
	at com.hazelcast.internal.util.executor.CompletableFutureTask.run(CompletableFutureTask.java:64)
	at com.hazelcast.internal.util.executor.CachedExecutorServiceDelegate$Worker.run(CachedExecutorServiceDelegate.java:217)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at com.hazelcast.internal.util.executor.HazelcastManagedThread.executeRun(HazelcastManagedThread.java:64)
	at com.hazelcast.internal.util.executor.HazelcastManagedThread.run(HazelcastManagedThread.java:80)
```

Standard output:
```
Hiccups measured while running test 'test_transactional_withRestarts_graceful_exOnce(com.hazelcast.jet.impl.connector.WriteJdbcPTest):'
01:55:30, accumulated pauses: 66 ms, max pause: 12 ms, pauses over 1000 ms: 0
01:55:35, accumulated pauses: 117 ms, max pause: 67 ms, pauses over 1000 ms: 0
```

viliam-durina:
Fixes #2132

### DIFF
--- a/hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/JetTestSupport.java
+++ b/hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/JetTestSupport.java
@@ -29,6 +29,7 @@ import com.hazelcast.jet.config.JetConfig;
 import com.hazelcast.jet.function.RunnableEx;
 import com.hazelcast.jet.impl.JetService;
 import com.hazelcast.jet.impl.JobExecutionRecord;
+import com.hazelcast.jet.impl.JobExecutionService;
 import com.hazelcast.jet.impl.JobRepository;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.Logger;
@@ -52,6 +53,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 
 import static com.hazelcast.jet.Util.idToString;
+import static com.hazelcast.jet.core.JobStatus.RUNNING;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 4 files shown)


hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/JetTestSupport.java
```diff
@@ -146,6 +148,44 @@ public static void assertJobStatusEventually(Job job, JobStatus expected) {
...
+     *     job.restart();
+     *     assertJobRunningEventually(instance, job, oldExecutionId);
+     *     // now we're sure that a new execution is running
+     * }</pre>
+     *
+     * @param ignoredExecutionId If job is running and has this execution ID,
+     *      wait longer. If null, no execution ID is ignored.
+     * @return the execution ID of the new execution
+     */
+    public static long assertJobRunningEventually(JetInstance instance, Job job, Long ignoredExecutionId) {
```
olukas, line 177: Can we rename this method please? It does not assert whether job is running but whether job was restarted (and running). 

If this method is called independently on `job.restart();` it will never return. Should not be rather `job.restart();` called inside this method to have it more clear and robust?

viliam-durina: It does assert that the job is running, but it can **optionally** exclude certain execution ID. If the given execution ID is null, it ignores no execution ID. If it is not null, it's useful after a restart to ignore the previous execution.

What name do you suggest?

olukas: You're right, I misread condition. What about the name like `assertJobRunningAfterRestartEventually` ? However since `ignoredExecutionId` can be `null` then name which I suggest is also bad. Maybe let's keep that name as it is.

PR #2142, opened 2020-04-06
[005] Productionize CDC sources

jbartok:
Checklist
- [x] Tags Set
- [x] Milestone Set


extensions/cdc-mongodb/pom.xml
```diff
@@ -0,0 +1,119 @@
...
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.hazelcast.jet</groupId>
+            <artifactId>hazelcast-jet-cdc-core</artifactId>
+            <version>4.1-SNAPSHOT</version>
```
eminn, line 79: ${project.parent.version}

extensions/cdc-mysql/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.hazelcast.jet</groupId>
+            <artifactId>hazelcast-jet-cdc-core</artifactId>
+            <version>4.1-SNAPSHOT</version>
```
eminn, line 79: ${project.parent.version}

extensions/cdc-postgres/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.hazelcast.jet</groupId>
+            <artifactId>hazelcast-jet-cdc-core</artifactId>
+            <version>4.1-SNAPSHOT</version>
```
eminn, line 79: ${project.parent.version}

extensions/cdc-sqlserver/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.hazelcast.jet</groupId>
+            <artifactId>hazelcast-jet-cdc-core</artifactId>
+            <version>4.1-SNAPSHOT</version>
```
eminn, line 79: ${project.parent.version}

extensions/cdc-mongodb/pom.xml
```diff
@@ -0,0 +1,119 @@
...
+
+    <parent>
+        <groupId>com.hazelcast.jet</groupId>
+        <artifactId>hazelcast-jet-extensions</artifactId>
+        <version>4.1-SNAPSHOT</version>
+    </parent>
+
+    <properties>
+        <debezium.version>1.1.0.Final</debezium.version>
+        <testcontainers.version>1.13.0</testcontainers.version>
```
eminn, line 37: These two property repeats for each module, maybe we can move them to a parent one to avoid repetititon ?

extensions/cdc-mongodb/pom.xml
```diff
@@ -0,0 +1,119 @@
...
+        <dependency>
+            <groupId>io.debezium</groupId>
+            <artifactId>debezium-connector-mongodb</artifactId>
+            <version>${debezium.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <version>1.13.0</version>
```
eminn, line 107: property defined above but not used

extensions/cdc-mysql/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+            <groupId>org.testcontainers</groupId>
+            <artifactId>mysql</artifactId>
+            <version>${testcontainers.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
+            <version>1.7.26</version>
```
eminn, line 108: pull to a property?

extensions/cdc-core/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>connect-api</artifactId>
+            <version>2.3.1</version>
```
eminn, line 78: pull to a property? similar dependencies without version properties found below across multiple cdc modules. Not repeating the comment for each.

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/impl/KafkaConnectSource.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+
+import static com.hazelcast.jet.impl.util.ExceptionUtil.rethrow;
+
+public class KafkaConnectSource {
```
eminn, line 41: this can be package private

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/impl/JsonParsing.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    public static <T> Optional<List<Optional<T>>> getList(JsonNode node, String key, Class<T> clazz) {
+        JsonNode value = node.get(key);
+        return getList(clazz, value);
+    }
+
+    public static Optional<Object> getObject(JsonNode node, String key) {
+        Object value = node.get(key);
+        return getObject(value);
```
eminn, line 71: I think we can inline these `getXXX(JsonNode)` methods

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
```diff
@@ -0,0 +1,182 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.hazelcast.jet.annotation.EvolvingApi;
+
+import javax.annotation.Nonnull;
+import java.io.Serializable;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Arbitrary part of a {@link ChangeEvent}, as big as the whole body orz
```
eminn, line 28: `orz` -> `or`

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEvent.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Identifies the particular record or document being affected
+     * by the change event.
+     */
+    @Nonnull
+    ChangeEventKey key();
+
+    /**
+     * Describes the actual change affected on the record or ducument
```
eminn, line 82: ducument -> document

extensions/cdc-mysql/src/test/java/com/hazelcast/jet/cdc/mysql/MySqlIntegrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ThreadLocalRandom;
+
+import static com.hazelcast.jet.cdc.Operation.DELETE;
+import static com.hazelcast.jet.pipeline.test.AssertionSinks.assertCollectedEventually;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.testcontainers.containers.MySQLContainer.MYSQL_PORT;
+
+public class MySqlIntegrationTest extends AbstractIntegrationTest {
```
eminn, line 53: `MySql` -> `Mysql`

extensions/cdc-mysql/src/main/java/com/hazelcast/jet/cdc/mysql/MySqlCdcSources.java
```diff
@@ -0,0 +1,204 @@
...
+import org.apache.kafka.connect.data.Values;
+
+/**
+ * Contains factory methods for creating change data capture sources
+ * based on MySQL databases.
+ *
+ * @since 4.1
+ */
+@EvolvingApi
+public final class MySqlCdcSources {
```
eminn, line 34: `MySql` -> `Mysql`, even though I'm not sure, JDBC connector uses this way internally.

extensions/cdc-mysql/src/test/java/com/hazelcast/jet/cdc/mysql/GenericDebeziumIntegrationTest.java
```diff
@@ -0,0 +1,136 @@
...
+import java.sql.DriverManager;
+import java.util.concurrent.CompletionException;
+
+import static com.hazelcast.jet.cdc.Operation.DELETE;
+import static com.hazelcast.jet.pipeline.test.AssertionSinks.assertCollectedEventually;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.testcontainers.containers.MySQLContainer.MYSQL_PORT;
+
+public class GenericDebeziumIntegrationTest extends AbstractIntegrationTest {
```
eminn, line 47: I think this should be in the CDC core module along with the base class.

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEventValue.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Describes the change being done by the event. Only used by
+     * MongoDB updates.
+     *
+     * @throws ParsingException if this {@code ChangeEventValue} doesn't
+     *                          have an 'patch' sub-element
+     */
+    @Nonnull
+    ChangeEventElement change() throws ParsingException;
```
eminn, line 92: Do we really want to make this MongoDb specific concern first class citizen of the API ?

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
```diff
@@ -0,0 +1,181 @@
...
+     * the specified key is not found or if it's value is null
+     * @throws ParsingException if the whole structure containing this
+     *                          element is unparsable or the mapping
+     *                          fails to produce a result
+     */
+    @Nonnull
+    <T> T mapToObj(Class<T> clazz) throws ParsingException;
+
+    /**
+     * Best effor method for finding the specified (top level) key in
```
eminn, line 59: effort

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/HazelcastListDatabaseHistory.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.io.IOException;
+import java.util.function.Consumer;
+
+/**
+ * Database history implementation backed by Hazelcast IList {@link IList}.
+ *
+ * @since 4.1
+ */
+public class HazelcastListDatabaseHistory extends AbstractDatabaseHistory {
```
eminn, line 41: I think we can move this to impl package and make it package private

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEventValue.java
jbartok: Assuming I remove it, what will I do when I do want to include Mongo later? Just add the method then, would that be ok?

eminn: No, maybe from scratch mongo source can emit a different event which contains `patch` and `filter` fields in case of an update event.

extensions/cdc-mysql/src/main/java/com/hazelcast/jet/cdc/mysql/MySqlCdcSources.java
jbartok: Seems harder to read to me... `MySQL` would be the perfectly correct version, but that doesn't work with the camel-hump format, so `MySql` seems a good compromise. If you don't have strong feelings/arguments about it, I won't change it.

extensions/cdc-mysql/src/test/java/com/hazelcast/jet/cdc/mysql/MySqlIntegrationTest.java
jbartok: Same as below.

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/HazelcastListDatabaseHistory.java
jbartok: Have moved it, but it has to remain public (I think it's is being used via reflection).

hz-highfive:
https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/2747/build_log.txt), [artifacts](https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/2747/jet-oss-pr-builder-2747.zip)). 
<details>
<pre>

</details>
</pre>

jbartok:
@eminn & @cangencer  pls. re-review when you can. It seems to me the MySQL only version is ready, I would like to sort out any remaining problems, if you see any.

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEvent.java
```diff
@@ -0,0 +1,109 @@
...
+ * <p>
+ * Each event has a <i>key</i>, which identifies the particular
+ * record or document being affected, and a <i>value</i>, which
+ * describes the actual change itself.
+ * <p>
+ * Most events have an <i>operation</i> associated with them which
+ * specifies the type of change being described (insertion, delete or
+ * update). This is really a property of the value, it's only replicated
+ * at this level for convenience. Only some special events, like
+ * heartbeats don't have an operation value.
```
eminn, line 37: Could you please explain what is the `heartbeat` events ?

jbartok: If you set `heartbeat.interval.ms` in the Debezium connectors' config (not just the MySQL one, but in general), they will produce periodic heartbeat events. I haven't exposed this in our sources, but it's possible to enable (via `com.hazelcast.jet.cdc.impl.AbstractSourceBuilder.setCustomProperty`) and encounter them while using CDC via Jet. But is it worth a more than cursory mention in our Javadoc? Don't know. They get some mention, because they are the reason `Operation.UNSPECIFIED` exists... What would you like me to change?

extensions/cdc-core/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <packaging>jar</packaging>
+    <name>hazelcast-jet-cdc-core</name>
+    <description>Change Data Capture support for Hazelcast Jet</description>
+    <url>http://www.hazelcast.com/</url>
+
+    <artifactId>hazelcast-jet-cdc-core</artifactId>
```
cangencer, line 27: should be called just debezium I think

extensions/cdc-core/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>com.hazelcast.jet.cdc-core</Automatic-Module-Name>
```
cangencer, line 58: I don't think hypens are allowed in module names

extensions/cdc-core/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>${jackson.databind.version}</version>
+        </dependency>
+
+        <!-- TEST -->
+        <dependency>
```
cangencer, line 111: we could probably move this to extensions/pom.xml , since all extensions depend on it

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * Arbitrary part of a {@link ChangeEvent}, as big as the whole body or
+ * as small as a single patch expression, based on a complete JSON
+ * expression. Contains various methods for retrieving component values
+ * or for mapping itself to data objects.
+ *
+ * @since 4.1
+ */
+@EvolvingApi
+public interface ChangeEventElement {
```
cangencer, line 35: how is ChangeEventElement different than JsonNode? I don't see anything specific to a ChangeEvent in this interface

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/impl/AbstractSourceBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+import com.hazelcast.function.FunctionEx;
+import com.hazelcast.jet.cdc.ChangeEvent;
+import com.hazelcast.jet.core.Processor;
+import com.hazelcast.jet.pipeline.SourceBuilder;
+import com.hazelcast.jet.pipeline.StreamSource;
+
+import javax.annotation.Nonnull;
+import java.util.Properties;
+
+public abstract class AbstractSourceBuilder<SELF extends AbstractSourceBuilder<SELF>> {
```
cangencer, line 28: why you would need this?

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/impl/HazelcastListDatabaseHistory.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.io.IOException;
+import java.util.function.Consumer;
+
+/**
+ * Database history implementation backed by Hazelcast IList {@link IList}.
+ *
+ * @since 4.1
+ */
+public class HazelcastListDatabaseHistory extends AbstractDatabaseHistory {
```
cangencer, line 41: how do these lists get cleaned up?

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/impl/PropertyRules.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+
+public class PropertyRules {
+
+    private final Set<String> mandatories = new HashSet<>();
```
cangencer, line 27: should be `required`

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Returns the value of the specified (top level) key in the
+     * underlying JSON message as a child {@code ChangeEventElement}.
+     *
+     * @return optional {@code ChangeEventElement}, which is empty if
+     * the specified key is not found or if the value is null.
+     * @throws ParsingException if the underlying JSON message, or any
+     *                          of its parent messages are unparsable
+     */
+    Optional<ChangeEventElement> getChild(String key) throws ParsingException;
```
cangencer, line 64: keep in mind when you return `Optional`, it's creating more garbage

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
```diff
@@ -0,0 +1,189 @@
...
+     * so the parameter class needs to be annotated accordingly.
+     *
+     * @return optional {@code Object} value, which is empty only if the
+     * specified key is not found or if it's value is null
+     * @throws ParsingException if the whole structure containing this
+     *                          element is unparsable or the mapping
+     *                          fails to produce a result
+     */
+    @Nonnull
+    <T> T mapToObj(Class<T> clazz) throws ParsingException;
```
cangencer, line 53: missing annotations (same for other methods)

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/DebeziumCdcSources.java
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Creates a CDC source that streams change data from your Debezium
+     * supported database to the Hazelcast Jet pipeline.
+     *
+     * @param name       name of this source, needs to be unique, will be
+     *                   passed to the underlying Kafka Connect source
+     * @return builder that can be used to set source properties and also
+     * to construct the source once configuration is done
+     */
+    public static Builder debezium(String name, String connectorClass) {
```
cangencer, line 43: missing annotations

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/Operation.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Describes the nature of the event in CDC data. Equivalent to various
+ * actions that can affect a database record: insertion, update and
+ * deletion. Has some extra special values like "sync" which is just
+ * like an insert, but originates from a database snapshot (as opposed
+ * database changelog) and "unspecified" which is used for a few special
+ * CDC events, like heartbeats.
+ *
+ * @since 4.1
```
cangencer, line 32: 4.2

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
jbartok: Made more sense when we didn't yet exclude MongoDB. There it had nothing to do with JsonNode. But even now, do we really want to have `JsonNode` in our API? Even if we do the same thing?

cangencer: I don't see an issue, it's well documented and has more capabilities? I believe we are going to be shading Jackson as part of our public API?

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/impl/AbstractSourceBuilder.java
jbartok: Initially I was expecting a lot of common functionality in the builders of the various sources, but it wasn't so, due to lots of subtle differences. But, the generic method for setting any kind of Debezium property is common and some other internal code. Do you think it would be better not to have it?

extensions/cdc-core/pom.xml
jbartok: Why? Isn't Debezium an implementation detail which we might replace in the future? Or at least aren't we striving towards that goal?

jbartok: Ok, so as discussed will change this to `cdc-debezium`.

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/impl/HazelcastListDatabaseHistory.java
jbartok: No idea, this was code I inherited and never thought of checking. Will do so now.

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
jbartok: I know, but I wanted to keep ParsingException for actual parsing errors... maybe use `NoSuchElementException` instead of some `Optional`s + some `hasXXX` where absolutely necessary? will give it a try

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEvent.java
eminn: I see, I mean we can at least provide a link to it's definition and configuration ? I'm also fine if the explanation above present in Javadoc

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
jbartok: This is not exactly how I ended up, `JsonNode` is gone too now, since moving to Jackson jr, but `ChangeEventElement` is greatly simplified now, we no longer implement parsing shenanigans, everything is in Jackson code. Take a look if you can.

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
jbartok: All `Optional`s are gone now, together with any parsing code of our own.

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/ChangeEvent.java
jbartok: Ok. I don't want to explain too much, because I don't want to encourage anybody to use the heartbeats in Jet, but will add some clarifications.

extensions/cdc-core/src/main/java/com/hazelcast/jet/cdc/impl/HazelcastListDatabaseHistory.java
jbartok: Fixed clean-up in it (has a `stop()` method, which is called by the connector and the list will be destroyed there).

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
```diff
@@ -0,0 +1,75 @@
...
+    @Nonnull
+    Map<String, Object> asMap() throws ParsingException;
+
+    /**
+     * Returns raw JSON string which the content of this event element
+     * is based on. To be used when parsing fails for some reason (for
+     * example on some untested DB-connector version combination).
+     */
+    @Nonnull
+    String asJson();
```
cangencer, line 73: wouldn't it be better to just use `toString()` here?

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
```diff
@@ -0,0 +1,75 @@
...
+package com.hazelcast.jet.cdc;
+
+import com.hazelcast.jet.annotation.EvolvingApi;
+
+import javax.annotation.Nonnull;
+import java.util.Map;
+
+/**
+ * Arbitrary part of a {@link ChangeEvent}, as big as the whole body or
+ * as small as a single patch expression, based on a complete JSON
```
cangencer, line 26: javadoc seems outdated

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/ChangeEvent.java
```diff
@@ -0,0 +1,112 @@
...
+ * events are parts of a database snapshot constructed when we start
+ * monitoring the database and their timestamps are accordingly
+ * artificial. Identifying snapshot events is possible most of the time,
+ * because their operation will be {@link Operation#SYNC} instead of
+ * {@link Operation#INSERT} (one notable exception being MySQL).
+ *
+ * @since 4.2
+ */
+@EvolvingApi
+public interface ChangeEvent {
```
cangencer, line 57: I would prefer `ChangeRecord`

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
```diff
@@ -0,0 +1,75 @@
...
+ * expression. Contains various methods for retrieving component values
+ * or for mapping itself to data objects.
+ *
+ * @since 4.2
+ */
+@EvolvingApi
+public interface ChangeEventElement {
```
cangencer, line 33: this name should be changed to. something like `RecordPart` ?

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/ChangeEventElementJsonImpl.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    @Nonnull
+    public <T> T asPojo(@Nonnull Class<T> clazz) throws ParsingException {
+        Objects.requireNonNull(clazz, "class");
+        try {
+            return J.beanFrom(clazz, json);
+        } catch (IOException e) {
+            throw new ParsingException(e.getMessage(), e);
```
cangencer, line 53: what's the point of rethrowing here? what's wrong with IOException?

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/AbstractSourceBuilder.java
```diff
@@ -0,0 +1,99 @@
...
+import com.hazelcast.jet.cdc.ChangeEvent;
+import com.hazelcast.jet.core.Processor;
+import com.hazelcast.jet.pipeline.SourceBuilder;
+import com.hazelcast.jet.pipeline.StreamSource;
+
+import javax.annotation.Nonnull;
+import java.util.Objects;
+import java.util.Properties;
+
+public abstract class AbstractSourceBuilder<SELF extends AbstractSourceBuilder<SELF>> {
```
cangencer, line 29: this is a private class but extended by a public one

extensions/cdc-debezium/pom.xml
```diff
@@ -0,0 +1,139 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>io.debezium</groupId>
+            <artifactId>debezium-core</artifactId>
+            <version>${debezium.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.jr</groupId>
```
cangencer, line 105: is this going to be shaded in jet-core? can we sync changes with JSON PR?

extensions/cdc-debezium/pom.xml
```diff
@@ -0,0 +1,139 @@
...
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>mysql</artifactId>
+            <version>${testcontainers.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
```
cangencer, line 133: we don't use log4j12 anymore

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/ChangeEvent.java
```diff
@@ -0,0 +1,112 @@
...
+    ChangeEventElement value();
+
+    /**
+     * Returns raw JSON string which the content of this event is
+     * based on. Mean to be used when higher level parsing (see other
+     * methods) fails for some reason (for example on some untested
+     * DB-connector version combination).
+     */
+    @Nonnull
+    String asJson();
```
cangencer, line 111: we can keep this as `toString()`

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/Operation.java
```diff
@@ -0,0 +1,83 @@
...
+     * Parses the string constants used in CDC messages for describing
+     * operations into enum instances.
+     * <p>
+     * Null will be parsed as {@link #UNSPECIFIED}.
+     *
+     * @throws ParsingException if the input string doesn't represent
+     * an expected value.
+     */
+    public static Operation get(@Nullable String id) throws ParsingException {
+        Operation[] values = values();
```
cangencer, line 75: seems now you have to pay cost of comparison twice: once per figuring out the string mapping, and then the other conditional operation you will do for the operation. Why not just publish it as string and refer to constants?

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
jbartok: In my opinion no, because a separate method can provide a clear contract and format of the `String` it returns, while people would not expect that much from `toString()`. As far as I know it's good design not to rely on any particular format of what `toString()` provides. I would not use `toString()` for this, even though we could.

jbartok: I would even prefer to remove the `asJson` method completely to using `toString()`.

mtopolnik: If JSON is the primary representation, it can be OK to use `toString` for it. An overriding method can narrow down the contract of `toString`, there are precedents for this in the JDK. `toString` can also delegate to `asJson`.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/ChangeEventElementJsonImpl.java
jbartok: I use `ParsingException` to signal various failures in parsing, which in my mind is not I/O (failing to receive the JSON string from somewhere, that would be I/O, parsing it after it's in memory, that doesn't seem to be... arguable). 

Anyways, beyond my rumblings, `IOException` is too generic, even Jackson jr have their own `JsonProcessingException`. 

I can change it if you feel strongly about it, I don't mind, just wanted to clarify my reasoning.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
jbartok: Yes, seems like I forgot about a few javadoc, changing.

extensions/cdc-debezium/pom.xml
jbartok: I am planning on using Ali's Json Convenience, once it's merged in and that does chade Jackson jr. I just moved to Jackson jr as an intermediary state, because that will make the final adoption quite simple.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
jbartok: Ok. `toString()` already delegates to `asJson()`. Beyond that I prefer to keep `asJson()`, but am ok with both versions. Pls. decide.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSource.java
```diff
@@ -0,0 +1,201 @@
...
+        Map<String, String> config = new HashMap<>();
+        config.put("add.fields", "op, ts_ms");
+        config.put("delete.handling.mode", "rewrite");
+        smt.configure(config);
+
+        return smt;
+    }
+
+    private static ChangeEvent extractEvent(SourceRecord record) {
+        String keyJson = Values.convertToString(record.keySchema(), record.key());
```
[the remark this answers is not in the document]
jbartok: Yes, it does. But `Struct` is on hand Kafka Connect code and on the other hand terribly hard to work with. So I don't think we will ever want to directly work with it. But if we will at some point, we can change this and other implementation stuff, it's not exposed.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSource.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.Properties;
+
+import static com.hazelcast.jet.impl.util.ExceptionUtil.rethrow;
+
+class CdcSource {
+
+    private final SourceConnector connector;
+    private final SourceTask task;
+    private final Map<String, String> taskConfig;
+    private final ExtractNewRecordState<SourceRecord> smt;
```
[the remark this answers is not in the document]
jbartok: Single Message Transform (https://docs.confluent.io/current/connect/transforms/index.html), but yeah, I see what you mean, no point in being cryptic in the code. Will rename it.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/ChangeEventElement.java
mtopolnik: I'm OK with both methods.

extensions/cdc-debezium/pom.xml
jbartok: Ok, changed to use [Log4j 2 SLF4J Binding](https://logging.apache.org/log4j/2.x/log4j-slf4j-impl/index.html) for the Debezium connectors' logging.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/HazelcastListDatabaseHistory.java
```diff
@@ -0,0 +1,107 @@
...
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public void stop() {
+        super.stop();
+        list.destroy();
```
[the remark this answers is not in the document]
jbartok: Yeah, this was wrong. I've lost test I had for job restart when I've split the PR branch in two. Have put it back now. For now I've removed the `list.destroy()` from the above `stop()` method. It works properly now, but there is again no clean-up for the list. Have to think about that a bit more, figure out what to do.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/Operation.java
jbartok: I've tried a couple of versions, but I couldn't find one that I liked... Considering that the calls in questions will be made only on demand and that the operation values doesn't play any part in serde either, I'm leaning towards keeping it as is, for the moment. Will give it another look tomorrow, with a fresh mind.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/HazelcastListDatabaseHistory.java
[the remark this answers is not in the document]
jbartok: I think I'll have to give these lists a job specific name and clean them up the same way snapshot maps are cleaned up in `JobRepository.cleanup()`. Let's see, will reexamine with a fresh mind.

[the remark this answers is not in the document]
jbartok: I have implemented the above solution (based on the assumption that the source should have the same life-cycle as the Jet job): the lists have a job specific name now and will be cleaned up by the periodic scan implemented in `JobRepository.cleanup()`, once the job is no longer active. Pls. review this specifically (last commit).

hz-highfive:
https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/2782/build_log.txt), [artifacts](https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/2782/jet-oss-pr-builder-2782.zip)). 
<details>
<pre>
--------------------------
---------SUMMARY----------
--------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:checkstyle (default) on project hazelcast-jet-cdc-debezium: An error has occurred in Checkstyle report generation. Failed during checkstyle execution: There are 7 errors reported by Checkstyle 8.29 with /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/checkstyle/checkstyle.xml ruleset. -> [Help 1]
--------------------------
---------ERRORS-----------
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSource.java:50:5: Instance variable definition in wrong order. [DeclarationOrder]
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSource.java:51:5: Instance variable definition in wrong order. [DeclarationOrder]
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSource.java:52:5: Instance variable definition in wrong order. [DeclarationOrder]
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSource.java:53:5: Instance variable definition in wrong order. [DeclarationOrder]
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSource.java:62:5: Instance variable definition in wrong order. [DeclarationOrder]
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSource.java:63:5: Instance variable definition in wrong order. [DeclarationOrder]
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSource.java:65:5: Constructor definition in wrong order. [DeclarationOrder]
--------------------------
</details>
</pre>

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/Operation.java
cangencer: What was about it that you didn't like? I think you'll be creating more garbage by creating the extra Operation objects, without adding much value. When used with the map sink for example you'll need to do conditional operations on it.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcJsonDataSerializerHook.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public DataSerializableFactory createFactory() {
+        return new Factory();
+    }
+
+    private static class Factory implements DataSerializableFactory {
+        @Override
+        public IdentifiedDataSerializable create(int typeId) {
+            switch (typeId) {
+                case ELEMENT:
```
cangencer, line 48: the names here don't seem aligned with the actual types

hazelcast-jet-distribution/pom.xml
```diff
@@ -208,6 +208,18 @@
...
+            <groupId>com.hazelcast.jet</groupId>
+            <artifactId>hazelcast-jet-cdc-debezium</artifactId>
+            <version>${project.version}</version>
+            <classifier>jar-with-dependencies</classifier>
+        </dependency>
+        <dependency>
+            <groupId>com.hazelcast.jet</groupId>
+            <artifactId>hazelcast-jet-cdc-mysql</artifactId>
+            <version>${project.version}</version>
+            <classifier>jar-with-dependencies</classifier>
```
cangencer, line 221: how big is each JAR? would it make sense to require two JARs for MySQL, rather than duplicating the content in each?

jbartok: 12M (debezium) & 18M (mysql), quite similar to the other ones: 14M (grpc), 14M (python), 10M (s3); it's convenient to need a single jar, but it's not that big of a deal; should I remove the duplication or leave it?

cangencer: Let's leave it for now, but we may revisit in the future.

jbartok: Actually... we will have at least 2 more like the mysql one (postgres, sqlserver). Unlikely that we're going to leave them like this. Maybe bite the bullet and do it now?

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/Operation.java
```diff
@@ -0,0 +1,113 @@
...
+     * an expected value.
+     */
+    public static Operation get(@Nullable String string) throws ParsingException {
+        if (string == null) {
+            return UNSPECIFIED;
+        }
+
+        if (string.length() == 1) {
+            char id = string.charAt(0);
+            Operation op = ARRAY[id - OFFSET];
```
cangencer, line 105: I think you don't need offset here, just use an array of size [max] and you can skip the subtraction altogether, simplifying the code.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/HazelcastListDatabaseHistory.java
```diff
@@ -0,0 +1,101 @@
...
+ * Database history implementation backed by Hazelcast IList {@link IList}.
+ *
+ * @since 4.2
+ */
+public class HazelcastListDatabaseHistory extends AbstractDatabaseHistory {
+
+    private volatile String instanceName;
+    private volatile String listName;
+
+    private volatile IList<byte[]> list;
```
cangencer, line 46: does these need to be volatile? probably not?

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/Operation.java
jbartok: Index based lookup in array type of solution is pushed.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/Operation.java
jbartok: Ok

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/Operation.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Parses the string constants used in CDC messages for describing
+     * operations into enum instances.
+     * <p>
+     * Null will be parsed as {@link #UNSPECIFIED}.
+     *
+     * @throws ParsingException if the input string doesn't represent
+     * an expected value.
+     */
+    public static Operation get(@Nullable String string) throws ParsingException {
```
mtopolnik, line 73: Maybe some nicer name instead of `string`? Like a name for the field in CDC terminology.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/Operation.java
```diff
@@ -0,0 +1,120 @@
...
+            ARRAY = new Operation[max - min + 1];
+            for (Operation op : values()) {
+                if (op.id == null) {
+                    continue;
+                }
+                ARRAY[op.id - OFFSET] = op;
+            }
+        }
+
+        static Operation get(char id) {
```
mtopolnik, line 116: This can throw `ArrayIndexOutOfBoundsException`. We should have a unit test that catches these problems.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/HazelcastListDatabaseHistory.java
jbartok: Is read and written by different threads and not safely published. But anyways, let's see if we can move this whole thing into the snapshot.

hz-highfive:
https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/2788/build_log.txt), [artifacts](https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/2788/jet-oss-pr-builder-2788.zip)). 
<details>
<pre>
--------------------------
---------SUMMARY----------
--------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:checkstyle (default) on project hazelcast-jet-cdc-debezium: An error has occurred in Checkstyle report generation. Failed during checkstyle execution: There is 1 error reported by Checkstyle 8.29 with /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/checkstyle/checkstyle.xml ruleset. -> [Help 1]
--------------------------
---------ERRORS-----------
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/extensions/cdc-debezium/src/test/java/com/hazelcast/jet/cdc/OperationTest.java:1: Line does not match expected header line of '/*'. [Header]
--------------------------
</details>
</pre>

hz-highfive:
https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/2789/build_log.txt), [artifacts](https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/2789/jet-oss-pr-builder-2789.zip)). 
<details>
<pre>
--------------------------
---------SUMMARY----------
--------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:checkstyle (default) on project hazelcast-jet-cdc-debezium: An error has occurred in Checkstyle report generation. Failed during checkstyle execution: There is 1 error reported by Checkstyle 8.29 with /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/checkstyle/checkstyle.xml ruleset. -> [Help 1]
--------------------------
---------ERRORS-----------
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/extensions/cdc-debezium/src/test/java/com/hazelcast/jet/cdc/OperationTest.java:1: Line does not match expected header line of '/*'. [Header]
--------------------------
</details>
</pre>

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/HazelcastListDatabaseHistory.java
jbartok: Have made the schema history part of the snapshot. There are two relatively shady parts in the change, I would be thankful if you would take a look at those:

1. see how I use `DatabaseHistoryImpl.container().setHistory(...)` to pass the right record list to `DatabaseHistoryImpl`; `DatabaseHistoryImpl` gets initilized by the surrounded `SourceTask.start(...)` and this happens on a blocking worker thread

2. in `CdcSource.State` I've used a `CopyOnWriteArrayList` for the schema history; the reasoning here was that items get added to this list on some internal debezium snapshot thread, while restore and our snapshotting reads it on a blocking worker thread, so I need to ensure proper visibility + the list is written only on schema changes, which should be a rare event after the initial db snapshot so the garbage created should be a one-time hit; I think it's correct, but doesn't feel right so I'm open to suggestions...

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSource.java
```diff
@@ -0,0 +1,248 @@
...
+        @Override
+        public int getClassId() {
+            return CdcJsonDataSerializerHook.SOURCE_STATE;
+        }
+
+        @Override
+        public void writeData(ObjectDataOutput out) throws IOException {
+            out.writeObject(partitionsToOffset);
+
+            List<byte[]> records = historyRecords;
```
cangencer, line 229: you can use writeObject to write a List.

jbartok: For some reason I though this would be more space-efficient, but I should have checked. Will revisit.

hazelcast-jet-distribution/pom.xml
jbartok: Done, mysql jar is now 6.8K (and works).

jbartok:
@cangencer pls. release this for merging in and then I'll openn separate PRs for Json convenience integration, map sink and such.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/CdcSinks.java
```diff
@@ -0,0 +1,131 @@
...
+     * Returns a sink equivalent to {@link #map}, but for a map in a
+     * remote Hazelcast cluster identified by the supplied {@code
+     * ClientConfig}.
+     * <p>
+     * Due to the used API, the remote cluster must be at least 3.11.
+     *
+     * @since 4.2
+     */
+    @Nonnull
+    public static <K, V> Sink<ChangeRecord> removeMap(
```
cangencer, line 101: typo, `remote`

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/CdcSinks.java
```diff
@@ -0,0 +1,131 @@
...
+     * <p>
+     * For the functionality of this sink it is vital that the order of
+     * the input items is preserved so the sink is non-distributed and
+     * its local parallelism is forced to 1. This way only a single
+     * instance will be created for each pipeline.
+     *
+     * @since 4.2
+     */
+    @Nonnull
+    public static <K, V> Sink<ChangeRecord> map(
```
cangencer, line 82: should also take a variant with `IMap<K,V>`

site/docs/api/sources-sinks.md
```diff
@@ -949,6 +949,74 @@ tests](https://github.com/hazelcast/hazelcast-jet-contrib/tree/master/xa-test)
 to get more information. This only applies to the JDBC sink, the source
 doesn't use XA transactions.
 
+### CDC
```
cangencer, line 952: should say Change Data Capture (CDC)

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcJsonDataSerializerHook.java
```diff
@@ -0,0 +1,60 @@
...
+
+import com.hazelcast.internal.serialization.DataSerializerHook;
+import com.hazelcast.internal.serialization.impl.FactoryIdHelper;
+import com.hazelcast.nio.serialization.DataSerializableFactory;
+import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
+
+import static com.hazelcast.jet.impl.JetFactoryIdHelper.JET_CDC_JSON_DS_FACTORY;
+import static com.hazelcast.jet.impl.JetFactoryIdHelper.JET_CDC_JSON_DS_FACTORY_ID;
+
+public class CdcJsonDataSerializerHook implements DataSerializerHook {
```
cangencer, line 27: when do we use `IdentifiedDataSerializable` and when just `SerializerHook`? seems for data types we use `SerializerHook`?

site/docs/api/sources-sinks.md
```diff
@@ -949,6 +949,74 @@ tests](https://github.com/hazelcast/hazelcast-jet-contrib/tree/master/xa-test)
...
+Pipeline pipeline = Pipeline.create();
+pipeline.readFrom(
+    MySqlCdcSources.mysql("customers")
+            .setDatabaseAddress("127.0.0.1")
+            .setDatabasePort(3306)
+            .setDatabaseUser("debezium")
+            .setDatabasePassword("dbz")
+            .setClusterName("dbserver1")
+            .setDatabaseWhitelist("inventory")
+            .setTableWhitelist("inventory." + "customers")
```
eminn, line 982: is this intentional ?

site/docs/tutorials/cdc.md
```diff
@@ -325,57 +320,31 @@ package org.example;
...
-                .with("include.schema.changes", "false")
-                .with("database.history.hazelcast.list.name", "test")
+        StreamSource<ChangeRecord> source = MySqlCdcSources.mysql("source")
+                .setDatabaseAddress("127.0.0.1")
+                .setDatabasePort(3306)
+                .setDatabaseUser("debezium")
+                .setDatabasePassword("dbz")
+                .setClusterName("dbserver1")
+                .setDatabaseWhitelist("inventory")
+                .setTableWhitelist("inventory." + "customers")
```
eminn, line 340: same here

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcJsonDataSerializerHook.java
jbartok: I wasn't really understanding what I was doing... will cover/fix in separate PR.

PR #2143, opened 2020-04-06
Turn off linkXRef option to remove build warning

frant-hartm:
This option is useful when a checkstyle report is generated,
it adds links to the source code directly from the checkstyle
html report.

We fail the build so the only report we ever get is empty report,
making this option not useful.

Checklist
- [x] Tags Set
- [x] Milestone Set


frant-hartm:
verify

PR #2161, opened 2020-04-09
Add checkNonNullAndSerializable util method

frant-hartm:
This combines Objects.requireNonNull with our Util.checkSerializable
into single method call. Useful in source/sink builders for checking
parameters.

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
@@ -32,7 +32,6 @@ import com.hazelcast.spi.impl.NodeEngine;
 
 import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 import java.io.IOException;
 import java.io.InvalidClassException;
 import java.io.NotSerializableException;
@@ -167,8 +166,7 @@ public final class Util {
      * @return given object
      * @throws IllegalArgumentException if {@code object} is not serializable
      */
-    @Nullable
-    public static <T> T checkSerializable(@Nullable T object, @Nonnull String objectName) {
+    public static <T> T checkSerializable(T object, String objectName) {
         if (object == null) {
             return null;
         }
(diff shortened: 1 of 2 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -182,6 +182,32 @@ public static void checkSerializable(Object object, String objectName) {
         }
     }
 
+    /**
+     * Checks that the {@code object} is not null and  implements
```
viliam-durina, line 186: Double space

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -182,6 +182,32 @@ public static void checkSerializable(Object object, String objectName) {
...
+     *     this.value = checkNonNullAndSerializable(value, "value");
+     * }
+     * }</pre>
+     *
+     * @param object     object to check
+     * @param objectName object description for the exception
+     * @return given object
+     * @throws IllegalArgumentException if {@code object} is not serializable
+     */
+    public static <T> T checkNonNullAndSerializable(T object, String objectName) {
```
viliam-durina, line 203: We can add `@Nonnull` annotation for the method and `@Nullable` for `object` parameter.

viliam-durina, line 203: `requireNonNullAndSerializable` would better match the standard `Objects.requireNonNull`. We can also use it inside the method...

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/Util.java
```diff
@@ -180,6 +181,33 @@ public static void checkSerializable(Object object, String objectName) {
...
```
cangencer, line 205: can we add Nullable / Nonnull annotations to these methods?

PR #2163, opened 2020-04-14
[007] Add gRPC Design Doc

cangencer:


site/docs/design-docs/007-grpc-support.md
```diff
@@ -0,0 +1,191 @@
...
+title: 007 - Extended gRPC Support
+description: Improved support for working with gRPC services
+---
+
+*Target Release*: 4.1
+
+## Goal
+
+Provide a convenient way to call gRPC services with good performance
+as part of a pipeline.
```
frant-hartm, line 11: Did we actually achieve this goal? Do we have some proof of _good performance_ (apart from the lessons learned from the Python integration)?

cangencer: I didn't get to have any time to do full benchmarks. @mtopolnik has some preliminary numbers. But it needs to be tested on a real cluster, with a non co-located service.

cangencer: I did a quick local bench and bidirectional is about 2x more throughput than unary.

Issue #2162, opened 2020-04-14
com.hazelcast.jet.grpc.GrpcServiceTest failed due to OutOfMemoryError


olukas:
_master_ (commit edb1e3c05c628881f9509a5277ed786e70b948c2)

Failed on Sonar Nightly build: http://jenkins.hazelcast.com/job/jet-oss-master-sonar-nightly/458/testReport/junit/com.hazelcast.jet.grpc/GrpcServiceTest/com_hazelcast_jet_grpc_GrpcServiceTest/

Stacktrace:
```
org.junit.runners.model.TestTimedOutException: test timed out after 900 seconds
	at sun.management.ThreadImpl.getThreadUserTime(ThreadImpl.java:293)
	at sun.management.ThreadImpl.getThreadUserTime(ThreadImpl.java:285)
	at com.hazelcast.test.AbstractHazelcastClassRunner.generateThreadDump(AbstractHazelcastClassRunner.java:351)
	at com.hazelcast.test.AbstractHazelcastClassRunner.access$000(AbstractHazelcastClassRunner.java:64)
	at com.hazelcast.test.AbstractHazelcastClassRunner$ThreadDumpAwareRunAfters.evaluate(AbstractHazelcastClassRunner.java:394)
	at com.hazelcast.test.DumpBuildInfoOnFailureRule$1.evaluate(DumpBuildInfoOnFailureRule.java:37)
	at com.hazelcast.test.jitter.JitterRule$1.evaluate(JitterRule.java:104)
	at com.hazelcast.test.metrics.MetricsRule$1.evaluate(MetricsRule.java:62)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at com.hazelcast.test.HazelcastSerialClassRunner.runChild(HazelcastSerialClassRunner.java:50)
	at com.hazelcast.test.HazelcastSerialClassRunner.runChild(HazelcastSerialClassRunner.java:29)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
[...]

PR #2167, opened 2020-04-14
Shutdown gRPC channels cleanly

frant-hartm:
Several fixes
- closing channels doesn't happen imediatelly - added awatTermination
- Server stored in local variable not shutdown at end of test
- onCompleted should call onCompleted on responseObserver

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/extensions/grpc/src/main/java/com/hazelcast/jet/grpc/GrpcServices.java
+++ b/extensions/grpc/src/main/java/com/hazelcast/jet/grpc/GrpcServices.java
@@ -25,14 +25,14 @@ import com.hazelcast.jet.grpc.impl.GrpcUtil;
 import com.hazelcast.jet.grpc.impl.UnaryService;
 import com.hazelcast.jet.pipeline.GeneralStage;
 import com.hazelcast.jet.pipeline.ServiceFactory;
-import com.hazelcast.logging.ILogger;
-import com.hazelcast.logging.Logger;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import io.grpc.stub.StreamObserver;
 
 import javax.annotation.Nonnull;
 import java.util.concurrent.CompletableFuture;
+
+import static com.hazelcast.jet.datamodel.Tuple2.tuple2;
 
 /**
  * Provides {@link ServiceFactory} implementations for calling gRPC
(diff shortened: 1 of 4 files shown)


cangencer:
Does this fix #2162?

extensions/grpc/src/main/java/com/hazelcast/jet/grpc/GrpcServices.java
```diff
@@ -107,7 +108,7 @@ private GrpcServices() {
                 .withCreateContextFn(ctx -> channelFn.get().build())
                 .withCreateServiceFn((ctx, channel) -> new UnaryService<>(channel, callStubFn))
                 .withDestroyServiceFn(UnaryService::destroy)
-                .withDestroyContextFn(ManagedChannel::shutdown);
+                .withDestroyContextFn(channel -> channel.shutdown().awaitTermination(5, TimeUnit.SECONDS));
```
cangencer, line 111: what happens if times out 5 seconds?

frant-hartm: Ha! I thought it throws TimeoutException, but it just returns true/false if it managed to shutdown in time.

What should we do? Throw the exception ourselves?

cangencer: is there some way to terminate it not so gracefully?

extensions/grpc/src/main/java/com/hazelcast/jet/grpc/GrpcServices.java
```diff
@@ -107,7 +112,7 @@ private GrpcServices() {
                 .withCreateContextFn(ctx -> channelFn.get().build())
                 .withCreateServiceFn((ctx, channel) -> new UnaryService<>(channel, callStubFn))
                 .withDestroyServiceFn(UnaryService::destroy)
-                .withDestroyContextFn(ManagedChannel::shutdown);
+                .withDestroyContextFn(channel -> GrpcUtil.shutdownChannel(channel, logger));
```
cangencer, line 115: it would be better to use the `logger` from ctx, you can do this by making createContextFn return `tuple2(channel, logger)` , kind of ugly, but it should be fine.

Issue #2059, opened 2020-03-12
Client shouldn't try to destroy Observable on disconnection


mtopolnik:
Since the Jet job will itself create and re-create an Observable on access, there is no point in trying to destroy it when the client disconnects.

PR #2170, opened 2020-04-15
Remove destroying Observables on JetInstance shutdown

jbartok:
Triggering `Observable.destroy()` on `JetInstance` shutdown has some unexpected consequences (for example: you kill a client, the client destroys an observable, job execution terminates a bit later and it re-creates the observable in the meantime, next execution starts with a few left-over results). This is also related to the fact that observables aren't job specific and that we allow them to be reused in multiple executions... But anyways, better to not destroy them in such unexpected ways (especially since it's not even documented). Leave all destroys be explicitly triggered by users.

Fixes [#2059 ](https://github.com/hazelcast/hazelcast-jet/issues/2059)

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJetInstance.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJetInstance.java
@@ -170,7 +170,6 @@ public abstract class AbstractJetInstance implements JetInstance {
 
     @Override
     public void shutdown() {
-        observables.values().forEach(Observable::destroy);
         hazelcastInstance.shutdown();
     }
 
--- a/hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/ObservableShutdownTest.java
+++ b/hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/ObservableShutdownTest.java
@@ -17,7 +17,6 @@
 package com.hazelcast.jet.core;
 
 import com.hazelcast.jet.JetInstance;
-import com.hazelcast.jet.JetTestInstanceFactory;
 import com.hazelcast.jet.Job;
 import com.hazelcast.jet.Observable;
 import com.hazelcast.jet.function.Observer;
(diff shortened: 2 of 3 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/AbstractJetInstance.java
```diff
@@ -170,7 +170,6 @@ public JetCacheManager getCacheManager() {
 
     @Override
     public void shutdown() {
-        observables.values().forEach(Observable::destroy);
```
cangencer, line 173: this does some other logic like removing listeners, this will be done automatically when client shutdowns I assume?

jbartok: No, listeners aren't killed on `JetInstance.shutdown()` now either, but indeed, they should be. Then the behaviour would be more consistent with what `getObservables()` is supposed to achieve too. I'll revisit this whole thing again and take care of the listeners.

cangencer: I think Listeners are killed automatically from server side when client disconnects

jbartok: You were right, listeners (even though they are local to the client) get killed on `JetInstance` shutdown (regardless if it's a client or a member), because they get a client/instance inactive exception and cancel themselves because of it. 

Have added a test for this. 

Initial change is enough, we don't need to do more.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/ObservableShutdownTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+    private Observable<Long> memberObservable;
+    private Observable<Long> clientObservable;
+
+    private TestObserver memberObserver;
+    private TestObserver clientObserver;
+
+    @Before
+    public void before() {
+        factory = new JetTestInstanceFactory();
```
cangencer, line 63: why not use `JetTestSupport` ?

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/ObservableShutdownTest.java
```diff
@@ -0,0 +1,128 @@
...
+        memberObservable = members[members.length - 1].newObservable();
+        memberObservable.addObserver(memberObserver);
+
+        clientObserver = new TestObserver();
+        clientObservable = client.newObservable();
+        clientObservable.addObserver(clientObserver);
+    }
+
+    @Test
+    public void cleanup() {
```
cangencer, line 67: can we have a more descriptive test name.. `when_then` etc

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/ObservableShutdownTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class ObservableShutdownTest extends JetTestSupport {
```
cangencer, line 39: you can mark this to be used with ParallelRunner (although there is only one test currently)

PR #2173, opened 2020-04-15
Add javac param to stop package info files sabotaging incremental compilation

jbartok:
Maven compile plugin has a [bug](https://issues.apache.org/jira/browse/MCOMPILER-368) dues to which our `package-info.java` files are always considered dirty and most modules are always recompiled, even if there aren't actual changes in them. 

Will try a workaround for this by setting the `javac` param `pkginfo`.

Checklist
- [x] Tags Set
- [x] Milestone Set


jbartok:
verify

PR #2179, opened 2020-04-16
Added script for version replacement

eminn:
We can start putting placeholders for Jet and IMDG versions in the `.md` files and before creating new version they can get replaced with the provided script.

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/site/website/replace-version-config.js
+++ b/site/website/replace-version-config.js
@@ -1,0 +1,17 @@
+const replace = require('replace-in-file');
+const options = {
+  files: "../docs/**/**.md",
+  from: [/{jet-version}/g, /{imdg-version}/g],
+  to: ['4.0', '4.0'],
+  isRegex: true,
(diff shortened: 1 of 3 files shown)


site/website/replace-version-config.js
```diff
@@ -0,0 +1,17 @@
+const replace = require('replace-in-file');
+const options = {
+  files: "../docs/**/**.md",
+  from: [/{jet-version}/g, /{imdg-version}/g],
```
viliam-durina, line 4: isn't `{` a reserved character in regex? This matches three a's:
```regex
a{3}
```

cangencer:
verify

site/website/replace-version-config.js
eminn: I know that is but in this format without escaping it works as expected.

PR #2185, opened 2020-04-16
UnaryService for gRPC doesn't need `volatile` on StreamObserver

cangencer:
see also: https://github.com/grpc/grpc-java/issues/6929

Add description here

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/extensions/grpc/src/main/java/com/hazelcast/jet/grpc/impl/UnaryService.java
+++ b/extensions/grpc/src/main/java/com/hazelcast/jet/grpc/impl/UnaryService.java
@@ -49,7 +49,7 @@ public final class UnaryService<T, R> implements GrpcService<T, R> {
     private static class Observer<R> implements StreamObserver<R> {
         private final CompletableFuture<R> future;
 
-        private volatile R value ;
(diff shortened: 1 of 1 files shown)


extensions/grpc/src/main/java/com/hazelcast/jet/grpc/impl/UnaryService.java
```diff
@@ -49,14 +49,15 @@ public void destroy() {
...
-        private volatile R value ;
+        private R value ;
 
         Observer() {
             this.future = new CompletableFuture<>();
         }
 
         @Override
         public void onNext(R value) {
+            assert this.value == null : "value should not be assigned twice in unary mode";
```
mtopolnik, line 60: This would be user error, so let's raise this to a hard check.

cangencer: grpc doesn't allow this, and throws exception. The only case would be if this observer was used for a non-unary service, whcih I think the signature wouldn't allow.

PR #2194, opened 2020-04-17
Introduce version replacement tags

jbartok:
Applies the version replacement tags created in #2179.

Those tags get replaced before deploying a new version. The process can be run manually by issuing the `npm run replace-version` command in the `site/website` folder. It will run and replace the tags with actual versions in all the files. 

This makes checking the changes automatically possible, because after you run the above command all the content should be identical with what's on the `master` branch. And it is (well, almost, but few differences, simple to check)!

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/site/docs/api/data-structures.md
+++ b/site/docs/api/data-structures.md
@@ -244,7 +244,7 @@ coordination tasks.
 ## ICache
 
 `ICache` is a fully compliant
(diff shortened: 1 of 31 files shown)


jbartok:
**_DRAFT_** pull request, I still intend to go over it one more time and check, do not review and conclude that I'm a moron! (At least not yet...)

PR #2196, opened 2020-04-17
Add Demos Page to the Website

ufukyilmaz:
Add a new page including the demo applications to the website.

### DIFF
--- a/site/website/package.json
+++ b/site/website/package.json
@@ -31,6 +31,8 @@
     "wait-on": "^4.0.0"
   },
   "dependencies": {
-    "concurrently": "^5.1.0"
+    "concurrently": "^5.1.0",
+    "@material-ui/core": "^4.9.10",
+    "react": "^16.13.1"
   }
 }
--- a/site/website/package.json
+++ b/site/website/package.json
@@ -31,8 +31,6 @@
     "wait-on": "^4.0.0"
   },
   "dependencies": {
-    "concurrently": "^5.1.0",
-    "@material-ui/core": "^4.9.10",
-    "react": "^16.13.1"
+    "concurrently": "^5.1.0"
   }
 }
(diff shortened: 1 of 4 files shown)


site/website/package.json
```diff
@@ -31,6 +31,8 @@
     "wait-on": "^4.0.0"
   },
   "dependencies": {
-    "concurrently": "^5.1.0"
+    "concurrently": "^5.1.0",
+    "@material-ui/core": "^4.9.10",
```
eminn, line 35: Do we really need Material UI just for Card component ?

ufukyilmaz: I removed the unnecessary dependencies.

site/website/pages/en/demos.js
```diff
@@ -0,0 +1,187 @@
+const React = require('react');
+const CompLibrary = require('../../core/CompLibrary');
+const Container = CompLibrary.Container;
+const MarkdownBlock = CompLibrary.MarkdownBlock;
+
+class Demos extends React.Component {
+
+    render() {
+        const {config: siteConfig, language = ''} = this.props;
+        const cardStyle = {
```
cangencer, line 10: would it better to move these to the css?

ufukyilmaz: I moved these styles to a jss file.

site/website/pages/en/demos.js
```diff
@@ -2,60 +2,20 @@ const React = require('react');
 const CompLibrary = require('../../core/CompLibrary');
 const Container = CompLibrary.Container;
 const MarkdownBlock = CompLibrary.MarkdownBlock;
+const cardStyle = require(process.cwd() + "/static/jss/card.js");
```
cangencer, line 5: shouldn't you use relative path here?

ufukyilmaz: When I tried to use a relative path, I got an error for every path I tried. I think the docusaurus does not support it. When I looked at how to import this jss component, I came across [this](https://docusaurus.io/docs/en/api-pages#page-require-paths) and applied it.

cangencer:
@eminn could you review?

PR #2208, opened 2020-04-22
Grpc how to guide

frant-hartm:
Updated the grpc example to use both unary and bidirectional service.
Added how to guide.


site/docs/how-tos/grpc.md
```diff
@@ -0,0 +1,183 @@
...
+gRPC service. Currently, two types of gRPC services are supported:
+
+- unary service
+- bidirectional streaming service.
+
+This how to guide shows how to use these services in your pipeline.
+
+## Module dependency
+
+Add a dependency on `hazelcast-jet-grpc` module to use the gRPC services
```
cangencer, line 22: add dependency where? where do you get the module and put it on the classpath?

site/docs/how-tos/grpc.md
```diff
@@ -0,0 +1,183 @@
...
+
+```xml
+<dependency>
+    <groupId>com.hazelcast.jet</groupId>
+    <artifactId>hazelcast-jet-grpc</artifactId>
+    <version>4.1</version>
```
cangencer, line 28: should use the placeholder version tag

site/docs/how-tos/grpc.md
```diff
@@ -0,0 +1,183 @@
...
+## Unary service
+
+A unary service allows to send a single request and receive a response,
+similar to a normal function call.
+
+For example, given the protobuf definition below
+
+```proto
+service ProductService {
+  rpc ProductInfo (ProductInfoRequest) returns (ProductInfoReply) {}
```
cangencer, line 41: you also need definition of ProductInfoRequest and ProductionInfoReply

site/docs/how-tos/grpc.md
```diff
@@ -0,0 +1,183 @@
...
+```java
+StreamStage<Trade> trades = ...
+trades.mapUsingServiceAsync(productService,
+(service, trade) -> {
+    ProductInfoRequest request = ProductInfoRequest.newBuilder().setId(trade.productId()).build();
+    return service.call(request).thenApply(productReply -> tuple2(trade, productReply.getProductName()));
+})
+```
+
+## Bidirectional streaming service
```
cangencer, line 84: for bidirectional we may show the server-side implementation as well, because it's not immediately obvious perhaps? just as a snippet, we probably don't need the full code.

cangencer:
could you add labels and tags as well?

site/docs/how-tos/grpc.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+To call this service, use `GrpcServices.unaryService()`:
+
+```java
+ServiceFactory<?, ? extends GrpcService<ProductInfoRequest, ProductInfoReply>>
+productService = unaryService(
+    () -> ManagedChannelBuilder.forAddress("localhost", PORT)
+                               .useTransportSecurity()
+                               .usePlaintext(),
```
cangencer, line 60: doesn't these two contradict each other? plaintext and transportSecurity?

site/docs/how-tos/grpc.md
cangencer: this is still unresolved. I believe we also show dependencies as gradle, so let's follow the standard there.

site/docs/how-tos/grpc.md
frant-hartm: I think the user either already knows gRPC and just wants to use it from Jet, or doesn't and in that case would need to follow some other tutorial to create the server. A snippet from e.g. here wouldn't be enough

https://github.com/hazelcast/hazelcast-jet/blob/dbbe962fc14043f39df80eb72371643b7adde5d6/examples/grpc/src/main/java/com/hazelcast/jet/examples/grpc/BrokerServiceImpl.java#L38

frant-hartm:
@mtopolnik I moved your other doc fixes to separate PR #2212, let's keep the history nice and focused.

PR #2212, opened 2020-04-23
Clean up code sample and other docs

frant-hartm:
Doc updates originally part of #2208, but unrelated to grpc how to guide.

### DIFF
--- a/site/docs/api/sources-sinks.md
+++ b/site/docs/api/sources-sinks.md
@@ -1,5 +1,5 @@
 ---
-title: Sources and Sinks 
+title: Sources and Sinks
 description: Birds-eye view of all pre-defined sources available in Jet.
 ---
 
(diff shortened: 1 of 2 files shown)


site/docs/how-tos/custom-batch-source.md
```diff
@@ -1,28 +1,25 @@
 ---
-title: Create a Batch Source
+title: Add Batching to Custom Source
```
cangencer, line 2: title seems wrong, it's about creating a batch source, not adding batching to a source

site/docs/api/sources-sinks.md
```diff
@@ -1097,7 +1098,8 @@ Pipeline p = Pipeline.create();
 p.readFrom(TestSources.items("a", "b", "c", "d"))
  .writeTo(Sinks.observable(observable));
 
-Future<List<String>> future = observable.toFuture(s -> s.collect(Collectors.toList()));
+Future<List<String>> future = observable.toFuture(
+    s -> s.collect(Collectors.toList()));
```
cangencer, line 1102: can we align the `);` on the next line? it's easier to read like that.

cangencer:
Noting also that you can use `docs-no-build` label for docs only PRs to skip the full PR builder

site/docs/how-tos/custom-batch-source.md
mtopolnik: Actually the only new point this How-To introduces is the batching. Originally there was a tutorial, then it got split into the "basics" which are in _Sources and Sinks_ and this how-to, which explains how to add batching.

cangencer: ok, but it's confusing if user is trying to look for how to create a custom source, not just add batching to it, where 's the entry point?

mtopolnik: In the section titled "Sources and Sinks", subsection "Custom Sources and Sinks".

PR #2221, opened 2020-04-28
Fix jet version in docs

gierlachg:
Fixed jet version in docs - replaced {hazelcast.jet.version} with {jet-version.}

Checklist
- [x] Tags Set
- [x] Milestone Set


cangencer:
can we fix in 4.0 as well?

gierlachg:
That part was added with 4.1

PR #2236, opened 2020-04-30
Various minor fixes

cangencer:


site/website/javadoc.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+find ../javadoc -name "*.tar.gz" -exec tar xzvf {} -Cstatic/javadoc \;
```
olukas, line 3: what about to add `mkdir -p static/javadoc` before this line. Currently tt will fail if `static/javadoc` dir does not exist.

olukas: hm, actually `mkdir -p` will rewrite it. We should rather check whether dir is existed and create it only if not.

PR #2242, opened 2020-05-07
Add equivalent assert methods for streaming type sinks

jbartok:
`AssertionSinks` has some convenience methods that would be useful in tests with streaming jobs too. Adding them.

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/test/AssertionSinks.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/test/AssertionSinks.java
@@ -48,13 +48,13 @@ public final class AssertionSinks {
     }
 
     /**
-     * Asserts that the previous stage emitted the exact sequence of expected
-     * items and nothing else. If the assertion fails, the job will fail with an
-     * {@link AssertionError} with the given message.
+     * Asserts that the previous stage emitted the exact sequence of
+     * expected items and nothing else. If the assertion fails, the job
+     * will fail with an {@link AssertionError} with the given message.
      * <p>
-     * Since Jet jobs are distributed, input from multiple upstream processors
-     * is merged in a non-deterministic way. Therefore this assertion is usable
-     * only for testing of non-distributed sources.
+     * Since Jet jobs are distributed, input from multiple upstream
+     * processors is merged in a non-deterministic way. Therefore this
+     * assertion is usable only for testing of non-distributed sources.
      * <p>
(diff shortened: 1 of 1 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/test/AssertionSinks.java
```diff
@@ -138,6 +146,46 @@ private AssertionSinks() {
...
+            @Nonnull Collection<? extends T> expected
+    ) {
+        final List<? super T> exp = new ArrayList<>(expected);
+        return assertCollectedEventually(timeoutSeconds,
+                received -> assertEquals(null, exp, received));
+    }
+
+    /**
+     * Asserts that the previous stage emitted the expected items in any order,
+     * but nothing else.
```
viliam-durina, line 169: How do you check the "nothing else" part?

cangencer: The javadoc is wrong, since you don't. It completes as long as those items are received but there can be subsequent items.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/test/AssertionSinks.java
```diff
@@ -138,6 +146,46 @@ private AssertionSinks() {
             .build();
     }
 
+    /**
+     * Asserts that the previous stage emitted the exact sequence of
+     * expected items and nothing else.
+     * <p>
+     * Is just a wrapper around {@link #assertCollectedEventually(int, ConsumerEx)},
```
cangencer, line 153: javadoc needs to be clear and doens't need to list the implementation details. see the other methods' javadoc.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/test/AssertionSinks.java
```diff
@@ -106,9 +114,9 @@ private AssertionSinks() {
     }
 
     /**
-     * Asserts that the previous stage emitted all of the given items in any order.
-     * If the assertion fails, the job will fail with a {@link AssertionError} with
-     * the given message.
+     * Asserts that the previous stage emitted all of the given items in
+     * any order. If the assertion fails, the job will fail with a
+     * {@link AssertionError} with the given message.
```
viliam-durina, line 119: This one is also not usable in streaming jobs - the upstream must complete.

PR #2245, opened 2020-05-11
Improve gRPC module

frant-hartm:
Changes:
- create channel per processor
- make service destroy timeout and channel shutdown timeout configurable
- Use common ForkJoinPool as executor for the channel

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/extensions/grpc/src/main/java/com/hazelcast/jet/grpc/impl/GrpcProperties.java
+++ b/extensions/grpc/src/main/java/com/hazelcast/jet/grpc/impl/GrpcProperties.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2020 Hazelcast Inc.
+ *
+ * Licensed under the Hazelcast Community License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://hazelcast.com/hazelcast-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.jet.grpc.impl;
+
+import com.hazelcast.spi.properties.HazelcastProperty;
+import io.grpc.stub.StreamObserver;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * Properties of the grpc module
+ */
+public final class GrpcProperties {
+
+    /**
+     * Time to wait for {@link StreamObserver#onCompleted()} confirmation when destroying a
+     * {@link BidirectionalStreamingService}
+     * You may want to increase this value when your gRPC calls take longer than 1 s to cleanly shutdown the service.
+     * <p>
+     * The default value is 1 s
+     *
+     * @since 4.2
+     */
+    public static final HazelcastProperty DESTROY_TIMEOUT
+            = new HazelcastProperty("jet.grpc.destroy.timeout", 1, SECONDS);
+
+    /**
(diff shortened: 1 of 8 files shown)


site/docs/how-tos/grpc.md
```diff
@@ -246,6 +246,15 @@ stage.mapUsingServiceAsyncBatched(bidiService,
 })
 ```
 
+If your batch takes close to or more than 1 s (including the network
+overhead) you should increase the value for following properties to
+perform clean shutdown:
+
+```
+jet.grpc.destroy.timeout
```
cangencer, line 254: maybe we should point the user to where you can learn more about these proeprties?

extensions/grpc/src/main/java/com/hazelcast/jet/grpc/impl/GrpcProperties.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.hazelcast.spi.properties.HazelcastProperty;
+import io.grpc.stub.StreamObserver;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * Properties of the grpc module
+ */
+public final class GrpcProperties {
```
cangencer, line 27: missing `@since`, you can include here and drop the rest on the same file.

cangencer:
@mtopolnik @frant-hartm please have a look at this again

extensions/grpc/src/main/java/com/hazelcast/jet/grpc/impl/GrpcUtil.java
```diff
@@ -52,8 +52,8 @@ public static Throwable translateGrpcException(Throwable exception) {
      * Tries orderly shutdown first {@link ManagedChannel#shutdown()}, then forceful shutdown
      * {@link ManagedChannel#shutdownNow()}.
      */
-    public static void shutdownChannel(ManagedChannel channel, ILogger logger) throws InterruptedException {
-        if (!channel.shutdown().awaitTermination(1, SECONDS)) {
+    public static void shutdownChannel(ManagedChannel channel, ILogger logger, long timeout) throws InterruptedException {
```
mtopolnik, line 55: Javadoc has a grammar error: "Shutdowns". Should be `Shuts down {@link ManagedChannel}.`

site/docs/how-tos/grpc.md
```diff
@@ -246,6 +246,15 @@ stage.mapUsingServiceAsyncBatched(bidiService,
 })
 ```
 
+If your batch takes close to or more than 1 s (including the network
```
mtopolnik, line 249: "If your batch takes **more than ~0.8 seconds** (including the network overhead), you should increase the value **of the** following properties so that **the clean shutdown succeeds**"

extensions/grpc/src/main/java/com/hazelcast/jet/grpc/impl/GrpcUtil.java
frant-hartm: Most of the internet agrees with you, I bet both will be correct in 20 years! :-)

extensions/grpc/src/main/java/com/hazelcast/jet/grpc/impl/GrpcProperties.java
```diff
@@ -0,0 +1,53 @@
...
+     * Time to wait for {@link StreamObserver#onCompleted()} confirmation when destroying a
+     * {@link BidirectionalStreamingService}
+     * You may want to increase this value when your gRPC calls take longer than 1 s to cleanly shutdown the service.
+     * <p>
+     * The default value is 1 s
+     *
+     * @since 4.2
+     */
+    public static final HazelcastProperty DESTROY_TIMEOUT
+            = new HazelcastProperty("jet.grpc.destroy.timeout", 1, SECONDS);
```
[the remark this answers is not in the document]
frant-hartm: Added.

PR #2256, opened 2020-05-15
Fix the markdown linter error.

ufukyilmaz:
Fix the linter error MD025-singletitle by reducing each header's level by one


cangencer:
verify

PR #2257, opened 2020-05-18
RocksDB State Backend

MohamedMandouh:
Introduces a state backend for Hazelcast Jet using [RocksDB ](https://github.com/facebook/rocksdb) persistent storage.
Batch processors that require large state, more than available memory, and don't need cutting-edge throughput such as joins and aggregations would benefit from this feature.

This is Jet's project for [Google Summer of Code 2020](https://summerofcode.withgoogle.com/projects/#5385534219747328)
This project is mentored by @mtopolnik 

Note by mentor:
The goal of the work underlying this PR was primarily research. It exposed some gaps when introducing file operations to cooperative Jet processors. Further work is needed to close these gaps. This further work will happen after GSoC.

Note:
Work on this PR inspired #2469 introducing In-Memory sorting feature for Jet that will be included in 4.3 release.


hazelcast-jet-core/src/test/java/com/hazelcast/jet/rocksdb/RocksMapTest.java
```diff
@@ -0,0 +1,111 @@
...
+    @After
+    public void cleanup() {
+        rocksDBStateBackend.deleteKeyValueStore();
+        serializationService.dispose();
+    }
+
+
+    @Test
+    public void test_getMap() {
+        assert rocksMap != null;
```
mtopolnik, line 56: Unit tests don't use the Java `assert` feature. Use `Assert.assertNotNull()` in this case and other methods as appropriate, and write the messages which are important when a test fails. Also, this isn't a great test because if there's an exception in `getMap`, it will fail in the init phase and not in this test.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/rocksdb/RocksMapTest.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Test
+    public void test_put_and_get() {
+        rocksMap.put("Hello", 1);
+        assert rocksMap.get("Hello") == 1;
```
mtopolnik, line 62: `assertEquals("rocksMap.get() returned the wrong value", 1, rocksMap.get("Hello"));`. Note the order, first the expected then the actual value.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/rocksdb/RocksMapTest.java
```diff
@@ -0,0 +1,111 @@
...
+        Map<String, Integer> map = new HashMap<>();
+        map.put("hello", 1);
+        map.put("bye", 2);
+        rocksMap.putAll(map);
+        assert rocksMap.get("bye") == 2;
+        assert rocksMap.get("hello") == 1;
+    }
+
+    @Test
+    public void test_putAll_and_getAll() {
```
mtopolnik, line 76: As the complexity of the test grows, you can see the advantage of adopting this structure:

```
// Given
... everything before the one thing that is under the test

// When
rocksMap.putAll(map);

// Then
... everything after that method call, including the assertions
```

This test isn't a unit test because it tries to test both `putAll` and `getAll` in one go. Tests should isolate individual functionalities.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/HashJoinP.java
```diff
@@ -111,10 +114,19 @@ public HashJoinP(
...
+        store = context.rocksDBStateBackend();
+    }
+
     @Override
     protected boolean tryProcess(int ordinal, @Nonnull Object item) {
         assert !ordinal0Consumed : "Edge 0 must have a lower priority than all other edges";
-        lookupTables.set(ordinal - 1, (Map) item);
+
+        RocksMap<Object, Object> map = store.getMap();
+        rocksMapList.add(map);
```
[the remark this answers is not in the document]
MohamedMandouh: yes, there's no need for it. I was trying to keep track of all maps acquired in order to release them. I'll fix it.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/rocksdb/RocksMapTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+    @Test
+    public void test_put_and_get() {
+        //Given
+        RocksMap<String, Integer> rocksMap = rocksDBStateBackend.getMap();
+        //When
+        rocksMap.put("key1", 1);
+        //Then
+        Assert.assertNotNull("the insert wasn't applied", rocksMap.get("key1"));
+        if (rocksMap.get("key1") != 1) {
```
mtopolnik, line 65:
```
@Test
public void when_putKeyValue_then_getKeyReturnsValue() {
    // Given
    String key = "key1";
    Integer value = 1;

    // When
    rocksMap.put(key, value);

    // Then
    assertEquals("rocksMap.get() didn't return the value used in rocksMap.put()", value, rocksMap.get(key));
}
```

Note:
- test method naming
- extracted parameters to protect from typos, also explicitly sets the type to `Integer` (relevant for `assertEquals` overload resolution) 
- usage of `assertEquals`, statically imported
- careful wording of error messages
- blank lines between phases for readability

hazelcast-jet-core/src/test/java/com/hazelcast/jet/rocksdb/RocksMapTest.java
```diff
@@ -0,0 +1,165 @@
...
+    public void test_update() {
+        //Given
+        RocksMap<String, Integer> rocksMap = rocksDBStateBackend.getMap();
+        //When
+        rocksMap.put("key1", 1);
+        rocksMap.put("key1", 2);
+        //Then
+        Assert.assertNotNull("the insert wasn't applied", rocksMap.get("key1"));
+        if (rocksMap.get("key1") != 2) {
+            Assert.fail("get returns the old value for key1");
```
mtopolnik, line 82: Watch out not to write misleading messages. You didn't explicitly test for the old value.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/rocksdb/RocksMapTest.java
```diff
@@ -0,0 +1,165 @@
...
+        //When
+        rocksMap.put("key1", 1);
+        //Then
+        Assert.assertNotNull("the insert wasn't applied", rocksMap.get("key1"));
+        if (rocksMap.get("key1") != 1) {
+            Assert.fail("get doesn't return the correct value for key1");
+        }
+    }
+
+    //Assert.assertEquals doesn't accept nullable parameters
```
mtopolnik, line 70: This is incorrect. It is about overload resolution and autoboxing, see above.

hazelcast-jet-core/src/test/java/com/hazelcast/jet/rocksdb/RocksMapTest.java
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+
+    @Test
+    public void test_putAll() {
+        //Given
+        RocksMap<String, Integer> rocksMap = rocksDBStateBackend.getMap();
+        Map<String, Integer> map = new HashMap<>();
+        //When
+        map.put("key1", 1);
```
mtopolnik, line 92: the local map should be populated in the `Given` phase, it is not the thing under test.

MohamedMandouh:
Now we don't create a RocksDB state backend for jobs that don't need it.
It's lazily created whenever a processor wants it.

MohamedMandouh:
Now we can specify the directory where RocksDB will operate through `JobConfig.attachDirectory()`
The directory will only be created when the first processor asks for the state backend.
Therefore jobs that don't do joins or aggregations don't need to specify a directory.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/rocksdb/RocksDBStateBackend.java
```diff
@@ -0,0 +1,169 @@
...
+    public void close() throws JetException {
+        if (db != null) {
+            for (final ColumnFamilyHandle cfh : columnFamilyHandles) {
+                try {
+                    db.dropColumnFamily(cfh);
+                } catch (RocksDBException e) {
+                    throw new JetException("Failed to delete column family", e);
+                }
+            }
+            db.close();
```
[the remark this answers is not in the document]
MohamedMandouh: okay, that's simpler.
`ExecutionContext` would delete the directory at the end anyway since it will be one of the  `tempDirectories` it manages.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/rocksdb/RocksDBStateBackend.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static com.hazelcast.jet.core.JetProperties.JET_HOME;
+
+
+/**
+ * Responsible for managing one RocksDB instance, opening, closing and deleting the database.
+ * Processors use this class to acquire any number of RocksDB-backed maps they require.
+ * Each map is associated with only one RocksDB ColumnFamily.
+ * There is only one instance of this class associated with each job.
```
mtopolnik, line 39: This is not currently true.

mtopolnik:
We'll reopen this PR when we figure out how to approach the problem of blocking file I/O vs. cooperative multithreading.

PR #2259, opened 2020-05-18
simplify download steps

cangencer:


cangencer:
> Is it ok to blatantly not care about Windows users and other such creatures? I don't mind, just asking.

I think they're used to it, I think this makes it easier for people who just wants to copy-paste.

PR #2272, opened 2020-05-25
Blog: Choice of JDK and GC 

mtopolnik:


site/docs/operations/jdk-gc-choice.md
```diff
@@ -0,0 +1,68 @@
+---
+title: Choice of JDK and GC
+description: Choosing the right JDK-GC combination is important.
+---
+
+The Java runtime has been evolving more rapidly in recent years and,
+after two decades, we finally got a new default garbage collector: the
+G1. Two more GCs are on their way to production and are available as
+experimental features: Oracle's Z and OpenJDK's Shenandoah. We at
+Hazelcast thought it was time to put all these new options to test and
```
jbartok, line 10: "to THE test" maybe?

site/docs/operations/jdk-gc-choice.md
```diff
@@ -0,0 +1,68 @@
...
+  Shenandoah
+
+And here are our overall conclusions:
+
+1. JDK 8 is an antiquated runtime. The Parallel collector enters huge
+   Major GC pauses and the G1, although better than that, is stuck in an
+   old version that uses just one thread when falling back to Full GC,
+   again entering very long pauses. Even on a moderate heap of 12 GB,
+   the pauses were exceeding 20 seconds.
+2. On more modern JDK versions, the G1 is one monster of a collector. It
```
jbartok, line 55: "A monster of a collector" maybe?

site/docs/operations/jdk-gc-choice.md
```diff
@@ -0,0 +1,68 @@
...
+  scenario 1, is still relevant
+
+We tried the following combinations:
+
+1. JDK 8 with the default Parallel collector and the optional G1
+2. JDK 11 with the default G1 collector
+3. JDK 14 with the default G1 as well as the experimental Z and
+  Shenandoah
+
+And here are our overall conclusions:
```
jbartok, line 48: Very nice, clean and concise, but maybe would be good to have a "where to go next" for some details for people who are interested. Anybody who finds any of your conclusions odd or doesn't agree with something has no way to "dig deeper" to see that you are actually right. Ie all the work and thoroughness you've put into these clean conclusions doesn't show. I know the conclusions are solid, because I've seen what's behind them, but simple readers won't.

site/docs/operations/gc-related-concerns.md
```diff
@@ -0,0 +1,118 @@
...
+
+1. JDK 8 with the default Parallel collector and the optional
+   ConcurrentMarkSweep and G1
+2. JDK 11 with the default G1 collector
+3. JDK 14 with the default G1 as well as the experimental Z and
+  Shenandoah
+
+These are our key findings:
+
+1. Due to the large improvements in the later versions, JDK version 8
```
viliam-durina, line 20:
```suggestion
1. Due to the large improvements in the latter versions, JDK version 8
```

mtopolnik: "Later" is appropriate, "latter" implies a former-latter distinction which we aren't making here.

viliam-durina: I'm not sure. I thought "latter" here means "recent". See definitions 2 and 3 [here](https://www.dictionary.com/browse/latter). When "later" is used, it's later than what? Maybe it could be "late versions".

mtopolnik: Versions later than 8. "Latter" could possibly still mean something, like the versions I mention above, after 8. But saying that wasn't my intention. "Later" conveys exactly the meaning I intended.

site/docs/operations/gc-related-concerns.md
```diff
@@ -0,0 +1,119 @@
+---
+title: Concerns Related to Garbage Collection
```
cangencer, line 2: this will probably look a little too long on the sidebar?

site/docs/operations/gc-concerns.md
```diff
@@ -0,0 +1,108 @@
+---
+title: Concerns Related to GC
+description: Choosing the right JDK-GC combination is important.
```
cangencer, line 3: a link to the blog posts somewhere in the article would make sense too

site/docs/operations/gc-concerns.md
```diff
@@ -0,0 +1,108 @@
...
+   Z is available in OpenJDK as well, we recommend it for scenarios
+   where G1's GC pauses are too long. The version of Z in JDK 14 has
+   crude ergonomics to decide on a good background GC thread count, in
+   some cases you can improve the throughput by using
+   `-XX:ConcGCThreads`.
+
+Here are some more details, organized by the type of workload you are
+using Jet for.
+
+## Moderate State, Low-Latency Stream Processing
```
viliam-durina, line 41: Maybe we should quantify what's moderate or large state.

mtopolnik: I think the example below, 100,000 devices, quantifies it.

viliam-durina: It's much easier to quantify in GBs of heap data. I tried to calculate the state size for your example and there's a missing piece: the window size and lag. I guess it's a window of N seconds, sliding by 10-20ms and 100k keys. The 500Hz is irrelevant, it doesn't increase the state size. So if I assume 5-second lag and 5-second window and 32 bytes per entry, the formula is `100,000 * 1000/15 * 10 * 32`, so the state is 2GB, which is moderate. But if you use 1-minute smoothing window, the state is 13GB. And both answers are correct, because the window size isn't given. Maybe I also misestimated the per-element state: if you use string-encoded UUID for device ID, it will be much more...

What I think most people would do is to run jvisualvm, do a GC and see how much heap they need.

mtopolnik: GB is not a good quantification either because these costs scale with object count and not size. If you have GBs of large strings, it will slow down the GC about the same as MBs of small objects.

viliam-durina: Still it gives a better idea IMO. Ppl have no idea how much memory or how many objects are needed for "100,000 devices at 500Hz".

PR #2296, opened 2020-06-03
[012] Mqtt Connector Design

gurbuzali:
Add description here

Checklist
- [X] Tags Set
- [X] Milestone Set

Fixes #NNNN

List of breaking changes:

* ..


site/docs/design-docs/015-mqtt-connector.md
```diff
@@ -0,0 +1,300 @@
+---
+title: 015 - Mqtt Connector
+description: Mqtt Connector (source and sink)
+---
+
+*Target Release*: 4.3
```
jbartok, line 6: Target release can't be 4.x, because the connector is in the contrib repo. I still think it shouldn't be there. It should be in the main repo, because this should be a professional source/sink (written by us). But we will probably need to take a new look at how we add such extensions, because we are starting to have many of them and they are all in the main distribution jar.

jbartok (changes requested):

site/docs/design-docs/015-mqtt-connector.md
```diff
@@ -0,0 +1,305 @@
...
+#### Fault Tolerance
+
+The sink is not fault-tolerant and does not save any state. In case of
+a restart, some messages can be duplicated.
+
+#### Error handling
+
+The sink uses sync client to publish the messages. Any error/exception
+encountered while publishing the messages will fail the job. User can
+configure the retrying of the messages by providing a retry function.
```
jbartok, line 301: I don't like this retry function, because it's a new abstraction for something we already have an abstraction for: `com.hazelcast.jet.retry.RetryStrategy`. Why not use that?

jbartok (changes requested):
Ok, so to sum up:
1. retryFn --> RetryStrategy?
2. source still not distributed, do we have a good reason for that?

mtopolnik (approved):

jbartok (approved):

PR #2323, opened 2020-06-12
Small typo fix in the GC blog post. 

ufukyilmaz:
Fixed the small typo in the GC benchmark blog post.


mtopolnik:
Small typo, but big effect. This corrects the number of keys and the order of their traversal. Thanks!

PR #2333, opened 2020-06-16
Use SerialTest for Elasticsearch integration tests

frant-hartm:
Having parallel tests causes multiple Elasticsearch containers to start,
making the tests slower on low-core machines (4C).

The main time consuming part of the tests now is the container start,
which can't be avoided - we already reuse containers between test
classes.


jbartok:
Screenshot 2020-06-17 at 10 42 15

PR #2336, opened 2020-06-17
Upgrade spotbugs plugin to 4.0.0

frant-hartm:
Removes 'WARNING: An illegal reflective access operation has occurred'


cangencer:
I think this should be applied on the enterprise repo as well.

PR #2349, opened 2020-06-22
Add jvm.options file

frant-hartm:
Adds simple way to persistently modify java options.

File supports comments with # (only at beginnig of a line).

We may extend this to show recommended / useful / commonly used options.

Checklist
- [x] Tags Set
- [x] Milestone Set


hazelcast-jet-distribution/src/bin-filemode-755/jet-start
```diff
@@ -18,6 +18,18 @@ if [ "${PROMETHEUS_PORT}" ]; then
     PROMETHEUS="-javaagent:${JET_HOME}/lib/jmx_prometheus_javaagent-${prometheus.version}.jar=${PROMETHEUS_PORT}:${JET_HOME}/config/jmx_agent_config.yaml"
 fi
 
+# Read jvm.options file
```
gurbuzali, line 21: can we move this to common.sh so that `jet` script can use it too ?

hazelcast-jet-distribution/src/root/config/jvm-client.options
```diff
@@ -0,0 +1,12 @@
+# JVM Configuration for bin/jet command
+
+# One option per line
+# Lines starting with # and empty lines are ignored
+
```
cangencer, line 5: we should add some more, for example to use g1gc if you're using Java 8 and also options to enable GC logging

hazelcast-jet-distribution/src/root/config/jvm.options
```diff
@@ -0,0 +1,49 @@
...
+# Enable JMX
+#-Dcom.sun.management.jmxremote
+#-Dcom.sun.management.jmxremote.port=9010
+#-Dcom.sun.management.jmxremote.authenticate=false
+#-Dcom.sun.management.jmxremote.ssl=false
+#-Dcom.sun.management.jmxremote.local.only=false
+
+# GC configuration
+
+# G1GC Configuration
```
cangencer, line 28: can we note here that this should be used if you using Java 8?

hazelcast-jet-distribution/src/root/config/jvm.options
```diff
@@ -0,0 +1,49 @@
...
+#-XX:+UseG1GC
+#-XX:MaxGCPauseMillis=50
+
+# ZGC
```
cangencer, line 32: and some notes here saying which versions support this.

hazelcast-jet-distribution/src/root/config/jvm.options
```diff
@@ -0,0 +1,49 @@
...
+
+# Shenandoah
+#-XX:+UseShenandoahGC
+
+## JDK 8 GC logging
+#-XX:+PrintGCDetails
+#-XX:+PrintGCDateStamps
+#-XX:+PrintTenuringDistribution
+#-XX:+PrintGCApplicationStoppedTime
+#-Xloggc:logs/gc.log
```
cangencer, line 43: will this use JET_HOME by default?

frant-hartm: Good point,

It would use the current directory, our tutorials expect that it is same as JET_HOME, but users may well be in the bin directory, or elsewhere.

I have added evaluation of the jvm.options file so we can use $JET_HOME there.

hazelcast-jet-distribution/src/root/config/jvm.options
frant-hartm: Added

hazelcast-jet-distribution/src/root/config/jvm.options
frant-hartm: Added.

frant-hartm:
@jbartok will test this works fine on mac, then we can merge.

jbartok:
Works fine on Macs.

PR #2359, opened 2020-06-24
Change default local parallelism from 2 to 4 for avro/file/json sources

### LINKED ISSUE #2352: Increase parallelism of avro source
Avro source seems quite CPU intensive when reading and can benefit from higher default parallelism. For example, going from 2 to 8 parallelism cut the reading speed in half.

gurbuzali:
Change default local parallelism from 2 to 4 for avro/file/json sources

Checklist
- [X] Tags Set
- [X] Milestone Set
- [X] Any breaking changes are documented

Fixes #2352

List of breaking changes:

* Change default preferedLocalParallelism from 2 to 4

### DIFF
--- a/extensions/avro/src/main/java/com/hazelcast/jet/avro/AvroSourceBuilder.java
+++ b/extensions/avro/src/main/java/com/hazelcast/jet/avro/AvroSourceBuilder.java
@@ -92,8 +92,8 @@ public final class AvroSourceBuilder<D> {
      * Any {@code IOException} will cause the job to fail. The files must not
      * change while being read; if they do, the behavior is unspecified.
      * <p>
-     * The default local parallelism for this processor is 2 (or 1 if just 1
-     * CPU is available).
+     * The default local parallelism for this processor is 4 (or available CPU
+     * count if it is less than 4).
      *
      * @param mapOutputFn the function which creates output object from each
      *                    record. Gets the filename and record read by {@code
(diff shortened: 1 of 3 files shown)


extensions/avro/src/main/java/com/hazelcast/jet/avro/AvroSourceBuilder.java
```diff
@@ -82,6 +82,18 @@
...
     }
 
+    /**
+     * Sets the preferred local parallelism of the source. Default values is
+     * {@code 4}.
+     *
+     * @since 4.2
+     */
+    @Nonnull
+    public AvroSourceBuilder<D> preferredLocalParallelism(int preferredLocalParallelism) {
```
cangencer, line 92: do we need this option? I think just to change the default would have been enough. Do we have this in any other source?

gurbuzali: I don't think we have, I'll remove then. So wdyt about `4` as default ?
Should we set it only for Avro source ? or both (avro and file) ?

cangencer:
We can use for 4 for both, as json parsing can also benefit from it

viliam-durina:
We normally document the default local parallelism in javadoc, most methods in `Sources` do, those for files don't.

gurbuzali:
Yes, we document them on builders, will change accordingly.

PR #2364, opened 2020-06-25
Add test for CDC join scenario

jbartok:
Intention: come up with a pipeline, which reads two different tables from a database via CDC and keeps an up-to-date map with their merged values, in this case customers and their orders. 

Particular details to pay attention to:
* I've added a version of `Sinks.mapWithEntryProcessor` where you can specify `maxParallelAsyncOps`; I have to set that to 1 for this scenario, to avoid Sink-to-IMap reordering due to multiple async requests
* I've changed multiple sinks to have partitioned input-edges, where a partitioning function is naturally available
* the test code probably runs on a local member only, a remote cluster will not know how to deserialize objects like `Customer`, `Order` and `OrdersToCustomer`; if/when we turn this into a tutorial or cookbook entry we will need to change it to work with simpler, standard objects

Point is to prove that we can achieve this joining functionality correctly.

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/extensions/cdc-postgres/src/test/java/com/hazelcast/jet/cdc/postgres/MultiTableCacheIntegrationTest.java
+++ b/extensions/cdc-postgres/src/test/java/com/hazelcast/jet/cdc/postgres/MultiTableCacheIntegrationTest.java
@@ -1,0 +1,314 @@
+/*
+ * Copyright (c) 2008-2020, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.jet.cdc.postgres;
+
+import com.hazelcast.jet.JetInstance;
+import com.hazelcast.jet.Job;
+import com.hazelcast.jet.cdc.ChangeRecord;
+import com.hazelcast.jet.cdc.Operation;
+import com.hazelcast.jet.cdc.ParsingException;
+import com.hazelcast.jet.cdc.RecordPart;
+import com.hazelcast.jet.cdc.impl.ChangeRecordImpl;
+import com.hazelcast.jet.function.TriFunction;
+import com.hazelcast.jet.pipeline.Pipeline;
+import com.hazelcast.jet.pipeline.Sinks;
+import com.hazelcast.jet.pipeline.StreamSource;
+import com.hazelcast.jet.pipeline.StreamStage;
+import com.hazelcast.map.EntryProcessor;
+import org.jetbrains.annotations.NotNull;
+import org.junit.Test;
+
+import java.io.Serializable;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.Statement;
+import java.util.Arrays;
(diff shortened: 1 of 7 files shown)


extensions/cdc-postgres/src/test/java/com/hazelcast/jet/cdc/postgres/MultiTableCacheIntegrationTest.java
```diff
@@ -0,0 +1,314 @@
...
+                            }
+                            return null;
+                        },
+                        (TriFunction<Sequence, RecordPart, Long, ChangeRecord>) (sequence, recordPart, aLong) -> null);
+    }
+
+    @NotNull
+    private static Map<Integer, OrdersOfCustomer> getIMapContent(JetInstance jet, String name) {
+        return jet.<Integer, OrdersOfCustomer>getMap(name).entrySet().stream()
+                .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
```
frant-hartm, line 157:
```suggestion
        return new HashMap<>(jet.getMap(name));
```

extensions/cdc-postgres/src/test/java/com/hazelcast/jet/cdc/postgres/MultiTableCacheIntegrationTest.java
```diff
@@ -0,0 +1,314 @@
...
+                        record -> (Integer) record.key().toMap().get("id"),
+                        CustomerEntryProcessor::new
+                ));
+
+        allRecords.filter(r -> r.table().equals("orders"))
+                .apply(this::fixOrdering)
+                .writeTo(Sinks.mapWithEntryProcessor(MAX_CONCURRENT_OPERATIONS, CACHE,
+                        record -> (Integer) record.value().toMap().get("purchaser"),
+                        OrderEntryProcessor::new
+                ));
```
frant-hartm, line 79: Will this work when loading larger amount of data from scratch?

extensions/cdc-postgres/src/test/java/com/hazelcast/jet/cdc/postgres/MultiTableCacheIntegrationTest.java
```diff
@@ -0,0 +1,314 @@
...
+            return false;
+        }
+
+        @Override
+        public String toString() {
+            return "source=" + source + ", sequence=" + sequence;
+        }
+    }
+
+    private static class OrdersOfCustomer implements Serializable {
```
frant-hartm, line 197: I hoped to have 

```
class Customer {
  Set<Order> orders;
}
```

is there anything blocking us from having this? The `OrdersOfCustomer` is kind of equivalent, but not as nice.

extensions/cdc-postgres/src/test/java/com/hazelcast/jet/cdc/postgres/MultiTableCacheIntegrationTest.java
jbartok: Ok.

extensions/cdc-postgres/src/test/java/com/hazelcast/jet/cdc/postgres/MultiTableCacheIntegrationTest.java
jbartok: We can change it. The thing is though, that: 1) neither the current nor the suggested solution will actually work in a real cluster, due to IMDG not knowing how to deserialize these classes 2) the purpose of this test is to prove that we have SOME way of doing this join correctly and getting the data into IMDG. So... beautifying it should happen probably some place else.

extensions/cdc-postgres/src/test/java/com/hazelcast/jet/cdc/postgres/MultiTableCacheIntegrationTest.java
jbartok: I don't know.

PR #2378, opened 2020-06-29
[005] Make Postgres dataloss on snapshot-stream sync less severe

jbartok:
Set snapshot mode to `exported` where data loss is one message only.

Checklist
- [x] Tags Set
- [x] Milestone Set

### DIFF
--- a/extensions/cdc-postgres/src/main/java/com/hazelcast/jet/cdc/postgres/PostgresCdcSources.java
+++ b/extensions/cdc-postgres/src/main/java/com/hazelcast/jet/cdc/postgres/PostgresCdcSources.java
@@ -53,13 +53,14 @@ public final class PostgresCdcSources {
      * to try to reconnect, at least for a certain amount of time. Future
      * versions will address the problem.
      * <p>
-     * <b>KNOWN ISSUE 2:</b> The Debezium PostgreSQL Connector this source is
-     * based on has data loss issues when synchronizing snapshots with event
-     * streams based on the Postgres WAL (Write-Ahead Log). For details see
-     * <a href="https://issues.redhat.com/browse/DBZ-2288">DBZ-2288</a>.
+     * <b>KNOWN ISSUE 2:</b> This source is based on the Debezium PostgreSQL
+     * Connector, which has some data loss issues when synchronizing snapshots
+     * with the stream of events coming from the Postgres WAL (Write-Ahead Log).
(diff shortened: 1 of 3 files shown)


extensions/cdc-postgres/src/main/java/com/hazelcast/jet/cdc/postgres/PostgresCdcSources.java
```diff
@@ -46,12 +46,20 @@ private PostgresCdcSources() {
...
-     * at least for a certain amount of time. Future versions will address the
-     * problem.
+     * <b>KNOWN ISSUE 1:</b> If Jet can't reach the database when it attempts to
+     * start the source or if it looses the connection to the database from an
+     * already running source, it throws an exception and terminate the
+     * execution of the job. This behaviour is not ideal, would be much better
+     * to try to reconnect, at least for a certain amount of time. Future
+     * versions will address the problem.
+     * <p>
+     * <b>KNOWN ISSUE 2:</b> The Debezium PostgreSQL Connector this source is
```
olukas, line 56: `The Debezium PostgreSQL Connector this source is based on has data loss issues when ...` - this should be rewritten, seem like there are some left words after rewritting

extensions/cdc-postgres/src/main/java/com/hazelcast/jet/cdc/postgres/PostgresCdcSources.java
```diff
@@ -46,12 +46,20 @@ private PostgresCdcSources() {
...
+     * based on has data loss issues when synchronizing snapshots with event
+     * streams based on the Postgres WAL (Write-Ahead Log). For details see
+     * <a href="https://issues.redhat.com/browse/DBZ-2288">DBZ-2288</a>.
+     * Currently the first streaming event that should come in right after the
```
olukas, line 60: It seems that more items still can be missed after some circumstances. It also not right for events which came a longer time after snapshot (i.e. if there is not update of DB during snapshotting)

extensions/cdc-postgres/src/main/java/com/hazelcast/jet/cdc/postgres/PostgresCdcSources.java
jbartok: No, it's actually correct, barring some comas missing, but I'll rephrase it to a simpler form, because that will make it clearer.

PR #2381, opened 2020-06-30
New front page for website

cangencer:
Updated design and new text for the front page.


jbartok:
Looks a bit weird on my machine, there is lots of space above the images and the text hangs out at the bottom. 

Screenshot 2020-07-02 at 09 30 52

jbartok:
Do we really need those 4 images? Nobody could guess what they symbolize without the text below them anyways and that probably means that they don't really add much value. Look at this other Docusaurus site for example, it's barebones and still looks good. Clean. https://pybowler.io/

jbartok:
These 3 sections below, two dark backgrounds and one light, at least in my mind don't create the intended impression, that of a table with alternatively coloured rows. It doesn't even look like a list of 3 things/messages to me. It looks more like a banner with something totally unrelated beneath. 

Maybe do something less visually striking, like this website: https://fbflipper.com/

Screenshot 2020-07-02 at 09 55 55

cangencer:
@jbartok can you have another look? is it better?

jbartok:
Yes, much better!

PR #2393, opened 2020-07-08
Simplify PartitionArrangement

viliam-durina:


hz-highfive:
https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3028/build_log.txt), [artifacts](https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3028/jet-oss-pr-builder-3028.zip)). 
<details>
<pre>

</details>
</pre>

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/init/PartitionArrangement.java
```diff
@@ -35,62 +33,58 @@
...
+    Map<Address, int[]> getRemotePartitionAssignment() {
+        return remotePartitionAssignment;
     }
 
     /**
-     * Determines for each processor instance the partition IDs it will be in charge of
-     * (processors are identified by their index). The method is called separately for
-     * each edge. For a distributed edge, only partitions owned by the local member need
-     * to be assigned; for a non-distributed edge, every partition ID must be assigned.
-     * Local partitions will get the same assignments in both cases, and repeating the
```
cangencer, line 54: are we removing this condition? if so, why?

viliam-durina: Yes, this PR is removing it? I'd reverse the question: why would we keep it? It's not used, it's not documented and it's very rarely beneficial. So the reason is to simplify the code.

cangencer: As discussed, currently we don't make use of this feature anywhere and it makes the code unnecessarily complex.

viliam-durina:
verify

PR #2396, opened 2020-07-09
Fix test failures on Windows

viliam-durina:
The error in `ResourceConfigTest` was 
```
Expected: (an instance of com.hazelcast.jet.JetException and exception with message a string containing "Not an existing, readable file: /i/do/not/exist")
     but: exception with message a string containing "Not an existing, readable file: /i/do/not/exist" message was "Not an existing, readable file: \i\do\not\exist"
```

### DIFF
--- a/hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/StreamFilesPTest.java
+++ b/hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/StreamFilesPTest.java
@@ -79,6 +79,11 @@ public class StreamFilesPTest extends JetTestSupport {
 
     @Before
     public void before() throws Exception {
+        // On Windows, "file changed" events are not delivered unless enough data is written
+        // or the file is closed. Some tests in this class rely that after a flush, the line is picked
+        // up by the processor.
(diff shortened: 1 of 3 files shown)


hazelcast-jet-core/src/test/java/com/hazelcast/jet/impl/connector/StreamFilesPTest.java
```diff
@@ -79,6 +79,11 @@
 
     @Before
     public void before() throws Exception {
+        // On Windows, "file changed" events are not delivered unless enough data is written
+        // or the file is closed. Some tests in this class rely that after a flush, the line is picked
+        // up by the processor.
+        assumeThatNoWindowsOS();
```
frant-hartm, line 85: You could do this check in `@BeforeClass` and avoid the nullcheck in `after()`

viliam-durina: The nullcheck in `after()` is good anyway if the creating of workdir or of the thread fails... I'm lazy now, I'm merging :)

PR #2421, opened 2020-07-29
Fix client config overriding by --targets option

### LINKED ISSUE #2373: When specifying --targets it overrides all config settings
When specifying --targets with `jet` command it overrides all the other config settings specified in `hazelcast-client.yaml` rather than just updating the relevant options.

caioguedes:
Make --targets update only the relevant options (cluster name and address) of `hazelcast-client.yaml`.

Checklist
- [ ] Tags Set
- [ ] Milestone Set

Fixes #2373


cangencer:
@frant-hartm could you have a look at this?

frant-hartm:
verify

PR #2423, opened 2020-07-31
"Sub-10 millisecond latency on Java: Concurrent GC with Green Threads"

mtopolnik:


hz-highfive:
The job `jet-oss-pr-builder-docs-no-build` of your PR failed ([log](
https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder-docs-no-build/251/build_log.txt), [artifacts](https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder-docs-no-build/251/jet-oss-pr-builder-docs-no-build-251.zip)). 
<details>
<pre>
--------------------------
---------SUMMARY----------
--------------------------
There are unexpected files.
Unexpected file: hazelcast-jet-core/src/main/java/org
</details>
</pre>

hazelcast-jet-core/src/main/java/org
```diff
@@ -0,0 +1 @@
+/Users/mtopol/dev/java/Benchmark/src/main/java/org
```
cangencer, line 1: this seems some error?

mtopolnik:
Yes, we'll add more advice to the Operations Guide.

viliam-durina:
Maybe we should mention that not all processing is on green threads. On the other hand, most non-cooperative processors don't use high parallelism, like we don't have 96 kafka source processors normally.

PR #2434, opened 2020-08-06
Add ElasticClients#client(java.lang.String) convenience method 

frant-hartm:
Takes host and optional scheme and port. Came up during the CDC/Elastic example. This is what most users most likely want to use.

### DIFF
--- a/extensions/elasticsearch/elasticsearch-5/src/main/java/com/hazelcast/jet/elastic/ElasticClients.java
+++ b/extensions/elasticsearch/elasticsearch-5/src/main/java/com/hazelcast/jet/elastic/ElasticClients.java
@@ -63,10 +63,11 @@ public final class ElasticClients {
      * http://elastic-host:9200}</pre>
      *
      * @see HttpHost#create(String)
+     * @since 4.3
      */
     @Nonnull
-    public static RestClientBuilder client(@Nonnull String s) {
-        return RestClient.builder(HttpHost.create(s));
+    public static RestClientBuilder client(@Nonnull String location) {
+        return RestClient.builder(HttpHost.create(location));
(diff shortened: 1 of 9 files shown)


extensions/elasticsearch/elasticsearch-5/src/main/java/com/hazelcast/jet/elastic/ElasticClients.java
```diff
@@ -53,6 +53,22 @@ public static RestClientBuilder client() {
...
+     *
+     * Valid examples:
+     * <pre>{@code elastic-host
+     * elastic-host:9200
+     * http://elastic-host:9200}</pre>
+     *
+     * @see HttpHost#create(String)
+     */
+    @Nonnull
+    public static RestClientBuilder client(@Nonnull String s) {
```
cangencer, line 68: parameter name should be more descriptive than `s`

extensions/elasticsearch/elasticsearch-5/src/main/java/com/hazelcast/jet/elastic/ElasticClients.java
```diff
@@ -53,6 +53,22 @@ public static RestClientBuilder client() {
...
+    /**
+     * Convenience method to create {@link RestClientBuilder} with given string, it must contain host, and optionally
+     * the scheme and a port.
+     *
+     * Valid examples:
+     * <pre>{@code elastic-host
+     * elastic-host:9200
+     * http://elastic-host:9200}</pre>
+     *
+     * @see HttpHost#create(String)
```
cangencer, line 65: missing `@since` tag

PR #2447, opened 2020-08-21
Fix cooperative emission in ReadMapP

### LINKED ISSUE #2448: Cooperative emission failure in ReadMapP
When `readMapP` has a projection to an object that doesn't implement `equals`, the cooperative emission can fail with 
```
java.lang.AssertionError: Different item offered after previous call returned false: expected=[Ljava.lang.Object;@71e5cb0a, got=[Ljava.lang.Object;@75da41ee
```
I think it works correctly if assertions are disabled, though I did not test.

viliam-durina:
Fixes #2448


viliam-durina:
verify

mtopolnik (approved):

PR #2453, opened 2020-08-24
Remove retries in a test, unify test across elastic versions

### LINKED ISSUE #2443: ElasticSinkBuilderTest test (elasticsearch-5) takes more than 6 minutes to run
Not sure if that run time is normal or not, expected or not, but it surely isn't acceptable. Let's investigate, try to fix and at least set the test to nightly run only (but it's actually too long for that too).

frant-hartm:
Fixes #2443

Checklist
- [x] Tags Set
- [x] Milestone Set


jbartok (approved):
Looks good to me, as long as it actually improves the test durations.

PR #2469, opened 2020-08-27
In-Memory Sorting

MohamedMandouh:
Introduces sorting feature for batch use cases of hazelcast jet.
Usage: 
`BatchStage.sort()` for sorting based on natural ordering of items.
`BatchStage.sort(ComparatorEx<V> comparator)` for sorting based on user-defined comparator.

Checklist
- [x] Tags Set
- [x] Milestone Set


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ConcurrentInboundEdgeStream.java
```diff
@@ -165,6 +177,51 @@ public ProgressState drainTo(Predicate<Object> dest) {
...
+                }
+                Object headObject = q.peek();
+                Object headItem;
+                if (headObject instanceof ObjectWithPartitionId) {
+                    headItem = ((ObjectWithPartitionId) headObject).getItem();
+                } else {
+                    headItem = headObject;
+                }
+                if (headItem == null) {
+                    return NO_PROGRESS;
```
mtopolnik, line 198: You may already have made progress by consuming some items in previous `while` loop iterations.

MohamedMandouh: Done

cangencer:
This also needs adding to /site/docs/api/stateful-transforms.md

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/Edge.java
```diff
@@ -371,6 +373,27 @@ public Edge isolated() {
...
+     *
+     * @since 4.3
+     */
+    public Edge monotonicOrder(@Nonnull ComparatorEx<Object> comparator) {
+        this.comparator = comparator;
+        return this;
+    }
+
+    /**
+     * Returns the comparator defined on this edge.
```
cangencer, line 388: we could add a reference back to monoticOrder here

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/BatchStage.java
```diff
@@ -66,6 +67,22 @@
...
+    @Nonnull
+    BatchStage<T> sort();
+
+    /**
+     * Attaches a stage that sorts the items based on user-defined comparator.
+     *
+     * @since 4.3
+     */
+    @Nonnull
+    BatchStage<T> sort(@Nonnull ComparatorEx<T> comparator);
```
cangencer, line 84: should be `? super T`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/pipeline/transform/SortTransform.java
```diff
@@ -0,0 +1,58 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import static com.hazelcast.function.FunctionEx.identity;
+import static com.hazelcast.jet.core.Edge.between;
+import static com.hazelcast.jet.core.processor.Processors.sortPrepareP;
+
+
+public class SortTransform<V> extends AbstractTransform {
```
cangencer, line 35: should use `T` here instead of `V`

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/SortPrepareP.java
```diff
@@ -0,0 +1,65 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+public class SortPrepareP<V> extends AbstractProcessor {
+    private final SortedSet<V> set;
```
cangencer, line 30: I don't think using SortedSet is correct here. It will skip duplicates. `PriorityQueue` would be more correct.

mtopolnik:
verify

mtopolnik:
This PR is still missing tests.

mtopolnik:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/SortP.java
```diff
@@ -0,0 +1,46 @@
...
+public class SortP<T> extends AbstractProcessor {
+    private final PriorityQueue<T> priorityQueue;
+    private final Traverser<T> resultTraverser;
+
+    public SortP(@Nullable Comparator<T> comparator) {
+        this.priorityQueue = new PriorityQueue<>(comparator);
+        this.resultTraverser = priorityQueue::poll;
+    }
+
+    @SuppressWarnings("unchecked")
```
ufukyilmaz, line 36: Override annotation is missing here.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/Processors.java
```diff
@@ -972,6 +976,22 @@ private Processors() {
...
+    /**
+     * Returns a supplier of processors for a vertex that sorts its input using
+     * a {@link PriorityQueue} and emits it in the {@code complete} phase.
+     * <p>
+     * The output edge of this vertex should be {@link Edge#distributed
+     * distributed} {@link Edge#monotonicOrder monotonicOrder} {@link
+     * Edge#allToOne allToOne} so it preserves the ordering when merging
+     * the data from all upstream processors.
+     *
+     * @since 4.3
```
ufukyilmaz, line 988: - Can you explain which kind of input data this processor can effectively sort (It seems to me that it cannot sort the streaming data completely, but I'm not sure.)?
- Is this processor's only use case to sort the data completely or can it be used for other purposes such as partial sorting? 
- Also, could you document what the comparator parameter does and what the type parameter is?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/BatchStage.java
```diff
@@ -66,6 +67,23 @@
     @Nonnull @Override
     BatchStage<T> rebalance();
 
+    /**
```
ufukyilmaz, line 70: Public APIs should include a detailed explanation, sample usage, and parameter and return value descriptions. Could you add them? You can refer to the methods in the `GeneralStage` interface.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/processor/Processors.java
mtopolnik: Also, we should change `V` to `T` for consistency.

MohamedMandouh: 1- It's only meant for batch use cases, it's not meant to be available in `StreamingStage` interface.
2- `SortP` can only do partial sorting, global sorting is done by combining partially sorted data.

ufukyilmaz: I think we should add this information to this processor's javadoc.

MohamedMandouh: Actually, this inspires adding a new Edge property that's a limit on the number of items allowed to travel over it.
This simple change introduces partial sorting to Jet!
usage: 
`items.partialSort(10)`, 
`items.partialSort(ComparatorEx.reverseOrder(), 10)`

MohamedMandouh: I guess partial sorting could still be done using `stage.sort().aggregate(topN())` or `stage.sort().aggregate(bottomN())` but that means sending all data first to the cluster member that does global sorting to complete the sorting stage then choosing top or bottom N items from them.
I assume having an edge that has a limit on the number of items that can be sent over it would be a better solution, that's why I suggested using the new `stage.partialSort()`

cangencer:
@MohamedMandouh if possible, you should do `partialSort` in another PR to keep the scope small for this one.

mtopolnik:
Just `stage.aggregate(topN(10))` will give you a sorted list as output, there is no need for any other upstream action. Theoretically, for optimization purposes, we could have every upstream processor limit its output to 10 items, and this could then become a higher-level stage operation that could be captured with `apply`.

mtopolnik (approved):

mtopolnik (approved):

mtopolnik:
veriy

mtopolnik:
verify

PR #2476, opened 2020-08-31
Fix source=true tag in metrics

### LINKED ISSUE #2475: A source vertex is not marked as source in metrics after a job restart
After a restart, each vertex is prepended with another vertex that reads the snapshot. So technically, such a vertex isn't a source, but from the user perspective it is.

The behavior was broken here in commit 3721541fb419792b2d5b00dc39468b8803c047d5. Currently we don't ignore inputs that restore the snapshot.

viliam-durina:
Fixes #2475


jbartok (approved):
Looks good. I've also run the manual tests I've initially used to explore the issue, the behaviour is correct after this fix.

PR #2490, opened 2020-09-03
collect receivedCount metric only if vertex is not a source

alex-dukhno:
Fix the case when `receivedCount` metric is collected for a vertex if it is not a source after job is restarted


viliam-durina:
I'm against this. It's good if metrics are raw, we should not try to make them look nice for the user. It's technically correct that a source vertex, after a restart, receives some items. Metrics are supposed to reveal problems.

alex-dukhno:
Closing, as I understood fix has to be on management centre side. Thanks!

PR #2532, opened 2020-09-18
Fix job hanging due to lost DONE_ITEM

### LINKED ISSUE #2158: ReceiverTasklet might never receive END_ITEM and cause the job to hang
This is a code review issue with no known reproducers.

Paraphrasing @devozerov:

Hazelcast has the `Connection.write(Packet)` method that returns `true` if the `Packet` was scheduled for sending. It is possible that the packet will never be delivered to the target member even if the target member is alive (examples: a half-opened socket, intermittent network failure that breaks the underlying channel). Hazelcast components are able to recover from it thanks to retry logic in the invocation service. But Jet doesn't use it and writes packets directly (we do the same in SQL).

Now, consider the following sequence of events:

1. `SenderTasklet` reaches EOS in `tryFillInbox()`, writes `DONE_ITEM`
2. We try to send the last packet in the call method, `Connection.write()` returns `true`
3. The connection is broken and the packet is never sent
4. Jet doesn't use this connection anymore as EOS has been reached, so the failure goes unnoticed
5. Remote member never receives the packet

viliam-durina:
The failure scenario was this:

- SenderTasklet reaches EOS in tryFillInbox(), writes DONE_ITEM

- We try to send the last packet in the call method, Connection.write()
returns true

- The connection is broken and the packet is never sent. The connection
is transparently reestablished, but the packet isn't sent again.

- Jet doesn't use this connection anymore in SenderTasklet as EOS has
been reached, so the failure goes unnoticed

- Remote member never receives the packet, the ReceiverTasklet will keep
waiting for the DONE_ITEM

The solution is that when the ReceiverTasklet creates data for the flow
control packet, it compares that the Connection object it was crated
with is the same.

We send the flow control packet regularly so any transparent
reconnection will cause the job to fail. Currently it would fail only if
some SenderTasklet tried to send something.

Fixes #2158

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/Networking.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/Networking.java
@@ -23,6 +23,7 @@ import com.hazelcast.internal.nio.Connection;
 import com.hazelcast.internal.nio.Packet;
 import com.hazelcast.internal.serialization.InternalSerializationService;
 import com.hazelcast.jet.impl.execution.ExecutionContext;
+import com.hazelcast.jet.impl.execution.ReceiverTasklet;
 import com.hazelcast.jet.impl.execution.SenderTasklet;
 import com.hazelcast.jet.impl.serialization.MemoryReader;
 import com.hazelcast.logging.ILogger;
@@ -31,6 +32,7 @@ import com.hazelcast.spi.impl.NodeEngineImpl;
 
 import java.io.IOException;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.concurrent.ScheduledFuture;
 
@@ -113,11 +115,11 @@ public class Networking {
     private void broadcastFlowControlPacket() {
         try {
             getRemoteMembers(nodeEngine).forEach(member -> uncheckRun(() -> {
-                final byte[] packetBuf = createFlowControlPacket(member);
+                Connection conn = getMemberConnection(nodeEngine, member);
+                final byte[] packetBuf = createFlowControlPacket(member, conn);
                 if (packetBuf.length == 0) {
                     return;
                 }
-                Connection conn = getMemberConnection(nodeEngine, member);
                 if (conn != null) {
                     conn.write(new Packet(packetBuf)
                             .setPacketType(Packet.Type.JET)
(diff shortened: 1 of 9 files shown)


hz-highfive:
https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3305/build_log.txt), [artifacts](https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3305/jet-oss-pr-builder-3305.zip)). 
<details>
<pre>
--------------------------
---------SUMMARY----------
--------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:checkstyle (default) on project hazelcast-jet-core: An error has occurred in Checkstyle report generation. Failed during checkstyle execution: There is 1 error reported by Checkstyle 8.29 with /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/checkstyle/checkstyle.xml ruleset. -> [Help 1]
--------------------------
---------ERRORS-----------
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/Networking.java:142: Line is longer than 123 characters (found 129). [LineLength]
--------------------------
</details>
</pre>

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/Networking.java
```diff
@@ -129,23 +131,26 @@ private void broadcastFlowControlPacket() {
...
     }
 
-    private byte[] createFlowControlPacket(Address member) throws IOException {
+    private byte[] createFlowControlPacket(Address member, Connection expectedConnection) throws IOException {
         try (BufferObjectDataOutput output = createObjectDataOutput(nodeEngine, lastFlowPacketSize)) {
-            final boolean[] hasData = {false};
+            boolean hasData = false;
             Map<Long, ExecutionContext> executionContexts = jobExecutionService.getExecutionContextsFor(member);
             output.writeInt(executionContexts.size());
-            executionContexts.forEach((execId, exeCtx) -> uncheckRun(() -> {
```
mtopolnik, line 137: Is this change mostly about "FP delete"? The new code has explicit types on everything, but with the types being Integer, Map etc., their true meaning is not known, and previously there were lambda parameters with descriptive names. I don't see a win here.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ReceiverTasklet.java
```diff
@@ -208,9 +217,14 @@ public int updateAndGetSendSeqLimitCompressed() {
...
      * @param timestampNow value of the timestamp at the time the method is called. The timestamp
      *                     must be obtained from {@code System.nanoTime()}.
+     * @param expectedConnection The connection to which the result will be sent. We use it
+     *                           to check that it's the same connection the tasklet was crated with.
      */
     // Invoked sequentially by a task scheduler
-    int updateAndGetSendSeqLimitCompressed(long timestampNow) {
+    int updateAndGetSendSeqLimitCompressed(long timestampNow, Connection expectedConnection) {
+        if (!Objects.equals(expectedConnection, memberConnection)) {
+            connectionChanged = true;
```
mtopolnik, line 226: Does this just make a problem less likely? Since the read and write of this variable are concurrent to each other?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/Networking.java
viliam-durina: I guess FP means functional programming. Then yes. It also creates less garbage and it's easier to debug.

The only real change is the argument to `updateAndGetSendSeqLimitCompressed` IIRC.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/ReceiverTasklet.java
viliam-durina: It should prevent the problem completely. The `connectionChanged` is checked in a loop so if it's changed to `true` (and it's only ever changed to `true`), it will be noticed, the error will be thrown and the job will fail and not get stuck.

mtopolnik (approved):

PR #2548, opened 2020-09-22
Adjust blog post titles for improved readability

jbartok:


jbartok:
@mtopolnik: how about like this?

mtopolnik (approved):

PR #2550, opened 2020-09-22
Fix remote ops failure handling in job coordination

### LINKED ISSUE #2533: Job can get stuck due to failed IMap operation
Due to the nature of IMDG, any `IMap` (or `DistributedObject` object) operation can fail if the connection between members is disrupted. The connection can be transparently reestablished so there is no topology change, but operations executed during that time or their responses can be lost. An `OperationTimeoutException` is thrown.

We manipulate IMaps a lot during job startup and cleanup. We don't defend against exceptions there, the job usually gets stuck in a broken state.

viliam-durina:
Fixes #2533


viliam-durina:
verify

mtopolnik (approved):

PR #2557, opened 2020-09-24
Add programming guide for error handling

jbartok:
Needed as a discoverable place about the whole error handling topic.

### DIFF
--- a/site/docs/api/error-handling.md
+++ b/site/docs/api/error-handling.md
@@ -28,12 +28,12 @@ interesting part.
 
 Before we dig deeper, let's list some typical error one might encounter:
 
-* *corrupt data*: e.g., user-provided code is parsing JSON lines, but
-  some of them are incorrectly formatted
-* *bugs in user code*: e.g., a user-provided filtering lambda throws a
-  `NullPointerException` for certain input
-* *connection to external services lost*: Jet sink writing into an
-  `IMap` loses connection to the remote Hazelcast cluster hosting it
+* *input data errors*: unexpected input data, e.g. unexpected null
+  value, out-of-range date (`0000-00-00`) or malformed JSON document
+* *coding errors*: e.g., a user-provided filtering lambda throws a
+  `NullPointerException` for certain input or stateful mapping code
+  contains memory leaks
+* *I/O errors*: network errors or remote service failures
 
 ## Specific Solutions
 
@@ -44,13 +44,14 @@ address them.
 
 Jet comes out of the box with some [in-memory distributed data
 structures](data-structures) which can be used as a data source or a
-sink. This are hosted by Hazelcast clusters, which can be the same as
+sink. These are hosted by Hazelcast clusters, which can be the same as
 the one Jet runs on or separate, remote ones.
 
-In this case, all error handling (for example, communications failures
-with the remote cluster, but not just) is handled transparently to Jet,
-by the Hazelcast cluster itself. For further details and configuration
-options, read the [relevant section in the Hazelcast
+In this case, a large portion of error handling (for example,
+communications failures with the remote cluster, but not just those) is
+handled transparently to Jet, by the Hazelcast cluster itself. For
+further details and configuration options, read the [relevant section in
+the Hazelcast
 manual](https://docs.hazelcast.org/docs/latest/manual/html-single/#handling-failures).
 
 ### Individual Sources & Sinks
@@ -92,7 +93,7 @@ jobs can be re-run from the beginning.
 
 Jobs with mutable state, those with a [processing
 guarantee](../architecture/fault-tolerance#processing-guarantee-is-a-shared-concern)
-set, achieve their fault tolerance by periodically saving [recovery
+set, achieve fault tolerance by periodically saving [recovery
 snapshots](../architecture/fault-tolerance#distributed-snapshot). When
 such a job fails, not only does its execution stop, but also its
 snapshots get deleted. This makes it impossible to resume it without
@@ -100,7 +101,9 @@ loss.
 
 To cope with this situation, you can configure a job to be suspended in
 the case of a failure, instead of failing completely. For details see
-[`JobConfig.setSuspendOnFailure`](/javadoc/{jet-version}/com/hazelcast/jet/config/JobConfig.html#setSuspendOnFailure(boolean)).
+[`JobConfig.setSuspendOnFailure`](/javadoc/{jet-version}/com/hazelcast/jet/config/JobConfig.html#setSuspendOnFailure(boolean))
+and
+[`Job.getSuspensionCause`](/javadoc/{jet-version}/com/hazelcast/jet/Job.html#getSuspensionCause()).
 
 Note: this option was introduced in Hazelcast Jet 4.3. In order to
 preserve the behavior from older versions, it is not the default, so it
@@ -126,11 +129,6 @@ upgrades](../enterprise/job-update). In that case you can:
 One caveat of the suspend-on-failure feature is that the latest snapshot
 is not a "failure snapshot." Jet can't take a full snapshot right at the
 moment of the failure, because its processors can produce accurate
-snapshots only when in a healthy state. Instead Jet simply keeps the
+snapshots only when in a healthy state. Instead, Jet simply keeps the
 latest periodic snapshot it created. Even so, the recovery procedure
 preserves the at-least-once guarantee.
-
-You can find more information on the topic in the feature's design
-document, for example, how clients that have submitted the job can
-[discover if it got suspended this
-way](../design-docs/012-improved-job-resilience#notifying-the-client).
--- a/site/docs/api/error-handling.md
+++ b/site/docs/api/error-handling.md
@@ -1,0 +1,136 @@
+---
+title: Error Handling Strategies
+description: Ways of coping with unexpected error in Hazelcast Jet.
+---
+
+There is unavoidable risk in running Jet jobs. They ingest data from
+external sources, the correctness and consistency of which can't be
+guaranteed. The processing they perform includes the running of user
+code, which is written by humans and, as such, is unavoidably imperfect.
+
+So even if we assume no bugs in the Jet framework itself, errors can and
+will happen. What options Jet offers for dealing with them is the topic
+of the next sections.
+
+## Scope
+
+A Jet cluster, at any given time, is running an arbitrary number of
+independent jobs in parallel. These jobs do not interact in any way, and
+the failure of one should not lead to any consequences for any other.
+When considering error handling, the basic unit of discussion will be
+the job, mainly what happens with it once it encounters a problem.
+
+The default way Jet reacts to an error is to fail the job that has
+produced it, but what one can do with such a failed job afterward is the
+interesting part.
+
+## Types of Errors
+
+Before we dig deeper, let's list some typical error one might encounter:
+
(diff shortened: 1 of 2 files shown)


site/docs/api/error-handling.md
```diff
@@ -0,0 +1,136 @@
...
+
+To cope with this situation, starting from Jet 4.3, jobs can be
+configured to be only suspended when failures occur, instead of having
+them fail completely. For details see
+[`JobConfig.setSuspendOnFailure`](/javadoc/{jet-version}/com/hazelcast/jet/config/JobConfig.html#setSuspendOnFailure(boolean)).
+Note: to preserve behavior from older versions, this configuration
+option is not the default, so it must be enabled explicitly.
+
+Having them in a suspended state preserves their snapshots and makes
+resumptions possible, once the root cause of the failure has been
```
nfrankel, line 110: Learned a new word: _resumption_

site/docs/api/error-handling.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Types of Errors
+
+Before we dig deeper, let's list some typical error one might encounter:
+
+* *corrupt data*: ex. user-provided code is parsing JSON lines, but some
+  of them are incorrectly formatted
+* *bugs in user code*: ex. user provided filtering lambda throws a
+  `NullPointerException` for certain input
+* *connection to external services lost*: Jet sink writing into an
```
nfrankel, line 35: Why use Jet here? We can also use an unreachable web-service.

nfrankel:
Great idea! I was thinking about that the other day, and I was missing this data.

1. Perhaps some code snippets within specific use-cases would be a good addition?
2. I'm a visual person. Perhaps add a diagram to make it more visual? A simple summary of IF a, THEN A, IF b, THEN b

site/docs/api/error-handling.md
```diff
@@ -0,0 +1,136 @@
...
+Restarted streaming jobs lacking mutable state can just resume
+processing the input data flow from the current point in time, batch
+jobs can be re-run from the beginning.
+
+### Processing Guarantees
+
+Jobs with mutable state, the ones that have [processing
+guarantees](../architecture/fault-tolerance#processing-guarantee-is-a-shared-concern)
+set, achieve their fault tolerance by periodically saving [recovery
+snapshots](../architecture/fault-tolerance#distributed-snapshot). When
```
vladoschreiner, line 97: So why do the jobs snapshot? It feels that there is no benefit unless one configures suspend on failure

It's important to say that snapshots are useful to overcome member failures.

site/docs/api/error-handling.md
```diff
@@ -0,0 +1,136 @@
...
+moment of the failure, because its processors can produce accurate
+snapshots only when in a healthy state. So the latest snapshot we will
+have available will be whatever has been saved at the last periodic
+snapshotting phase. But even so, the recovery procedure should preserve
+at-least-once guarantees.
+
+More information on the topic can also be found in the feature's design
+document, for example, how clients that have submitted the job can
+[discover if it got suspended this
+way](../design-docs/012-improved-job-resilience#notifying-the-client).
```
vladoschreiner, line 136: Catch exception on a processor level as another section. This is what we recommend to deal with user code.

vladoschreiner (changes requested):

site/docs/api/error-handling.md
```diff
@@ -0,0 +1,136 @@
...
+
+Before we dig deeper, let's list some typical error one might encounter:
+
+* *corrupt data*: ex. user-provided code is parsing JSON lines, but some
+  of them are incorrectly formatted
+* *bugs in user code*: ex. user provided filtering lambda throws a
+  `NullPointerException` for certain input
+* *connection to external services lost*: Jet sink writing into an
+  `IMap` loses connection to the remote Hazelcast cluster hosting the
+  data
```
viliam-durina: This discrimination is too anecdotal, we can do something more abstract:
```suggestion
* *input data error*: unexpected input data, e.g. unexpected null value, out-of-range date (`0000-00-00`) or malformed JSON document
* *coding errors*: e.g. NullPointerException, memory leaks, ...
* *I/O errors*: network errors, remote service failures
```
Also "ex." -> "for example," or "e.g."

site/docs/api/error-handling.md
```diff
@@ -0,0 +1,136 @@
...
+## Specific Solutions
+
+Some concrete types of error have solutions specifically tailored to
+address them.
+
+### In-Memory Data Structures
+
+Jet comes out of the box with some [in-memory distributed data
+structures](data-structures) which can be used as a data source or a
+sink. This are hosted by Hazelcast clusters, which can be the same as
```
viliam-durina, line 47: "This are" -> "These are"

site/docs/api/error-handling.md
```diff
@@ -0,0 +1,136 @@
...
+the one Jet runs on or separate, remote ones.
+
+In this case, all error handling (for example, communications failures
+with the remote cluster, but not just) is handled transparently to Jet,
```
viliam-durina, line 51: "but not just *those*"
"to Jet" -> "by Jet"

site/docs/api/error-handling.md
```diff
@@ -0,0 +1,136 @@
...
+address them.
+
+### In-Memory Data Structures
+
+Jet comes out of the box with some [in-memory distributed data
+structures](data-structures) which can be used as a data source or a
+sink. This are hosted by Hazelcast clusters, which can be the same as
+the one Jet runs on or separate, remote ones.
+
+In this case, all error handling (for example, communications failures
```
viliam-durina, line 50: "all error handling" -> _all errors_, really?

site/docs/api/error-handling.md
```diff
@@ -0,0 +1,136 @@
...
+restarting them (once the cause of the failure has been addressed).
+Restarted streaming jobs lacking mutable state can just resume
+processing the input data flow from the current point in time, batch
+jobs can be re-run from the beginning.
+
+### Processing Guarantees
+
+Jobs with mutable state, the ones that have [processing
+guarantees](../architecture/fault-tolerance#processing-guarantee-is-a-shared-concern)
+set, achieve their fault tolerance by periodically saving [recovery
```
viliam-durina, line 95: ~their~ fault tolerance

site/docs/api/error-handling.md
```diff
@@ -0,0 +1,136 @@
...
+such jobs fail, not only does their execution stop, but also their
+snapshots get deleted. This makes it impossible to resume them without
+loss.
+
+To cope with this situation, starting from Jet 4.3, jobs can be
+configured to be only suspended when failures occur, instead of having
+them fail completely. For details see
+[`JobConfig.setSuspendOnFailure`](/javadoc/{jet-version}/com/hazelcast/jet/config/JobConfig.html#setSuspendOnFailure(boolean)).
+Note: to preserve behavior from older versions, this configuration
+option is not the default, so it must be enabled explicitly.
```
viliam-durina: "this configuration option is not the default" -> "this option is disabled by default"

site/docs/api/error-handling.md
viliam-durina:
```suggestion
resumptions possible once the root cause of the failure has been
```

site/docs/api/error-handling.md
```diff
@@ -0,0 +1,136 @@
...
+  data
+* resubmit a new job based on the exported snapshot and the updated pipeline
+
+One caveat of the suspend-on-failure feature is that the latest snapshot
+is not a "failure snapshot." Jet can't take a full snapshot right at the
+moment of the failure, because its processors can produce accurate
+snapshots only when in a healthy state. So the latest snapshot we will
+have available will be whatever has been saved at the last periodic
+snapshotting phase. But even so, the recovery procedure should preserve
+at-least-once guarantees.
```
viliam-durina: It preserves exactly-once guarantee, but it's exactly-once semantics, not exactly-once processing. Exactly-once processing is impossible.

site/docs/api/error-handling.md
viliam-durina, line 136: We should not refer to description of important features in TDD. We can refer to TDD to read more why the design is as it is or so.

viliam-durina (commented):
I don't like the style very much. It's too much prose and quite imprecise, quick in generalizations and anecdotal. But I'll approve after my comments are addressed.

site/docs/api/error-handling.md
jbartok: I don't agree with the "to -> by" change. I wanted to say that Jet is not aware of the error, and for that "transparent to Jet" seems the right expression.

mtopolnik: Instead of this confusion just use the active voice and say explicitly who is the actor.

viliam-durina: @jbartok you're right, I missed it.

site/docs/api/error-handling.md
jbartok: I don't really understand this particular comment. Our processors are not capable of dealing with their own errors right now and we've decided to not implement that at the moment, because it's not something achievable with reasonable effort. So if we would recommend anything to users, that would be to catch exceptions from their own code, which is just common sense, I think. 

Did you mean something else?

site/docs/api/error-handling.md
```diff
@@ -0,0 +1,136 @@
...
+affecting our jobs.
+
+### Jobs Without Mutable State
+
+For many streaming jobs, specifically the ones which don't have any
+processing guarantee configured, the pipeline definition and the job
+config are the only parts we can identify as state, and those are
+immutable.
+
+Batch jobs also fall into the category of immutable state jobs.
```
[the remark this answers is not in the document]
jbartok: You are perfectly right, theoretically. Batch jobs do have state but they don't actually save it, so in practical terms they don't have state... You shouldn't even enable snapshotting for batch jobs, because there are known scenarios when it wouldn't work well (for example batch job with keyed windowing). Viliam can list a ton of related issues just from the top of his head... I will try to find the right formulation.

site/docs/api/error-handling.md
jbartok: I will emphasize that.

jbartok:
I've tried to address all concerns expressed so far. Pls. give it another round of comments. I can't say that I've enjoyed them so far, but the results seems (and hopefully is) better.

site/docs/api/error-handling.md
```diff
@@ -0,0 +1,128 @@
+---
+title: Error Handling Strategies
+description: Ways of coping with unexpected error in Hazelcast Jet.
+---
+
+Jet may experience errors on multiple levels:
+
+* on the *cluster level*: nodes can be removed from or added to the
+  cluster at any given time, due to, for example, internal errors, loss
+  of networking, or even scheduled maintenance
```
viliam-durina, line 10: - scheduled maintenance isn't an error condition: you can do an orderly shutdown

viliam-durina (approved):

jbartok:
Thanks @viliam-durina, I'll wait for @vladoschreiner 's approval too.

site/docs/api/error-handling.md
vladoschreiner: > if we would recommend anything to users, that would be to catch exceptions from their own code, which is just common sense

That is what I thought. We can skip this part for now.

vladoschreiner (approved):

PR #2567, opened 2020-09-26
Fix global aggregation with no input

### LINKED ISSUE #2560: stage.aggregate() produces no output when there's no input
This pipeline:

```java
static void main(String[] args) {
    p.readFrom(TestSources.items(0, 1, 2))
     .filter(i -> i > 2)
     .aggregate(AggregateOperations.counting())
     .writeTo(Sinks.logger());
}
```

produces no output to the sink because it didn't observe any item. We implement global aggregation in terms of keyed aggregation with a constant key function. Since, for the keyed case, you want to emit only those keys you observed, when you receive zero items, you emit zero output.

mtopolnik:
Even without any input, the aggregation stage must produce its output. 

This PR fixes the behavior of those aggregate operations that can represent an "empty" result (zero sum, empty list), but not of those whose result is `null` without any input (`minBy`, `maxBy`). These operations keep the current behavior of emitting nothing.

Fixes #2560. 

Checklist

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperation.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperation.java
@@ -235,7 +235,7 @@ public interface AggregateOperation<A, R> extends Serializable {
      * i}, as returned by {@link #accumulateFn(int)}.
      */
     @Nonnull
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings("rawtypes")
     AggregateOperation<A, R> withAccumulateFns(BiConsumerEx... accumulateFns);
 
     /**
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
@@ -1794,8 +1794,8 @@ public final class AggregateOperations {
      * calls.
      * <p>
      * Using {@code IMap} aggregations can be desirable when you want to make
-     * use of {@linkplain IMap#addIndex indices} when doing aggregations
-     * and want to use the Jet aggregations API instead of writing a custom
+     * use of {@linkplain IMap#addIndex indices} when doing aggregations and
+     * want to use the Jet aggregations API instead of writing a custom
      * {@link Aggregator}.
      * <p>
      * For example, the following aggregation can be used to group people by
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/aggregate/AggregateOperations.java
@@ -1047,7 +1047,7 @@ public final class AggregateOperations {
      * @see #groupingBy(FunctionEx, AggregateOperation1)
      * @see #toMap(FunctionEx, FunctionEx)
      */
-    @SuppressWarnings("unchecked")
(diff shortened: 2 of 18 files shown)


viliam-durina (commented):
With this change the behavior of `max` is now wrong. The aggregation op returns `null` in this case, but if we try to emit null, it will fail somewhere.

cangencer:
pr missing labels , milestones etc

mtopolnik:
>With this change the behavior of max is now wrong. 
Actually after the change the behavior is identical to the current one. I wonder what value we should return. For `double` we would be able to return `NaN` but for integers, I don't know. I think a new aggregate operation with a user-supplied sentinel value.

viliam-durina:
I think this is a broader issue we didn't think about. The processor, correctly, returns whatever the aggregate op returns when nothing was accumulated. I think we should add a comment to `aggregate` methods: 
>if the aggregate operation returns `null` from `finish` or `export` function, nothing is emitted
This is especially in `BatchStage.aggregate()` and `max`. And I think only min & max ever do that.

viliam-durina (approved):

viliam-durina (commented):
The change to `Tuple2-4` needs also to be applied to `Tuple5`.
It might be also useful to add a note to `AggregateOperations.maxBy` and `minBy`, because they are the only ones affected.

hz-highfive:
https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3384/build_log.txt), [artifacts](https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3384/jet-oss-pr-builder-3384.zip)). 
<details>
<pre>
--------------------------
---------SUMMARY----------
--------------------------
[ERROR] COMPILATION ERROR : 
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/hazelcast-jet-core/src/test/java/com/hazelcast/jet/aggregate/AggregateOperationsTest.java:[260,17] no suitable method found for allOf(com.hazelcast.jet.aggregate.AggregateOperation1<java.lang.Object,com.hazelcast.jet.accumulator.LongAccumulator,java.lang.Long>,com.hazelcast.jet.aggregate.AggregateOperation1<T,com.hazelcast.jet.accumulator.MutableReference<T>,T>)
--------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:testCompile (default-testCompile) on project hazelcast-jet-core: Compilation failure
--------------------------
---------ERRORS-----------
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/hazelcast-jet-core/src/test/java/com/hazelcast/jet/aggregate/AggregateOperationsTest.java:[260,17] no suitable method found for allOf(com.hazelcast.jet.aggregate.AggregateOperation1<java.lang.Object,com.hazelcast.jet.accumulator.LongAccumulator,java.lang.Long>,com.hazelcast.jet.aggregate.AggregateOperation1<T,com.hazelcast.jet.accumulator.MutableReference<T>,T>)
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/hazelcast-jet-core/src/test/java/com/hazelcast/jet/aggregate/AggregateOperationsTest.java:[260,17] no suitable method found for allOf(com.hazelcast.jet.aggregate.AggregateOperation1<java.lang.Object,com.hazelcast.jet.accumulator.LongAccumulator,java.lang.Long>,com.hazelcast.jet.aggregate.AggregateOperation1<T,com.hazelcast.jet.accumulator.MutableReference<T>,T>)
--------------------------
</details>
</pre>

PR #2581, opened 2020-10-02
Improve safety of working with job suspension cause programmatically

jbartok:
Attempting to make working programmatically with job suspension cause, for example in MC, less error prone (than relying on specific formatting of Strings).

Breaking changes:
* API

Checklist:

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2008-2020, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.jet.core;
+
+import com.hazelcast.jet.config.JobConfig;
+
+import javax.annotation.Nullable;
+
+/**
+ * Description of the cause that has lead to the job being suspended.
+ * <p>
+ * One reason for the job to be suspended is the user explicitly requesting it.
+ * Another reason is encountering an error during execution, while being
+ * configured to get suspended in such cases, instead of just failing. See
+ * {@link JobConfig#setSuspendOnFailure(boolean)}.
+ * <p>
+ * When the job has been suspended due to an error, then the cause of that
+ * error is provided.
+ */
+public interface JobSuspensionCause {
+
+    /**
+     * True if the job has been suspended due to an explicit request from the
+     * user.
+     */
+    boolean requestedByUser();
+
+    /**
+     * True if the job has been suspended due to encountering an error during
+     * execution.
+     */
+    boolean dueToError();
+
+    /**
+     * Provides the error that has lead to the suspension of the job, or returns
+     * null if the cause of the suspension is different.
+     */
+    @Nullable
+    Throwable errorCause();
+
+}
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
@@ -30,6 +30,8 @@ import javax.annotation.Nullable;
  * <p>
  * When the job has been suspended due to an error, then the cause of that
  * error is provided.
+ *
+ * @since 4.4
  */
 public interface JobSuspensionCause {
 
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
@@ -18,10 +18,10 @@ package com.hazelcast.jet.core;
 
 import com.hazelcast.jet.config.JobConfig;
 
-import javax.annotation.Nullable;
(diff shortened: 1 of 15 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
```diff
@@ -0,0 +1,57 @@
...
+ * Another reason is encountering an error during execution, while being
+ * configured to get suspended in such cases, instead of just failing. See
+ * {@link JobConfig#setSuspendOnFailure(boolean)}.
+ * <p>
+ * When the job has been suspended due to an error, then the cause of that
+ * error is provided.
+ *
+ * @since 4.4
+ */
+public interface JobSuspensionCause {
```
viliam-durina, line 36: What about adding a `@Nonnull getDescription()` method for simple cases that just want some information? The `toString()` can also delegate to this method.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * True if the job has been suspended due to an explicit request from the
```
mtopolnik, line 39: "True if the user explicitly suspended the job"

viliam-durina (changes requested):
Overall I think this change is unnecessary. But it's also broken currently, after my other comments are addressed, I'm willing to approve.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
```diff
@@ -0,0 +1,57 @@
...
+     * user.
+     */
+    boolean requestedByUser();
+
+    /**
+     * True if the job has been suspended due to encountering an error during
```
mtopolnik, line 45: "True if Jet suspended the job because it encountered an error during execution"

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
```diff
@@ -0,0 +1,57 @@
...
+     * execution.
+     */
+    boolean dueToError();
+
+    /**
+     * Provides the error that has lead to the suspension of the job, or returns
+     * null if the cause of the suspension is different.
```
mtopolnik, line 52: ... is not an exception.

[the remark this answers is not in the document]
jbartok: Yes. Should it throw an exception instead?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobSuspensionCauseImpl.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public void writeData(ObjectDataOutput out) throws IOException {
+        out.writeObject(error);
+    }
+
+    @Override
+    public void readData(ObjectDataInput in) throws IOException {
+        error = in.readObject();
```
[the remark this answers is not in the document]
jbartok: Ok. I can then just keep the error and its stack trace formatted as a string, as before. The idea was to have an extra, safe status type checking, but while adding that I saw no reason not to carry the actual Throwable. But then there is one.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
jbartok: Ok.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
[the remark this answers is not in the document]
viliam-durina: Two options:
- throw a reconstructed exc in the same way as `Job.join` does
- return a String...

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * True if Jet suspended the job because it encountered an error during
+     * execution.
+     */
+    boolean dueToError();
+
+    /**
+     * Describes the error that has led to the suspension of the job. Throws
+     * {@link UnsupportedOperationException} if job was not suspended due to an
```
viliam-durina, line 51: `UnsupportedOperationException` is for cases when some behavior is prescribed by the contract, but for some reason isn't implemented.

Better would be `IllegalStateException`.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/JobSuspensionCause.java
```diff
@@ -0,0 +1,63 @@
...
+     * error.
+     */
+    @Nonnull
+    String errorCause();
+
+    /**
+     * Describes the job suspension cause in a human readable form.
```
viliam-durina, line 58:
```suggestion
     * Describes the job suspension cause in a human-readable form.
```
Without the dash it's a form that's human and also readable :).

hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/SuspendExecutionOnFailureTest.java
```diff
@@ -113,7 +113,11 @@ public void when_jobSuspendedByUser_then_suspensionCauseSaysSo() {
         assertJobStatusEventually(job, RUNNING);
         job.suspend();
         assertJobStatusEventually(job, SUSPENDED);
-        assertEquals("Requested by user", job.getSuspensionCause());
```
viliam-durina, line 116: You can keep this assert against the `description` method.

viliam-durina (approved):

jbartok:
@viliam-durina, this last push should take care of review comments. Do you still think that this does not add value? Should I merge it or not, or should we at least discuss it with @mtopolnik too?

viliam-durina:
@jbartok I already approved, seems that others agreed with you that this change is useful. My opinion is that the the `additionalGain - additionalBurdenInComplexity` is slightly negative...

PR #2616, opened 2020-10-26
Reactivate TestContainers resource reaper in CDC network tests

jbartok:
CDC network tests are still leaking containers on Jenkins.... One situation which could lead to this is when the JVM/OS process running the tests is killed. To solve it I'm going to re-activate the Ryuk resource reaper for the tests. It solves the problem locally, let's see. (I can't really tell if this is what is actually going on usually on Jenkins, when containers get leaked.)

Checklist:


jbartok:
run-cdc-mysql-tests

jbartok:
run-cdc-postgres-tests

olukas (approved):

Issue #2623, opened 2020-10-29
com.hazelcast.jet.cdc.mysql.MySqlCdcIntegrationTest.customers fails on JDK 15


olukas:
_master_ (commit c6423f9a5bbcf1b8f24f4a0d27e0484c9a218c93)

Failed on Zulu JDK 15: http://jenkins.hazelcast.com/view/Jet%20-%20Compatibility/job/jet-oss-master-zulu-jdk15/2/testReport/com.hazelcast.jet.cdc.mysql/MySqlCdcIntegrationTest/customers/

Stacktrace:
```
java.lang.AssertionError: expected:<4> but was:<0>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at com.hazelcast.test.HazelcastTestSupport$9.run(HazelcastTestSupport.java:1125)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1356)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1458)
	at com.hazelcast.test.HazelcastTestSupport.assertEqualsEventually(HazelcastTestSupport.java:1122)
	at com.hazelcast.jet.cdc.mysql.MySqlCdcIntegrationTest.customers(MySqlCdcIntegrationTest.java:91)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:64)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
[...]

PR #2627, opened 2020-10-29
Ignore MySql CDC tests with JDK 15

olukas:
MySQL CDC connector is not able to work when Zulu JDK 15 is used (https://github.com/hazelcast/hazelcast-jet/issues/2623). This is problematic for nightly profile since it causes those tests take too long (see http://jenkins.hazelcast.com/job/jet-oss-master-nightly-zulu-jdk15/) and make nightly jobs unusable. This PR causes those test be ignored if JDK version contains `15.0.1`. We should remove those `Assume`s once it will be fixed. Job which ran with this change - http://jenkins.hazelcast.com/job/jet-oss-master-nightly-zulu-jdk15-temp/3/

Checklist:
- [N/A] Added a line in `hazelcast-jet-distribution/src/root/release_notes.txt` (for any non-trivial fix/enhancement/feature)
- [N/A] New public APIs have `@Nonnull/@Nullable` annotations
- [N/A] New public APIs have `@since` tags in Javadoc
- [N/A] Updated `examples/README.md` (when adding a new code sample)

### DIFF
--- a/extensions/cdc-debezium/src/test/java/com/hazelcast/jet/cdc/DebeziumCdcIntegrationTest.java
+++ b/extensions/cdc-debezium/src/test/java/com/hazelcast/jet/cdc/DebeziumCdcIntegrationTest.java
@@ -27,6 +27,7 @@ import com.hazelcast.jet.pipeline.Sinks;
 import com.hazelcast.jet.pipeline.StreamSource;
 import com.hazelcast.test.annotation.NightlyTest;
 import org.jetbrains.annotations.NotNull;
+import org.junit.Assume;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.testcontainers.containers.MySQLContainer;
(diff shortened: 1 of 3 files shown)


extensions/cdc-debezium/src/test/java/com/hazelcast/jet/cdc/DebeziumCdcIntegrationTest.java
```diff
@@ -50,6 +51,9 @@
 
     @Test
     public void mysql() throws Exception {
+        Assume.assumeFalse("https://github.com/hazelcast/hazelcast-jet/issues/2623",
+                System.getProperty("java.version").equals("15.0.1"));
```
jbartok, line 55: I don't fully understand why this error we are addressing happens with JDK 15 (and why it doesn't with JDK 11 too), but I suspect it would happen with any sub-version of JDK 15, so maybe make this check less specific? Check only the major version?

jbartok (approved):

extensions/cdc-debezium/src/test/java/com/hazelcast/jet/cdc/DebeziumCdcIntegrationTest.java
olukas: originally I hoped that we will fix it during Jet 4.4 but we cannot be sure. I'll change it to `System.getProperty("java.version").startsWith("15")`

PR #2636, opened 2020-11-02
Add Jet SQL reference manual

viliam-durina:


site/docs/sql/00-intro.md
```diff
@@ -0,0 +1,141 @@
...
+```json
+{ key: ticker, value: price }
+```
+
+To use a remote topic as a table in Jet, an EXTERNAL MAPPING must be
+created first. This maps the JSON messages to a fixed list of columns
+with data types:
+
+```java
+// TODO
```
devozerov, line 49: Possibly leftover (TODO). There are many TODOs in other places as well.

site/docs/sql/00-intro.md
```diff
@@ -0,0 +1,141 @@
...
+**Note:** _The service is in beta state. Behavior and API might change in
+future releases. Binary compatibility is not guaranteed between minor or
+patch releases._
+
+## Overview
+
+In the first release Jet SQL supports following features:
+
+- SQL Queries over [Apache Kafka topics](05-kafka-connector.md) and
+[Files (local and remote)](04-files-connector.md)
```
devozerov, line 20: Sorry if I am mistaken, but AFAIU from the SQL PR we support only IMap and Kafka in this release, but not files. Is this the case?

site/docs/sql/00-intro.md
```diff
@@ -0,0 +1,141 @@
+---
+title: SQL Introduction
+description: Introduction to Hazelcast Jet SQL features.
+---
+
+Hazelcast Jet SQL service executes distributed SQL statements over
+Hazelcast collections and external data sets.
```
devozerov, line 7: This first statement blurs the difference between IMDG and Jet SQL a bit. There is only one collection supported at the moment - `IMap`.  How an end-user should decide which product to use? Maybe instead we may mention here that Jet SQL extends IMDG SQL, providing new use cases and features as described in the "Overview" section?

site/docs/sql/00-intro.md
```diff
@@ -0,0 +1,141 @@
...
+JetInstance inst = ...;
+try (SqlResult result = inst.getSql().execute("SELECT ...")) {
+    for (SqlRow row : result) {
+        // Process the row.
+    }
+}
+
+```
+
+The query now runs in the Jet cluster and streams results to the Jet
```
devozerov, line 71: Since the streaming query possibly never ends, the example above is an infinite loop in the general case.  Maybe it makes sense to rewrite it a bit, to demonstrate how to finish the execution of the infinite query?

site/docs/sql/00a-basic-commands.md
```diff
@@ -0,0 +1,125 @@
...
+key can be overwritten multiple times.
+
+Some connectors support the `INSERT INTO` statement. If they do, the
+behavior is defined by the SQL standard. For example, the Apache Kafka
+connector supports it. Jet doesn't support `DELETE` or `UPDATE`
+statements.
+
+### Transactional behavior
+
+In SQL a statement is always atomic. In streaming SQL this is not
```
devozerov, line 107: Which statements are meant here - DML or DML+SELECT?

site/docs/sql/02-job-management.md
```diff
@@ -0,0 +1,99 @@
...
+```sql
+DROP JOB [IF EXISTS] job_name [WITH SNAPSHOT snapshot_name]
+```
+
+- `IF EXISTS`: don't throw an error if the job doesn't exist
+
+- `WITH SNAPSHOT`: export a named snapshot before cancelling the job
+  (enterprise feature)
+
+## Other job operations
```
devozerov, line 59: Instead of having an "other" section, we may move `ALTER JOB` to a job-related section, and introduce a separate snapshot section. It could make the structure a little bit more clear to users.

site/docs/sql/04-files-connector.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Files Connector
```
devozerov, line 2: Remove?

mtopolnik (approved):

site/docs/sql/00-intro.md
```diff
@@ -0,0 +1,202 @@
...
+- Windowed aggregation
+- JDBC
+
+## Example: How to query Apache Kafka using SQL
+
+We have a topic called `trades` with stock trade events. Trades are
+encoded as JSON messages:
+
+```json
+{ "ticker": "ABCD", "price": 5.5 }
```
olukas, line 42: price -> value

site/docs/sql/00-intro.md
```diff
@@ -0,0 +1,202 @@
...
+
+```sql
+CREATE MAPPING trades (
+    ticker VARCHAR,
+    value DECIMAL)
+TYPE Kafka
+OPTIONS (
+    valueFormat 'json',
+    "bootstrap.servers" '1.2.3.4',
+    "value.deserializer" 'org.apache.kafka.connect.json.JsonSerializer')
```
olukas, line 57: JsonSerializer -> JsonDeserializer

site/docs/sql/00-intro.md
```diff
@@ -0,0 +1,202 @@
...
+To use a remote topic as a table in Jet, first create an `EXTERNAL
+MAPPING`. It maps the JSON messages to a fixed list of columns with
+data types:
+
+```sql
+CREATE MAPPING trades (
+    ticker VARCHAR,
+    value DECIMAL)
+TYPE Kafka
+OPTIONS (
```
olukas, line 54: We should also put `key.deserializer` into `OPTIONS` here

site/docs/sql/00-intro.md
```diff
@@ -0,0 +1,202 @@
...
+inst.getSql().execute( /* query text */ );
+```
+
+A SQL query can now be used to read from the `trades` topic, as if it
+was a table:
+
+```java
+JetInstance inst = ...;
+try (SqlResult result = inst.getSql().execute("SELECT * FROM trades")) {
+    for (SqlRow row : result) {
```
olukas, line 74: Since `trades`  is streaming job which means it will never finish isn't this actually example of "how this should not be used"?

site/docs/sql/00-intro.md
```diff
@@ -0,0 +1,202 @@
...
+## Overview
+
+In the first release, Jet SQL supports the following features:
+
+- SQL Queries over [Apache Kafka topics](05-kafka-connector.md) and
+[files (local and remote)](04-files-connector.md)
+- Joining Kafka or file data with local IMaps (enrichment)
+- Filtering and projection using [SQL
+expressions](https://docs.hazelcast.org/docs/{imdg-version}/manual/html-single/index.html#expressions)
+- Aggregating data from files using predefined
```
olukas, line 23: "Aggregating data from files" - data from Kafka cannot be aggregated?

site/docs/sql/02-job-management.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+- `job_name`: a unique name identifying the job.
+
+- `query_spec`: the query the job will run. Currently we support `INSERT
+  INTO` and `SINK INTO` queries. `SELECT` is not supported by design
+  because it returns the results to the client.
+
+- `option_name`, `option_value`: the job configuration options. The list
+  of options matches the methods in the `JobConfig` class.
```
olukas, line 36: We should be more certain here because it seems that some of them are supported (for example `processingGuarantee`) but some of them not (for example `name`). I think it's also ambiguous - it's not clear whether name should be `processingGuarantee` or `setProcessingGuarantee` (it's can be understand from example but I think it should be mentioned here clearly).

site/docs/sql/00-intro.md
viliam-durina: What do you mean?

site/docs/sql/00-intro.md
viliam-durina: Why not? You can have a command-line client that prints the contents of a topic, until it's killed with ctrl-c. It's clearly noted below.

site/docs/sql/00-intro.md
viliam-durina: Nope, you can't aggregate a streaming data - you need complete data and they are never complete. For imap it doesn't work either, but in this case just because it isn't implemented yet.

site/docs/sql/02-job-management.md
viliam-durina: will create a new issue about this.

site/docs/sql/00-intro.md
olukas: I meant we have json `{ "ticker": "ABCD", "price": 5.5 }` and mapping `CREATE MAPPING trades (ticker VARCHAR, value DECIMAL)`. It should be json `{ "ticker": "ABCD", "value": 5.5 }`.

PR #2638, opened 2020-11-03
Remove reflections test dependency

frant-hartm:
Causes incorrect dependency resolution for hadoop module fat jar.
Causes incorrect dependency resolution for distribution,
resulting in missing reflections library and its dependencies
in NOTICE.

### DIFF
--- a/hazelcast-jet-distribution/src/root/NOTICE
+++ b/hazelcast-jet-distribution/src/root/NOTICE
@@ -114,6 +114,7 @@ Apache License, Version 2.0
   lang-mustache:7.0.0
   parent-join:7.0.0
   rank-eval:7.0.0
+  Javassist:3.26.0-GA
   LZ4 and xxHash:1.5.0
   SnakeYAML Engine:1.0
   snappy-java:1.1.1.3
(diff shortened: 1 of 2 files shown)


mtopolnik (approved):

frant-hartm:
verify

viliam-durina:
Looks like there's some issue when evaluting the `expectedException` here: https://github.com/hazelcast/hazelcast/blob/51b6a4be6d87466a31aaba2188fd7c8c88366c70/hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java#L419

It tries to load the `Predicate` class from guava: the `ReflectionUtils` class referenced on this line depends on `org.reflections.Reflections`, which depends on the guava `Predicate`. Looks like some maven mess-up: this dependency is available on other builds, but not on this one...

frant-hartm:
It is actually not possible to remove the reflections library because it is used by Hazelcast test infrastructure (imdg, not jet). For some reason, the code using it is not run on my local machine when running tests, but it is on Jenkins.

I have resolved the conflict in the Hadoop connector by specifying a newer version of guava for all modules.

@jbartok I did a quick check and we don't include the reflections library in cdc jar with deps. I don't know if it is needed, but it is a dependency of debezium core.

PR #2665, opened 2020-11-16
Improve SQL documentation

viliam-durina:


olukas (approved):

site/docs/sql/intro.md
```diff
@@ -117,19 +124,19 @@ statement:
 )
 ```
 
-The SINK INTO command will look like this:
+The `SINK INTO` command will look like this:
 
 ```sql
 SINK INTO latest_trades(__key, this)
```
vladoschreiner, line 130: Please replace `__key` and `this` with explicit references. It’s a default that’s not well known.

site/docs/sql/intro.md
```diff
@@ -49,14 +49,21 @@ data types:
 ```sql
 CREATE MAPPING trades (
```
vladoschreiner, line 50: Extend the sample so that it includes Kafka key as well. People use keys in their Kafka topics.

vladoschreiner (changes requested):
Make the intro a Tutorial. So, please add docker commands to setup the Kafka environment.

hz-highfive:
https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3633/build_log.txt), [artifacts](https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3633/jet-oss-pr-builder-3633.zip)). 
<details>
<pre>
There are changed files after building. This is forbidden, they should be add to PR or PR should be fixed. This mostly happens if there is any change in pom.xml which is not reflect in NOTICE file or if you are not using latest version of master branch.
 M hazelcast-jet-distribution/src/root/NOTICE
</details>
</pre>

site/docs/sql/intro.md
```diff
@@ -0,0 +1,306 @@
...
+
+These are some of the features on our roadmap:
+
+- Joins with arbitrary external data sources
+- Windowed aggregation of streaming data
+- JDBC
+
+## Installation
+
+You need the `hazelcast-jet-sql` module on your classpath. For
```
vladoschreiner, line 38: Mention client-server before embedded. Client-server is the main scenario.

site/docs/sql/intro.md
```diff
@@ -0,0 +1,306 @@
...
+If you use other features from SQL such as the Kafka or File connectors,
+also add the `hazelcast-jet-kafka` or `hazelcast-jet-hadoop` modules in
+the same way.
+
+## Example
+
+In this example we'll show how to query Apache Kafka using SQL and
+stream the messages to an IMap.
+
+First, make sure you have Apache Kafka up and running. Please follow any
```
vladoschreiner, line 74: Include commands to install Kafka. We need code that can be copy-pasted and works. Also, we want to control Kafka server versioning and ports.

site/docs/sql/intro.md
```diff
@@ -0,0 +1,306 @@
...
+)
+```
+
+The `valueFormat` option specifies the serialization format for the
+value. For other possible values see the [Kafka
+Connector](kafka-connector) page. The options not handled by Jet are all
+passed to the Kafka consumer or producer, such as the
+`bootstrap.servers` above.
+
+To submit the above query, use the Java API (we plan JDBC support and
```
vladoschreiner, line 118: We plan CLI at the first place. Please mention it.

site/docs/sql/intro.md
```diff
@@ -0,0 +1,306 @@
...
+FROM trades
+```
+
+It will put the fields from the Kafka topic into the IMap, replicating
+the `ticker` twice: once into the value and once into the key. Since we
+use the `ticker` as the key, every new trade for the same ticker will
+overwrite the previous trade for that ticker, hence the name
+`latest_trades`.
+
+### Creating a Streaming Job
```
vladoschreiner, line 252: I'd rather say "Creating a long-running job".

site/docs/sql/intro.md
```diff
@@ -0,0 +1,306 @@
...
+```
+
+The part after the `AS` keyword is same as the previous statement. The
+command will create a job named `trades_ingestion`. Now, even if the
+client disconnects, the cluster will continue running the job.
+
+### Cancelling the Job
+
+To cancel the job, use:
+
```
vladoschreiner, line 273: `latest_trades` map can be queried using SQL. Please provide a sample command and link to IMDG SQL. This is a marketing touch to present the big picture.

vladoschreiner (changes requested):

site/docs/sql/intro.md
viliam-durina: The current draft of CLI doesn't support multi-line commands. I'll add a todo to add it later.

vladoschreiner (approved):

PR #2682, opened 2020-11-19
Improve Kafka serde properties inference

gierlachg:
Added Kafka property inference for no key format.
Refactored Kafka property inference so it does not fail without Kafka on the classpath.
Improved tests.

Checklist:


hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/kafka/PropertiesResolver.java
```diff
@@ -49,11 +34,34 @@
     static final String VALUE_SERIALIZER = "value.serializer";
     static final String VALUE_DESERIALIZER = "value.deserializer";
 
-    static final String AVRO_SERIALIZER = "io.confluent.kafka.serializers.KafkaAvroSerializer";
-    static final String AVRO_DESERIALIZER = "io.confluent.kafka.serializers.KafkaAvroDeserializer";
+    // using strings instead of canonical names to not fail without Kafka on the classpath
```
viliam-durina, line 37: Do we care? It will fail sooner or later due to some `NoClassDefFoundError`. Especially the NullSerializer - it's not in kafka and we might rename/move it, the kafka serializers are unlikely to change FQN.

hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/kafka/PropertiesResolver.java
```diff
@@ -62,49 +70,51 @@ static Properties resolveProperties(Map<String, String> options) {
...
+            Properties properties
+    ) {
         String serializer = isKey ? KEY_SERIALIZER : VALUE_SERIALIZER;
         String deserializer = isKey ? KEY_DESERIALIZER : VALUE_DESERIALIZER;
 
         String format = options.get(isKey ? OPTION_KEY_FORMAT : OPTION_VALUE_FORMAT);
-        if (JAVA_FORMAT.equals(format)) {
+        if (format == null && isKey) {
+            properties.putIfAbsent(serializer, NULL_SERIALIZER);
+            properties.putIfAbsent(deserializer, NULL_DESERIALIZER);
```
viliam-durina, line 90: NullSerializer might be OK, we'll never try to write to the key. But NullDeserializer isn't, the key might be there, we just don't map it. I used `ByteArrayDeserializer` and `ByteArraySerializer`, these work with any data.

viliam-durina (changes requested):
We need to remove the `NullDeserializer`. Other comments are suggestions.

viliam-durina (approved):

PR #2686, opened 2020-11-20
Use FilterReduceRule

gierlachg:
Added `ReduceExpressionsRule.FILTER_INSTANCE` to the set of logical rules so for instance:
```
LogicalProject(v=[$0])
  LogicalFilter(condition=[AND(=($0, 10), IS NULL($0))])
    LogicalTableScan(table=[[hazelcast, public, t[projects=[0]]]])
```
results now in:
```
ValuesPhysicalRel(tuples=[[]])
```
instead of:
```
FullScanPhysicalRel(table=[[hazelcast, public, t[projects=[0], filter=AND(=($0, 10), IS NULL($0))]]])
```

Checklist:

### DIFF
--- a/hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/opt/logical/LogicalRules.java
+++ b/hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/opt/logical/LogicalRules.java
@@ -25,6 +25,7 @@ import org.apache.calcite.rel.rules.ProjectFilterTransposeRule;
 import org.apache.calcite.rel.rules.ProjectMergeRule;
 import org.apache.calcite.rel.rules.ProjectRemoveRule;
 import org.apache.calcite.rel.rules.PruneEmptyRules;
+import org.apache.calcite.rel.rules.ReduceExpressionsRule;
 import org.apache.calcite.tools.RuleSet;
 import org.apache.calcite.tools.RuleSets;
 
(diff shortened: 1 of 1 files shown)


hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/opt/logical/LogicalRules.java
```diff
@@ -40,6 +41,7 @@ public static RuleSet getRuleSet() {
                 FilterMergeRule.INSTANCE,
                 FilterProjectTransposeRule.INSTANCE,
                 FilterIntoScanLogicalRule.INSTANCE,
+                ReduceExpressionsRule.FILTER_INSTANCE,
```
viliam-durina, line 44: Shouldn't we also add PROJECT_INSTANCE and JOIN_INSTANCE?

gierlachg: `JOIN_INSTANCE` will be included with joins.
Regarding `PROJECT_INSTANCE` I could not find a case where it would be applicable apart from `CASE` which is not supported (yet).

viliam-durina: JOIN_INSTANCE - my mistake

PROJECT_INSTANCE: I thought it will simplify `a + 1 + 1` to `a + 2`, but it doesn't... Also that `a IS NULL AND a IS NOT NULL` is simplified to `false`, but it doesn't... I thought Calcite does this.

viliam-durina (approved):

PR #2689, opened 2020-11-20
Check IMap security for sql

### LINKED ISSUE #2662: Check IMap security in SQL
`SINK INTO my_map` currently can write to any map, regardless of the permissions.

viliam-durina:
Fixes #2662

### DIFF
--- a/hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/JetSqlBackend.java
+++ b/hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/JetSqlBackend.java
@@ -29,10 +29,10 @@ import com.hazelcast.jet.sql.impl.calcite.parser.JetSqlParser;
 import com.hazelcast.jet.sql.impl.opt.OptUtils;
 import com.hazelcast.jet.sql.impl.opt.logical.LogicalRel;
 import com.hazelcast.jet.sql.impl.opt.logical.LogicalRules;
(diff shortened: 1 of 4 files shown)


hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/JetSqlBackend.java
```diff
@@ -228,17 +235,49 @@ private ExecutionPlan toPlan(RelNode rel, List<String> fieldNames, OptimizerCont
...
+
+        physicalRel.accept(new RelShuttleImpl() {
+            @Override
+            public RelNode visit(TableScan scan) {
+                addPermissionForTable(scan.getTable(), ActionConstants.ACTION_READ);
+                return super.visit(scan);
+            }
+
+            @Override
+            public RelNode visit(RelNode other) {
```
gierlachg, line 261: Can't we use `visit(LogicalTableModify modify)`? Maybe we could skip the `table != null` condition in `addPermissionForTable`.

viliam-durina: The `InsertPhysicalRel` doesn't extend `LogicalTableModify`, but only `TableModify`. I actually don't know why `RelShuttle` has the `Logical` subclasses for almost everything except for `TableScan`...

gierlachg (approved):

PR #2699, opened 2020-11-23
Add logging for job status after job submit

### LINKED ISSUE #2426: Jet submit should log successful submission
It would be useful to print a confirmation of successful job submission (maybe with job id), especially for jobs which don't print anything to the user console.

Currently, a log is printed that the job is being submitted:

```
Submitting JAR '....' with arguments [...]
```

Then I need to check if the job is running by running `bin/jet list-jobs`

Possible output:
```
00:00:01.000 JobName id=... status=STARTING
00:00:02.000 JobName id=... status=STARTING
00:00:03.000 JobName id=... status=RUNNING
```

ufukyilmaz:
Fixes #2426 
Checklist:


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetBootstrap.java
```diff
@@ -124,6 +131,11 @@ public static synchronized void executeJar(@Nonnull Supplier<JetInstance> suppli
             String[] jobArgs = args.toArray(new String[0]);
             // upcast args to Object so it's passed as a single array-typed argument
             main.invoke(null, (Object) jobArgs);
+            List<Job> jobsAfterSubmit = JetBootstrap.supplier.get().getJobs();
+            printJobStatusInfo(jobsAfterSubmit
+                    .stream()
+                    .filter(job -> !jobsBeforeSubmit.contains(job))
```
ufukyilmaz, line 137: Here, I couldn't manage how to filter jobs submitted from a different client in a proper way.

hz-highfive:
https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3645/build_log.txt), [artifacts](https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3645/jet-oss-pr-builder-3645.zip)). 
<details>
<pre>
--------------------------
-------TEST FAILURE-------
--------------------------
[INFO] Results:
[INFO] 
[ERROR] Errors: 
[ERROR]   JetCommandLineTest.testTargetsMixin:519->testTargetsCommand:547->testTargetsCommandCluster:564 NullPointer
[ERROR]   JetCommandLineTest.test_submit_clientShutdownWhenDone:452->run:666 ? Execution
[INFO] 
[ERROR] Tests run: 39, Failures: 0, Errors: 2, Skipped: 0
[INFO] 

[ERROR] There are test failures.
</details>
</pre>

viliam-durina (commented):
I'm not sure this is what we wanted. Remember, the `main` method we called can do whatever. It can start and cancel jobs in a loop, it can never return. It might submit a batch job and join it: we'll still print `Job started.` even though the job is already over...

I imagined something simpler. When `newJob` returns just print `Job 1234-1234-1234-1234 submitted`. Or `Job 'foo' submitted.`, if the job has a name.

What others think?

mtopolnik:
I'm OK with Vilo's idea. @frant-hartm what do you think?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetBootstrap.java
```diff
@@ -138,6 +150,51 @@ public static synchronized void executeJar(@Nonnull Supplier<JetInstance> suppli
...
+            }
+            for (Job job : startedJobs) {
+                // The change of job statuses after the check above
+                // won't be a problem here. Because they cannot revert
+                // back to startup statuses.
+                if (job.getName() != null) {
+                    System.out.println("Job '" + job.getName() + "' is " + job.getStatus().name() + ".");
+                } else {
+                    System.out.println("Job " + job.getIdString() + " is " + job.getStatus().name() + ".");
+                }
```
frant-hartm, line 187: I think a date would be useful for debugging - e.g. when there is a large pause between submission and change to running state.

frant-hartm (changes requested):

frant-hartm:
This is basically what I asked for in the issue, so :+1:  from me.

Maybe the waiting for jobs to run/fail could be hidden behind an option though and the default would just print what vilo says.

mtopolnik:
Printing something when starting every job may also backfire if you start many jobs. You'll get too much noise in the output. Submitting jobs should primarily be a fire-and-forget thing, but maybe we can in the future add an option to await the completion of all the started jobs, as well as cancelling them if interrupted with ctrl-c. Then the user who wants that behavior won't have to write the quite tricky boilerplate to get it.

frant-hartm (approved):

PR #2728, opened 2020-12-04
Mqtt design decision regarding fault tolerence 

gurbuzali:
Checklist:
- [ ] Updated `examples/README.md` (when adding a new code sample)

### DIFF
--- a/site/docs/design-docs/015-mqtt-connector.md
+++ b/site/docs/design-docs/015-mqtt-connector.md
@@ -200,11 +200,17 @@ MqttSources.builder()
 
 #### Fault Tolerance
 
-MQTT protocol defines these levels of quality of service for subscribing
-to the topics: `AT_MOST_ONCE`, `AT_LEAST_ONCE`, `EXACTLY_ONCE`. But I've
-confirmed a loss of messages with `EXACTLY_ONCE` configuration even when
-the client restarted gracefully. I've tried both Paho client and HiveMQ
-client, the results are same.
+MQTT protocol defines these levels of quality of service (QoS) for
+subscribing to the topics: `AT_MOST_ONCE`, `AT_LEAST_ONCE`,
+`EXACTLY_ONCE`. For `AT_LEAST_ONCE` and `EXACTLY_ONCE`, client keeps
+in-flight messages in a persistence layer. The source uses an in-memory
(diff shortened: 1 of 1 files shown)


site/docs/design-docs/015-mqtt-connector.md
```diff
@@ -200,11 +200,24 @@ MqttSources.builder()
 
 #### Fault Tolerance
 
-MQTT protocol defines these levels of quality of service for subscribing
-to the topics: `AT_MOST_ONCE`, `AT_LEAST_ONCE`, `EXACTLY_ONCE`. But I've
-confirmed a loss of messages with `EXACTLY_ONCE` configuration even when
-the client restarted gracefully. I've tried both Paho client and HiveMQ
-client, the results are same.
+MQTT protocol defines these levels of quality of service (QoS) for
```
viliam-durina, line 203: An alternative text:
```
##### Source

To support fault tolerance for a source, Jet needs the remote system to
support one of these:

- provide an offset. After a failure, we'll use that offset and replay
the messages

- support manual acknowledgement. Jet will ack the messages in the
second phase of the snapshot

MQTT provides acknowledgements, however, the Paho client we use
acknowledges the messages automatically. There are other clients that
support manual acknowledgements, however the whole MQTT protocol is
designed for IoT and small resource usage. Many brokers (for example
Mosquitto) limit the number of non-acknowledged messages to a low value
(100 in case of Mosquitto). Therefore we can't really rely on this
feature.

We've also considered an option where Jet itself would save the messages
received since the last snapshot. We cannot save it to the snapshot
because the snapshot stores state created _before_ the snapshot is
taken, but we need to save messages received _after_ the snapshot. We
could save them to an IMap and maybe we could provide a persistence
implementation for Paho that we'll be able to "roll back". However, we
didn't implement a PoC for this option. We decided to not support any
fault tolerance for the MQTT source. It can be implemented in the future
if there's demand. Currently, if a fault-tolerant job fails, the source
will simply start where it left off and the messages acked since the
last snapshot will be lost.

##### Sink

The MQTT sink provides at-least-once guarantee - it sends the messages
synchronously. If a job fails and restarts, the same messages will be
sent again.
```

viliam-durina (approved):

PR #2743, opened 2020-12-09
Remove the remains of hazelcast-sql slim jar references

gierlachg:


viliam-durina (approved):
I don't understand much of this, but the build passed :)

gierlachg:
Basically, we don't want to rely on _old_ hazelcast-sql, that is:
```
<groupId>com.hazelcast</groupId>
<artifactId>hazelcast-sql</artifactId>
<classifier>original</classifier>
```
Future builds (with updated version) would simply fail because of no-such-dependency.

PR #2746, opened 2020-12-09
Fix Hadoop tests on Windows

viliam-durina:
- we must check `hasHadoopPrefix` first before
Paths.get(path).isAbsolute, because path such as `hdfs://something`
fails on windows due to the `:`

- there should be only one `assumeThatNoWindowsOS` to comment out if we
want to test on windows

- update hadoop to 3.3.0 due to
https://issues.apache.org/jira/browse/HDFS-14890 found in 3.2.1


hz-highfive:
https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3779/build_log.txt), [artifacts](https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3779/jet-oss-pr-builder-3779.zip)). 
<details>
<pre>
--------------------------
---------SUMMARY----------
--------------------------
[ERROR] Failed to execute goal org.codehaus.mojo:license-maven-plugin:1.19:add-third-party (default) on project hazelcast-jet-hadoop-all: There are some forbidden licenses used, please check your dependencies. -> [Help 1]
--------------------------
</details>
</pre>

extensions/hadoop/src/test/java/com/hazelcast/jet/hadoop/file/PathAndGlobFileSourceTest.java
```diff
@@ -51,6 +51,7 @@ public void shouldReadFilesMatchingGlobInTheMiddle() {
 
     @Test
     public void shouldReadFilesMatchingGlobInPath() {
+        assumeThatNoWindowsOS(); // * is not allowed in filename
```
frant-hartm, line 54: Why shouldn't this test run on windows? I understand why `shouldReadFileWithEscapedGlob` doesn't run because of * in filename is not allowed on windows, but here the * is in the path, matching any directory.

extensions/hadoop/src/test/java/com/hazelcast/jet/hadoop/file/TextFileFormatTest.java
```diff
@@ -35,10 +35,7 @@ public void readTextFileAsSingleItem() {
                                                       .glob("file.txt")
                                                       .format(FileFormat.text());
 
-        assertItemsInSource(source,
-                "Text contents of" + System.lineSeparator() +
-                        "the file." + System.lineSeparator()
-        );
+        assertItemsInSource(source, "Text contents of\nthe file.\n");
```
frant-hartm, line 38: This is what I had originally, but didn't pass on Windows due to different line endings. I think it depends on `git config core.autocrlf` value. Isn't setting this config to true a standard practice on windows? It results in the file having the following content:
`Text contents of\r\nthe file.\r\n`

frant-hartm (changes requested):

extensions/hadoop/src/test/java/com/hazelcast/jet/hadoop/file/PathAndGlobFileSourceTest.java
viliam-durina: I don't know, maybe hadoop supports glob chars, but it fails on [this line](https://github.com/hazelcast/hazelcast-jet/pull/2746/files#diff-7e44fe2919d2b846158485b1e0f5ec4ae9a388b2201f9e51613a5d287a35b60aR98) in `Paths.get`. `Paths` doesn't support glob characters. If this is the case, we have to check that the directory is absolute in some other way. Also, if I'm not mistaken, globs are matched differently on windows and linux, and java's `FileSystem.getPathMatcher()` specifies its own syntax (maybe it's that of linux, i don't know). 

Secondly, I don't know what should the `*` mean. The test is disabled for local file system - because in local file system we read only a single directory - will hadoop read multiple directories? Also, this behavior isn't specified:
```
     * The path must point to a directory. All files in the directory are
     * processed. The directory is not processed recursively.
```
IMO we should disallow this behavior...

extensions/hadoop/src/test/java/com/hazelcast/jet/hadoop/file/TextFileFormatTest.java
viliam-durina: I don't know, but we shouldn't depend on some git setting. It's typically better to not do any conversion. Today even Windows Notepad can open files with linux file endings. I have that option off.

I changed the text to load the file using `Files.readAllBytes` so that it can work regardless of that setting.

hz-highfive:
https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3836/build_log.txt), [artifacts](https://j-artifacts.s3.amazonaws.com/jet-oss-pr-builder/3836/jet-oss-pr-builder-3836.zip)). 
<details>
<pre>
--------------------------
---------SUMMARY----------
--------------------------
[ERROR] COMPILATION ERROR : 
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/file/MetadataResolver.java:[91,40] <T>FileSourceBuilder(java.lang.String) is not public in com.hazelcast.jet.pipeline.file.FileSourceBuilder; cannot be accessed from outside package
--------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:compile (default-compile) on project hazelcast-jet-sql: Compilation failure
--------------------------
---------ERRORS-----------
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/file/MetadataResolver.java:[91,40] <T>FileSourceBuilder(java.lang.String) is not public in com.hazelcast.jet.pipeline.file.FileSourceBuilder; cannot be accessed from outside package
--------------------------
[ERROR] /home/jenkins/jenkins_slave/workspace/jet-oss-pr-builder/hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/file/MetadataResolver.java:[91,40] <T>FileSourceBuilder(java.lang.String) is not public in com.hazelcast.jet.pipeline.file.FileSourceBuilder; cannot be accessed from outside package
--------------------------
</details>
</pre>

extensions/hadoop/src/test/java/com/hazelcast/jet/hadoop/file/PathAndGlobFileSourceTest.java
frant-hartm: Ok, it seems quite a mess. I have some bugs reported for the file connector, so will send it with those fixes.

frant-hartm (approved):
It's bit more than just _Fix Hadoop tests on Windows_, but for the better, so :+1:

Issue #2768, opened 2021-01-04
SQL console crashes


jerrinot:
it appears the sql console crashed when there is a null field:
```
sql> select * from lionking;
+--------------------+--------------------+--------------------+
|               __key|animal              |                 age|
+--------------------+--------------------+--------------------+
17:44:22.721 [ INFO] [c.h.c.LifecycleService] hz.client_1 [jet] [4.4-SNAPSHOT] [4.1.1] HazelcastClient 4.1.1 (20201222 - 27af558) is SHUTTING_DOWN
17:44:22.744 [ INFO] [c.h.c.i.c.ClientConnectionManager] hz.client_1 [jet] [4.4-SNAPSHOT] [4.1.1] Removed connection to endpoint: [10.0.0.49]:5701:f70c2a15-7d50-44c5-af21-c3af97e992f4, connection: ClientConnection{alive=false, connectionId=1, channel=NioChannel{/127.0.0.1:41435->/127.0.0.1:5701}, remoteAddress=[10.0.0.49]:5701, lastReadTime=2021-01-04 17:44:22.720, lastWriteTime=2021-01-04 17:44:22.717, closedTime=2021-01-04 17:44:22.722, connected server version=4.1.1}
17:44:22.745 [ INFO] [c.h.c.LifecycleService] hz.client_1 [jet] [4.4-SNAPSHOT] [4.1.1] HazelcastClient 4.1.1 (20201222 - 27af558) is CLIENT_DISCONNECTED
17:44:22.748 [ INFO] [c.h.c.LifecycleService] hz.client_1 [jet] [4.4-SNAPSHOT] [4.1.1] HazelcastClient 4.1.1 (20201222 - 27af558) is SHUTDOWN
com.hazelcast.jet.picocli.CommandLine$ExecutionException: Error while calling command (public void com.hazelcast.jet.server.JetCommandLine.sql(com.hazelcast.jet.server.JetCommandLine$Verbosity,com.hazelcast.jet.server.JetCommandLine$TargetsMixin)): java.lang.NullPointerException: Cannot invoke "Object.toString()" because the return value of "com.hazelcast.sql.SqlRow.getObject(int)" is null
	at com.hazelcast.jet.picocli.CommandLine.execute(CommandLine.java:1207)
	at com.hazelcast.jet.picocli.CommandLine.access$800(CommandLine.java:141)
	at com.hazelcast.jet.picocli.CommandLine$RunAll.handle(CommandLine.java:1416)
[...]

PR #2769, opened 2021-01-04
Fix sql cli crash

ufukyilmaz:
Formerly, the cli was crashing because of the absence of the null check for the column values. This PR adds null check to solve this issue.

Fixes: #2768 

Checklist:

### DIFF
--- a/hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetCommandLine.java
+++ b/hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetCommandLine.java
@@ -1128,10 +1128,15 @@ public class JetCommandLine implements Runnable {
         builder.append("|");
         int columnCount = row.getMetadata().getColumnCount();
         for (int i = 0; i < columnCount; i++) {
-            String colValue = row.getObject(i).toString();
-            colValue = sanitize(colValue, colWidths[i]);
+            String colString;
+            Object columnValue = row.getObject(i);
+            if (columnValue != null) {
+                colString = sanitize(columnValue.toString(), colWidths[i]);
+            } else {
+                colString = "NULL";
+            }
             builder.style(AttributedStyle.BOLD.foreground(PRIMARY_COLOR));
-            appendAligned(colWidths[i], colValue, alignments[i], builder);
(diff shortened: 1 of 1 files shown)


hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetCommandLine.java
```diff
@@ -1128,10 +1128,15 @@ private static void printRow(SqlRow row, int[] colWidths, Alignment[] alignments
         builder.append("|");
         int columnCount = row.getMetadata().getColumnCount();
         for (int i = 0; i < columnCount; i++) {
-            String colValue = row.getObject(i).toString();
-            colValue = sanitize(colValue, colWidths[i]);
+            String colString;
+            Object columnValue = row.getObject(i);
```
mtopolnik, line 1132: I think this would be more readable with the `? :` operator.

hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetCommandLine.java
```diff
@@ -1128,10 +1128,15 @@ private static void printRow(SqlRow row, int[] colWidths, Alignment[] alignments
...
+            if (columnValue != null) {
+                colString = sanitize(columnValue.toString(), colWidths[i]);
+            } else {
+                colString = "NULL";
+            }
```
viliam-durina, line 1137: Equivalent, but shorter code:
```suggestion
            String colString = sanitize(String.valueOf(row.<Object>getObject(i)), colWidths[i]);
```

viliam-durina (approved):

hazelcast-jet-all/src/main/java/com/hazelcast/jet/server/JetCommandLine.java
mtopolnik: Won't `String.valueOf` return a lowercase `null`, though?

viliam-durina: Yes, it will:
image

PR #2775, opened 2021-01-07
Improve logging a bit

### LINKED ISSUE #2770: Reduce logging noise when starting
Consider moving this from INFO to DEBUG:
```
2021-01-05 09:57:10,789 [ INFO] [main] [c.h.j.i.e.TaskletExecutionService]: Creating idler with jet.idle.cooperative.min.microseconds=50µs,jet.idle.cooperative.max.microseconds=250µs

2021-01-05 09:57:10,789 [ INFO] [main] [c.h.j.i.e.TaskletExecutionService]: Creating idler with jet.idle.noncooperative.min.microseconds=50µs,jet.idle.noncooperative.max.microseconds=1000µs
```

I am assuming vast majority of people don't care about these values everytime when they are starting a new instance.

jbartok:
Change the logging level for some statements, use utility methods in others.

Fixes #2770

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
@@ -768,7 +768,9 @@ public class JobCoordinationService {
     void scheduleSnapshot(MasterContext mc, long executionId) {
         long snapshotInterval = mc.jobConfig().getSnapshotIntervalMillis();
         ExecutionService executionService = nodeEngine.getExecutionService();
-        logFine(logger, "%s snapshot is scheduled in %d ms", mc.jobIdString(), snapshotInterval);
+        if (logger.isFineEnabled()) {
+            logger.fine(mc.jobIdString() + " snapshot is scheduled in " + snapshotInterval + "ms");
+        }
         executionService.schedule(COORDINATOR_EXECUTOR_NAME,
                 () -> mc.snapshotContext().startScheduledSnapshot(executionId),
                 snapshotInterval, MILLISECONDS);
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
@@ -768,9 +768,7 @@ public class JobCoordinationService {
     void scheduleSnapshot(MasterContext mc, long executionId) {
         long snapshotInterval = mc.jobConfig().getSnapshotIntervalMillis();
         ExecutionService executionService = nodeEngine.getExecutionService();
-        if (logger.isFineEnabled()) {
-            logger.fine(mc.jobIdString() + " snapshot is scheduled in " + snapshotInterval + "ms");
-        }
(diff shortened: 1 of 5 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JobCoordinationService.java
```diff
@@ -768,9 +768,7 @@ void scheduleRestart(long jobId) {
     void scheduleSnapshot(MasterContext mc, long executionId) {
         long snapshotInterval = mc.jobConfig().getSnapshotIntervalMillis();
         ExecutionService executionService = nodeEngine.getExecutionService();
-        if (logger.isFineEnabled()) {
-            logger.fine(mc.jobIdString() + " snapshot is scheduled in " + snapshotInterval + "ms");
-        }
+        logFine(logger, "%s snapshot is scheduled in %d ms", mc.jobIdString(), snapshotInterval);
```
viliam-durina, line 771: This was done this way because `mc.jobIdString()` isn't a simple getter. Users mostly have this logging off. Again below, multiple times.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/util/LoggingUtil.java
```diff
@@ -50,6 +50,12 @@ public static void logFine(ILogger logger, String template, Object arg1, Object
         }
     }
 
+    public static void logFine(ILogger logger, String template, Object... args) {
```
viliam-durina, line 53: We didn't have this variant because it creates a temporary `Object[]` for the 3rd argument. Maybe just add a variant with 4 arguments.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
```diff
@@ -378,17 +378,17 @@ private void runTasklet(TaskletTracker t) {
                 dismissTasklet(t);
             }
 
-            if (finestLogEnabled) {
+            if (fineLogEnabled) {
                 long elapsedMs = NANOSECONDS.toMillis((System.nanoTime() - start));
                 if (elapsedMs > COOPERATIVE_LOGGING_THRESHOLD) {
-                    logger.finest("Cooperative tasklet call of '" + t.tasklet + "' took more than "
+                    logger.fine("Cooperative tasklet call of '" + t.tasklet + "' took more than "
```
viliam-durina, line 384: I wonder if we won't have too many of these during soak tests, it's just 5ms. It also does `nanoTime` twice for each tasklet call, that's quite a lot and not negligible in benchmarks if FINE level is enabled.

viliam-durina (approved):

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
olukas: @viliam-durina We run soak tests with debug level. Even on debug level it produces dozens of GB logs.

viliam-durina: I'd rather keep the level as it was. I've never used this metric actually...

Issue #2781, opened 2021-01-08
Missing EXTERNAL NAME for MAPPING NAME


olukas:
When the name different from Kafka topic name should be used then there's a possibility to do it by `objectName` option. However it's not much visible. Provide `EXTERNAL NAME` for `MAPPING NAME` makes it more visible.

If topic `trades` should have name `trades2` in Jet then currently following has to be used:
```
CREATE EXTERNAL MAPPING trades2 (
    ...)
TYPE Kafka
OPTIONS (
    'objectName' = 'trades',
    ...
);
```
This seems to be more clear and visible:
```
CREATE EXTERNAL MAPPING trades EXTERNAL NAME trades (
    ...)
TYPE Kafka
OPTIONS (
    ...
);
```

gierlachg:
Addressed with #2791

PR #2791, opened 2021-01-11
Support mapping EXTERNAL NAME

gierlachg:
Added support for mappings `EXTERNAL NAME`. At the same time removed special handling of `objectName` option.

Addresses #2781

### DIFF
--- a/site/docs/sql/imap-connector.md
+++ b/site/docs/sql/imap-connector.md
@@ -21,7 +21,7 @@ arbitrary record to derive the schema and serialization type. You need
 to have at least one record on each cluster member for this to work. If
 you can't ensure this, create the mapping explicitly.
 
-## Serialization options
+## Serialization Options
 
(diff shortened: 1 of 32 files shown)


viliam-durina (changes requested):
We miss documentation changes. I'd add it to this PR to have a better feeling what it takes to explain the concept to the user...

site/docs/sql/imap-connector.md
```diff
@@ -128,10 +128,25 @@ the class using reflection and use its properties as column names. It
 recognizes public fields and JavaBeans-style getters. If some property
 has a non-primitive type, it will be mapped under the `OBJECT` type.
 
-## External Name
+## External Mapping Name
```
viliam-durina, line 131: I think we can omit this section. It's clear enough from the common documentation... Also for Kafka

viliam-durina (approved):

PR #2795, opened 2021-01-12
Fix double __key column

### LINKED ISSUE #2793: Double __key column
The following command will create a mapping with two `__key` columns:
```
CREATE MAPPING m (
__key BIGINT EXTERNAL NAME id,
field BIGINT
)
TYPE kafka
OPTIONS (
   'keyFormat'='json',
   'valueFormat'='json')
```
This query:
```sql
SELECT * FROM m
```
will fail with:
```
Column '__key' is ambiguous
```

viliam-durina:
Multiple issues fixed:

- maybeAddDefaultField() only looked ad key's or value's fields, not at
all user fields. It would add __key for a second time if it is mapped to
value

- if __key is added as a value field, resolvePrimitiveSchema() will just
add it again. Not it leaves the key unmapped (which will fail in case of
imap because it's mandatory, but not in case of kafka)

Fixes #2793

### DIFF
--- a/extensions/kafka/src/test/java/com/hazelcast/jet/kafka/impl/KafkaTestSupport.java
+++ b/extensions/kafka/src/test/java/com/hazelcast/jet/kafka/impl/KafkaTestSupport.java
@@ -81,6 +81,7 @@ public class KafkaTestSupport {
 
     private KafkaServer kafkaServer;
     private KafkaProducer<Integer, String> producer;
+    private KafkaProducer<String, String> stringStringProducer;
     private int brokerPort = -1;
     private String brokerConnectionString;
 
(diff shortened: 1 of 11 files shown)


hz-highfive:
The job `jet-oss-pr-builder` of your PR failed.
<details>
<pre>
--------------------------
-------TEST FAILURE-------
--------------------------
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   StreamKafkaPTest.when_partitionAddedWhileJobDown_then_consumedFromBeginning:403->HazelcastTestSupport.assertTrueEventually:1350->HazelcastTestSupport.assertTrueEventually:1248->lambda$when_partitionAddedWhileJobDown_then_consumedFromBeginning$8:404 expected:<1> but was:<0>
[INFO] 
[ERROR] Tests run: 35, Failures: 1, Errors: 0, Skipped: 0
[INFO] 

[ERROR] There are test failures.
</pre>
</details>

gierlachg (approved):

PR #2816, opened 2021-01-15
Simplify code in Kinesis connector

viliam-durina:


extensions/kinesis/src/main/java/com/hazelcast/jet/kinesis/impl/sink/ShardCountMonitorImpl.java
```diff
@@ -42,35 +47,26 @@
      */
     private static final double RATIO_OF_DESCRIBE_STREAM_RATE_UTILIZED = 0.1;
 
-    private final AtomicInteger shardCount;
```
gurbuzali, line 45: the reason this was an AtomicInteger was to share the shard count with other processors I guess. So only one of the processors (first one) runs periodically and sets the shard count (to this AtomicInteger) and other processors (those having a noop monitor) just gets the shard count from that AtomicInteger.

With this change other processors will see shard count as 0 all the time

viliam-durina: Ah, true. So I broke it... I couldn't make sense of the code.

viliam-durina: Fixed

extensions/kinesis/src/main/java/com/hazelcast/jet/kinesis/impl/sink/ShardCountMonitorImpl.java
```diff
@@ -42,35 +48,27 @@
...
     private final RandomizedRateTracker describeStreamRateTracker;
     private final RetryTracker describeStreamRetryTracker;
 
     private Future<DescribeStreamSummaryResult> describeStreamResult;
     private long nextDescribeStreamTime;
 
-    public ShardCountMonitor(
-            int totalInstances,
+    ShardCountMonitorImpl(
+            int knownShardCountInstances,
```
jbartok, line 59: I find this name misleading. It suggests that it's an integer related to the count of shards, as opposed to the count of monitors. Morever, it sounds a bit like an uncertain value, that there is an unknown part of the count and this is like an estimate. How about using `monitorInstances` or keeping `totalInstances` instead?

jbartok (commented):
Code changes look ok to me, besides the one comment about naming. I'll run my tests on it too.

extensions/kinesis/src/main/java/com/hazelcast/jet/kinesis/impl/sink/ShardCountMonitorImpl.java
viliam-durina: I renamed to `knownShardCounterInstances`.

>it sounds a bit like an uncertain value
That was exactly the point. There can be others calling shard count, this value includes the number of counters we know about. And that's why we run at 10% or the limit...

jbartok (approved):
Tests have confirmed that all works as it should.

PR #2818, opened 2021-01-15
Ensure optional file libs are used only when SQL table is queried

### LINKED ISSUE #2817: All commands fail after creating CSV file mapping without the CSV module
The CSV module is missing on classpath, and you execute this commands:
```
sql> CREATE MAPPING csv_trades (id TINYINT, name VARCHAR)
   > TYPE File
   > OPTIONS ('format'='csv',
   >     'path'='/path/to/hazelcast-jet-4.4', 'glob'='trades.csv');
OK
sql>  SELECT * FROM csv_trades;
Could not find ReadFileFnProvider for FileFormat: csv. Did you provide correct modules on classpath?
```
It's expected to fail, however after the above code any other SQL command fails:
```
sql> select * from generate_series(1);
Could not find ReadFileFnProvider for FileFormat: csv. Did you provide correct modules on classpath?
```

gierlachg:
Fixes #2817

### DIFF
--- a/hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/file/AvroMetadataResolver.java
+++ b/hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/file/AvroMetadataResolver.java
@@ -44,7 +44,7 @@ final class AvroMetadataResolver extends MetadataResolver<GenericRecord> {
     protected Metadata resolveMetadata(List<MappingField> resolvedFields, Map<String, ?> options) {
         return new Metadata(
                 toFields(resolvedFields),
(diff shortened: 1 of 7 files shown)


viliam-durina (changes requested):
There's still an issue. I created a mapping with bad path (not absolute). Even in this branch, all commands subsequently fail:
```
sql> CREATE MAPPING csv_trades (id TINYINT, name VARCHAR)
   > TYPE File
   > OPTIONS ('format'='csv',
   > 'path'='/path/to/hazelcast-jet-4.4', 'glob'='trades.csv');
OK
sql> show jobs;
Provided path must be absolute, path: foo
```

viliam-durina (approved):

Issue #2810, opened 2021-01-14
Failure in StreamKafkaPTest.when_partitionAddedWhileJobDown_then_consumedFromBeginning


frant-hartm:
Failed in this PR
https://github.com/hazelcast/hazelcast-jet/pull/2796

on Jenkins
http://jenkins.hazelcast.com/job/jet-oss-pr-builder/4067/testReport/junit/com.hazelcast.jet.kafka.impl/StreamKafkaPTest/when_partitionAddedWhileJobDown_then_consumedFromBeginning/

Other failures in that build should not be related. 
There are no changes related to Kafka in the PR

Stacktrace:
```
java.lang.AssertionError: expected:<1=1> but was:<0=0>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:120)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at com.hazelcast.jet.kafka.impl.StreamKafkaPTest.lambda$when_partitionAddedWhileJobDown_then_consumedFromBeginning$8(StreamKafkaPTest.java:405)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1248)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1350)
	at com.hazelcast.jet.kafka.impl.StreamKafkaPTest.when_partitionAddedWhileJobDown_then_consumedFromBeginning(StreamKafkaPTest.java:404)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
[...]

PR #2823, opened 2021-01-18
Fix race in StreamKafkaP

viliam-durina:
Fixes the 2nd cause in
https://github.com/hazelcast/hazelcast-jet/issues/2810#issuecomment-762066997,
the original cause still remains an issue.


viliam-durina:
It actually fixes both reports in #2810 - previously i completely omitted the `Not resuming job` - in both cases the `job.resume()` was called before the job actually finished suspension and was still running.

frant-hartm (approved):

PR #2846, opened 2021-01-22
Fix adding of 'key.serializer' options

### LINKED ISSUE #2845: `key.serializer` not added when primitive format is used
When you have a kafka topic mapping with the primitive key format and omit the serialization option, the `key.serializer` isn't added.

```
CREATE MAPPING my_topic (...)
TYPE Kafka
OPTIONS (
    'keyFormat'='int',
    /* ..., no 'key.serializer' option */
)
```

Then the query fails with `Missing 'key.serializer' option` from Kafka. If the long format is used (`'keyFormat'='java', 'keyJavaClass'='java.lang.Integer'`), it works.

viliam-durina:
Fixes #2845


gierlachg:
What about creating an utility (?) that based on format and class would give you a class?
This way:
- you don't have to modify original options
- you could reuse it with a non-key-value connector using Java serialization (if we ever have one)
- you remove java serialization specific code from a generic `KvMetadataResolvers`

mtopolnik (approved):

PR #2854, opened 2021-01-27
Implement processed offset feedback in CDC sources

### LINKED ISSUE #2841: CDC source doesn't send confirmation for processed offsets back to the DB
Issue discovered due to community inquiry: https://groups.google.com/g/hazelcast/c/eqhTvsBcUEQ

The problem is that while the CDC source consumes change events properly, it doesn't send a feedback to the DB about having processed them properly. Not an issue for MySQL, because there the Debezium connector handles binlog reading explicitly. It is however an issue for Postgres, where the Debezium connector uses Postgres' inbuilt replication slots and those slots need to clean up internal data, based on the missing feedback. 

Part of the Debezium Postgres Connector docs discussing the effects: https://debezium.io/documentation/reference/1.2/connectors/postgresql.html#postgresql-wal-disk-space

Implementation-wise, to see what the connector should be doing, look for `task.commitRecord(...)` and `task.commit()` in [EmbeddedEngine](https://github.com/debezium/debezium/blob/1.2/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java).

To do:
* add missing functionality
* figure out how this functionality is supposed to work in Pipelines without processing guarantees (normally that's when we want to flush/confirm, on snapshots, and there are none of those without processing guarantees)
* write integration tests that explicitly check the `confirmed_flush_lsn` advancing in the DB
* write tests for the heartbeat mechanism, mentioned in the Debezium Connector docs also work as expected and can be used for inactive tables 
* try to get fix released ASAP
* notify the original reporter of the problem in the above mentioned community thread

jbartok:
Fixes #2841 

Checklist:


jbartok:
run-cdc-postgres-tests

viliam-durina (changes requested):
1. in case we commit on snapshot, we do so in the 1st phase. That's wrong, the job can still fail before the snapshot is finished and the committed offsets will be gone.

2. i think we don't have to provide the option to set commit strategies to the user. In case of a snapshotted job we should always commit at snapshot time (in the 2nd phase). For non-snapshotted jobs we should commit after each batch. If committing later is a significant performance issue for the DB, we should provide an option to configure the source without a guarantee, even though the rest of the pipeline is ex-once or at-least-once. Notice that if we commit anytime sooner than in the phase 2, the guarantee is not at-least-once, but there's no guarantee. So i don't think this is a sensible option.

jbartok:
run-cdc-postgres-tests

jbartok:
run-cdc-mysql-tests

jbartok:
run-cdc-debezium-tests

hz-highfive:
The job `jet-oss-pr-builder-cdc-debezium-tests` of your PR failed.
<details>
<pre>

</pre>
</details>

jbartok:
run-cdc-debezium-tests

hz-highfive:
The job `jet-oss-pr-builder-cdc-debezium-tests` of your PR failed.
<details>
<pre>

</pre>
</details>

jbartok:
run-cdc-debezium-tests

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSourceP.java
```diff
@@ -255,6 +342,17 @@ public OffsetStorageReader offsetStorageReader() {
         }
     }
 
+    private static String getCause(Exception e) {
+        StringBuilder sb = new StringBuilder();
+        if (e.getMessage() != null) {
+            sb.append(" : ").append(e.getMessage());
```
viliam-durina, line 348: There shouldn't be a space before a `:`

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSourceP.java
```diff
@@ -48,81 +54,159 @@
...
+        return false;
+    }
+
+    @Nullable
+    protected abstract T map(SourceRecord record);
+
+    private void reconnect(RuntimeException re) {
+        if (reconnectTracker.shouldTryAgain()) {
+            logger.warning("Connection to database lost, will attempt to reconnect and retry operations from " +
+                    "scratch" + getCause(re));
```
viliam-durina, line 195: It might be useful to log the stack trace here and elsewhere where `getCause` is used:
```suggestion
                    "scratch" + getCause(re), re);
```
Currently we don't even have the exception class, just the message and the message of the first cause.

I personally would drop the whole `getCause` method and just rely on the stack trace:
```suggestion
                    "scratch: " + re, re);
```

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSourceP.java
```diff
@@ -273,24 +371,40 @@ public OffsetStorageReader offsetStorageReader() {
...
+        long periodMillis = periodString == null ? DEFAULT_COMMIT_PERIOD_MS : Long.parseLong(periodString);
+        return TimeUnit.MILLISECONDS.toNanos(periodMillis);
+    }
+
     private static boolean getClearStateOnReconnect(Properties properties) {
         String s = (String) properties.get(RECONNECT_RESET_STATE_PROPERTY);
         return Boolean.parseBoolean(s);
     }
 
+    private static void log(Object value, String property, String name, ILogger logger) {
```
viliam-durina, line 394: (tentative)
The parameter order could might match the order in the log: `logger, name, property, value`

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSourceP.java
```diff
@@ -48,81 +54,159 @@
...
+            if (commitPeriod == 0) {
+                task.commit();
             }
+
+            traverser = Traversers.traverseIterable(records)
+                    .flatMap(record -> {
+                        T t = map(record);
+                        return t == null ? Traversers.empty() :
+                                eventTimeMapper.flatMapEvent(t, 0, extractTimestamp(record)
+                                );
```
viliam-durina, line 176: The `)` should be on the same line

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSourceP.java
```diff
@@ -273,24 +371,40 @@ public OffsetStorageReader offsetStorageReader() {
...
 
+    private static void log(Object value, String property, String name, ILogger logger) {
+        if (logger.isInfoEnabled()) {
+            logger.info(name + " has `" + property + "` to " + value);
+        }
+    }
+
+    private static long extractTimestamp(SourceRecord record) {
+        if (record.valueSchema().field("ts_ms") == null) {
+            return 0L;
```
viliam-durina, line 402: We must use `NO_NATIVE_TIME` in this case, as the `flatMapEvent`'s javadoc specifies.

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSourceP.java
```diff
@@ -273,24 +371,40 @@ public OffsetStorageReader offsetStorageReader() {
...
+        } else {
+            return ((Struct) record.value()).getInt64("ts_ms");
```
viliam-durina, line 404: I wonder if it's the case the if the `ts_ms` field is present, it's always not null. It makes sense that it is, but we're dealing with a 3rd-party code and maybe could return the `NO_NATIVE_TIME` even if `getInt64` returns `null`.

viliam-durina (changes requested):

extensions/cdc-debezium/src/main/java/com/hazelcast/jet/cdc/impl/CdcSourceP.java
```diff
@@ -48,81 +54,159 @@
...
-                boolean added = addToBuffer(record, buf);
-                if (added) {
-                    Map<String, ?> partition = record.sourcePartition();
-                    Map<String, ?> offset = record.sourceOffset();
-                    state.setOffset(partition, offset);
-                }
+                Map<String, ?> partition = record.sourcePartition();
+                Map<String, ?> offset = record.sourceOffset();
+                state.setOffset(partition, offset);
+                task.commitRecord(record);
```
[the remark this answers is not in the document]
jbartok: Just saves the record's offset in a volatile variable:
image

jbartok:
run-cdc-debezium-tests

jbartok:
run-cdc-mysql-tests

jbartok:
run-cdc-postgres-tests

jbartok:
run-cdc-debezium-tests

jbartok:
run-cdc-mysql-tests

jbartok:
run-cdc-postgres-tests

viliam-durina (approved):

jbartok:
run-cdc-postgres-tests

jbartok:
run-cdc-mysql-tests

jbartok:
run-cdc-debezium-tests

jbartok:
run-cdc-postgres-tests

jbartok:
verify

PR #2887, opened 2021-02-12
Fix queue size display in Dot representation when default is changed

### LINKED ISSUE #2878: DAG printing does not reflect real queue size
Hello, I just stumbled into this little annoyance in Jet.

When you change the default queue size for the edges in the Jet configuration, the printed DAG of a pipeline does not reflect the new value, however the pipeline indeed does use the newly configured `queue_size`.

Consider this simple example for configuring an instance of Jet with a non default `queue_size`:

```java
// prepare Jet configuration
final JetConfig config = JetConfig.loadDefault();
config.getDefaultEdgeConfig().setQueueSize(4096);

// obtain a new Hazelcast Jet instance
final JetInstance jet = Jet.newJetInstance(config);
```

When starting a pipeline, the printed DAG still reports the default `queue_size` value. For example:
```
digraph DAG {
        "kafkaSource(some-topic)" [localParallelism=4];
        "jdbcSink" [localParallelism=1];
        "kafkaSource(some-topic)" -> "jdbcSink" [queueSize=1024];
}
```

Note that I confirmed that the configured `queue_size` is **indeed applied correctly**, just not printed correctly above.

The relevant code in Jet seems to be this:
https://github.com/hazelcast/hazelcast-jet/blob/84d30f20ed4167238de7f783720a38c731ca547f/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/DAG.java#L490-L492

Looks like that the correct thing to do is that, when `e.getConfig() == null`, the right thing to do is to pick the value from the `EdgeConfigDefault` and not directly from the default constant. If the default value is not changed in `EdgeConfigDefault`, it should contain anyway the value of the constant.

Or perhaps the bug is that `e.getConfig()` is not getting set at that point yet.

hhromic:
Fixes #2878

After studying the codebase and the implications, I decided to implement this fix without passing a `JetConfig` object instance but instead adding a new `defaultQueueSize` method argument. In this way, the `DAG` class remains uncoupled from any Jet instance and the caller is free to pass any value to be used for displaying a default queue size.

The `MasterJobContext.tryStartJob()` method is then adapted to pass the default queue size extracted from the `JetConfig` object of the running instance, similar to how the `defaultParallelism` value is passed.

I also took the opportunity to add some missing `@param` Javadoc annotations and made the `defaultLocalParallelism` argument name consistent across all methods in the `DAG` class.

Breaking changes:

The `DAG.toDotString(int defaultParallelism)` method signature is now `DAG.toDotString(int defaultLocalParallelism, int defaultQueueSize)`. Callers must now supply the queue size that will be shown if not overriden on the edge.

I believe this is a breaking change with very minimal impact as I don't think many users call this method directly.

Checklist:
- [x] Added a line in `hazelcast-jet-distribution/src/root/release_notes.txt` (for any non-trivial fix/enhancement/feature)

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/DAG.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/DAG.java
@@ -20,7 +20,6 @@ import com.hazelcast.function.SupplierEx;
 import com.hazelcast.internal.json.JsonArray;
 import com.hazelcast.internal.json.JsonObject;
 import com.hazelcast.internal.util.StringUtil;
-import com.hazelcast.jet.config.EdgeConfig;
 import com.hazelcast.jet.core.Edge.RoutingPolicy;
 import com.hazelcast.nio.ObjectDataInput;
 import com.hazelcast.nio.ObjectDataOutput;
@@ -41,6 +40,7 @@ import java.util.Set;
 import java.util.stream.IntStream;
 
 import static com.hazelcast.internal.util.Preconditions.checkTrue;
+import static com.hazelcast.jet.config.EdgeConfig.DEFAULT_QUEUE_SIZE;
 import static com.hazelcast.jet.core.Vertex.LOCAL_PARALLELISM_USE_DEFAULT;
 import static com.hazelcast.jet.impl.TopologicalSorter.topologicalSort;
 import static com.hazelcast.jet.core.Edge.DISTRIBUTE_TO_ALL;
@@ -441,25 +441,33 @@ public class DAG implements IdentifiedDataSerializable, Iterable<Vertex> {
      */
     @Nonnull
     public String toDotString() {
-        return toDotString(LOCAL_PARALLELISM_USE_DEFAULT);
+        return toDotString(LOCAL_PARALLELISM_USE_DEFAULT, DEFAULT_QUEUE_SIZE);
     }
 
     /**
-     * Returns a DOT format (graphviz) representation of the DAG and annotates
-     * the vertices using default parallelism with the supplied value.
+     * Returns a DOT format (graphviz) representation of the DAG, annotates
+     * the vertices using supplied default parallelism value, and the edges
+     * using supplied default queue size value.
+     *
+     * @param defaultLocalParallelism the local parallelism that will be shown if
+     *                                neither overridden on the vertex nor the
+     *                                preferred parallelism is defined by
+     *                                meta-supplier
+     * @param defaultQueueSize the queue size that will be shown if not overridden
+     *                         on the edge
      */
     @Nonnull
-    public String toDotString(int defaultParallelism) {
+    public String toDotString(int defaultLocalParallelism, int defaultQueueSize) {
         final StringBuilder builder = new StringBuilder(512);
         builder.append("digraph DAG {\n");
         int clusterCount = 0;
 
         for (Vertex v : this) {
-            int localParallelism = v.determineLocalParallelism(defaultParallelism);
+            int localParallelism = v.determineLocalParallelism(defaultLocalParallelism);
             String parallelism = localParallelism == LOCAL_PARALLELISM_USE_DEFAULT ?
-                defaultParallelism == LOCAL_PARALLELISM_USE_DEFAULT ?
+                defaultLocalParallelism == LOCAL_PARALLELISM_USE_DEFAULT ?
                     "default"
-                    : String.valueOf(defaultParallelism)
+                    : String.valueOf(defaultLocalParallelism)
                 : String.valueOf(localParallelism);
             builder.append("\t\"")
                    .append(escapeGraphviz(v.getName()))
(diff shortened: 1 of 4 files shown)


viliam-durina:
verify

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/DAG.java
```diff
@@ -441,25 +441,33 @@ public JsonObject toJson(int defaultLocalParallelism) {
...
-     * the vertices using default parallelism with the supplied value.
+     * Returns a DOT format (graphviz) representation of the DAG, annotates
+     * the vertices using supplied default parallelism value, and the edges
+     * using supplied default queue size value.
+     *
+     * @param defaultLocalParallelism the local parallelism that will be shown if
+     *                                neither overridden on the vertex nor the
+     *                                preferred parallelism is defined by
+     *                                meta-supplier
+     * @param defaultQueueSize the queue size that will be shown if not overriden
```
ufukyilmaz, line 456: Minor comment: "overriden` contains typo, should be overridden

hazelcast-jet-distribution/src/root/release_notes.txt
```diff
@@ -39,14 +39,16 @@ Thank you for your valuable contributions!
...
 3. Fixes
 
-[core] Jet now doesn't close System.out during JVM shutdown (#2649)
+[core] Fixed queue size display in `DAG.toDotString()` when default is changed (#2887).
 
 4. Breaking Changes
 
+[core] The `DAG.toDotString(int defaultParallelism)` method signature is now
+ `DAG.toDotString(int defaultLocalParallelism, int defaultQueueSize)`. Callers must
+ now supply the queue size that will be shown if not overriden on the edge.
```
ufukyilmaz, line 50: Same typo as other

ufukyilmaz (changes requested):
The implementation looks good to me. I only have a small request: Could you add a test that uses a non-default value for the queue size to [here]( https://github.com/hazelcast/hazelcast-jet/blob/master/hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/DotTest.java)?

hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/DAG.java
hhromic: Fixed! Thanks

hazelcast-jet-distribution/src/root/release_notes.txt
hhromic: I simplified now this note to:
```
[core] Added `defaultQueueSize` parameter to `DAG.toDotString(int)` method.
```
Similar to other existing release notes.

hhromic:
> The implementation looks good to me. I only have a small request: Could you add a test that uses a non-default value for the queue size to [here](https://github.com/hazelcast/hazelcast-jet/blob/master/hazelcast-jet-core/src/test/java/com/hazelcast/jet/core/DotTest.java)?

@ufukyilmaz yes, done now!  I also rebased the PR to latest `master`.
Let me know if you need something else before merging this PR. Thanks!

ufukyilmaz:
Hi @hhromic,
Thanks for your contribution and quick response.
After looking through the code a little more, I notice that there is another issue with this dot string:
We round up this configuration value to the next power of 2 while building the queues. But the values we show in this dot string are not these rounded values and I think we should show the real rounded up values in there. I think rounding up this value in this [setter](https://github.com/hazelcast/hazelcast-jet/blob/master/hazelcast-jet-core/src/main/java/com/hazelcast/jet/config/EdgeConfig.java#L70-L74) (earlier than we do now) solve this problem.

hhromic:
Ah interesting, I did not pay enough attenetion to the "round to power of 2" part of the Javadoc indeed.
I agree with you that the setter should be doing the rounding, specially if this is stated on its Javadoc block.

Even though I agree with you that we should be showing the real queue size, i.e. the rounded one, in the Dot string, I think maybe fixing where the rounding is done is subject of another PR? This PR was about making it display the queue size in the first place. I can work on that too but I think is better in a separate issue/PR ?

Let me know! Cheers!

ufukyilmaz:
I agree with you as it would be better to address this in a separate issue/PR.

ufukyilmaz (approved):

hhromic:
Hello again @ufukyilmaz , I went to check where the rounding is actually done and it seems to be in Hazelcast itself here:

https://github.com/hazelcast/hazelcast/blob/40d555dc1110da9f63fd66f504c83d4382d0384b/hazelcast/src/main/java/com/hazelcast/internal/util/concurrent/AbstractConcurrentArrayQueue.java#L82-L85

I don't think _moving_ the rounding out of there is a good idea, so are you suggesting to do the rounding _anyway_ in the setter in Jet to make sure is already rounded before reaching the SPSC queue?

ufukyilmaz:
Hi,
I'm not sure how to solve this. I cannot predict the bad consequences such use could lead to, so I don't want to comment on this. Let me open another issue for this.

ufukyilmaz:
verify

PR #2896, opened 2021-02-18
Add sink null coercion tests

gierlachg:
Addresses #2894


viliam-durina:
An alternative PR: #2897

PR #2898, opened 2021-02-19
Cleanup SQL java format/class resolution

gierlachg:
Removed java serialization specific code from generic `KvMetadataResolvers`.
Removed the necessity to copy (preprocess) options.

### DIFF
--- a/hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/kafka/PropertiesResolver.java
+++ b/hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/kafka/PropertiesResolver.java
@@ -26,6 +26,7 @@ import java.util.Properties;
 import java.util.Set;
 
 import static com.hazelcast.jet.sql.impl.connector.SqlConnector.AVRO_FORMAT;
+import static com.hazelcast.jet.sql.impl.connector.SqlConnector.JAVA_FORMAT;
 import static com.hazelcast.jet.sql.impl.connector.SqlConnector.JSON_FORMAT;
 import static com.hazelcast.jet.sql.impl.connector.SqlConnector.OPTION_KEY_CLASS;
 import static com.hazelcast.jet.sql.impl.connector.SqlConnector.OPTION_KEY_FORMAT;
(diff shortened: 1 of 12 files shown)


hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/keyvalue/JavaClassNameResolver.java
```diff
@@ -0,0 +1,62 @@
...
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.OffsetDateTime;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Stream;
+
+public final class JavaClassNameResolver {
+
+    private static final Map<String, String> CLASS_NAMES_BY_FORMAT = new HashMap<String, String>() {{
```
viliam-durina, line 30: We may rather use `ImmutableMap` rather than subclass, we depend on guava anyway due to calcite.

hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/kafka/PropertiesResolver.java
```diff
@@ -121,8 +116,15 @@ private static void withSerdeConsumerProperties(
...
-            String clazz = options.get(isKey ? SqlConnector.OPTION_KEY_CLASS : SqlConnector.OPTION_VALUE_CLASS);
+        } else if (AVRO_FORMAT.equals(format)) {
+            properties.putIfAbsent(deserializer, AVRO_DESERIALIZER);
+        } else if (JSON_FORMAT.equals(format)) {
+            properties.putIfAbsent(deserializer, BYTE_ARRAY_DESERIALIZER);
+        } else {
+            String clazz = JavaClassNameResolver.resolveClassName(
+                    format,
+                    options.get(isKey ? SqlConnector.OPTION_KEY_CLASS : SqlConnector.OPTION_VALUE_CLASS)
+            );
```
viliam-durina, line 127: I don't like this. So if there's `keyJavaClass` in the options, we ignore the format? If one does
```
OPTIONS (
    'keyFormat'='bigint',
    'keyJavaClass'='java.lang.Integer'
)
```
It will work exactly as `'keyFormat'='int'`. I think we should always check the format and if, based on that format, some other property is needed, use that. That is, if `keyFormat` is `java`, look ak `keyJavaClass`. If `keyFormat` is `int`, don't look at any other property (as we ignore unneeded properties).

Same below in `withSerdeProducerProperties`.

hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/keyvalue/KvMetadataJavaResolver.java
```diff
@@ -229,9 +232,10 @@ private KvMetadata resolveObjectMetadata(
     }
 
     private Class<?> loadClass(boolean isKey, Map<String, String> options) {
-        String classNameProperty = isKey ? OPTION_KEY_CLASS : OPTION_VALUE_CLASS;
-        String className = options.get(classNameProperty);
+        String format = options.get(isKey ? OPTION_KEY_FORMAT : OPTION_VALUE_FORMAT);
 
+        String classNameProperty = isKey ? OPTION_KEY_CLASS : OPTION_VALUE_CLASS;
+        String className = JavaClassNameResolver.resolveClassName(format, options.get(classNameProperty));
```
viliam-durina, line 238: Same here

viliam-durina (changes requested):

hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/keyvalue/KvMetadataJavaResolver.java
```diff
@@ -229,11 +232,17 @@ private KvMetadata resolveObjectMetadata(
...
-        if (className == null) {
-            throw QueryException.error("Unable to resolve table metadata. Missing '" + classNameProperty + "' option");
+        String className;
+        if (JAVA_FORMAT.equals(formatProperty)) {
+            className = options.get(classNameProperty);
+            if (className == null) {
+                throw QueryException.error("Unable to resolve table metadata. Missing '" + classNameProperty + "' option");
+            }
+        } else {
+            className = JavaClassNameResolver.resolveClassName(formatProperty);
```
viliam-durina, line 245: We can end up with null `className` here.

gierlachg: Technically, we can't but we can defensively leave the check for the posterity.

viliam-durina: Then we can `assert` it.

viliam-durina (approved):

PR #2922, opened 2021-02-24
Correct OR REPLACE description

### LINKED ISSUE #2920: Jet SQL does not allow replace-only
Jet 4.4 SQL allows

```
CREATE MAPPING mmm
```
and
```
CREATE OF REPLACE MAPPING mmm
```
but not
```
REPLACE MAPPING mmm
```

There is no way to change a mapping only if it exists.

Documentation suggests there is but it doesn't fail
image

gierlachg:
Fixes #2920

### DIFF
--- a/site/docs/sql/ddl.md
+++ b/site/docs/sql/ddl.md
@@ -36,8 +36,8 @@ TYPE type_identifier
 [ OPTIONS ( 'option_name' = 'option_value' [, ...] ) ]
 ```
 
-- `OR REPLACE`: if the mapping already exists, overwrite it; otherwise,
-  create one
+- `OR REPLACE`: if the mapping already exists, overwrite it instead
+  of failing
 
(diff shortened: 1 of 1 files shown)


site/docs/sql/ddl.md
```diff
@@ -36,8 +36,8 @@ TYPE type_identifier
 [ OPTIONS ( 'option_name' = 'option_value' [, ...] ) ]
 ```
 
-- `OR REPLACE`: if the mapping already exists, overwrite it; fail
-  otherwise
+- `OR REPLACE`: if the mapping already exists, overwrite it; otherwise,
+  create one
```
viliam-durina, line 40:
```suggestion
- `OR REPLACE`: if the mapping already exists, overwrite it instead
  of failing
```

viliam-durina (approved):

Issue #2925, opened 2021-02-24
Jet SQL, cannot determine external mapping name from information schema


neilstevenson:
On Jet 4.4, I can create a mapping

```
CREATE OR REPLACE MAPPING today EXTERNAL NAME day053
```
and replace it
```
CREATE OR REPLACE MAPPING today EXTERNAL NAME day054
```

The use case here would be to have an alias onto something different. So, `today` maps to `day053`, then `day054` if there was a map/whatever for each calendar day's data.

However, `SELECT * FROM information_schema.mappings` does not show the `external_mapping_name`, so I can't determine if the mapping is to today or yesterday

PR #2927, opened 2021-02-25
Add mapping external name to information_schema.mappings

gierlachg:
Addresses #2925


viliam-durina (changes requested):
We should add external name to columns too.

hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/connector/infoschema/MappingDefinition.java
```diff
@@ -14,52 +14,55 @@
...
 import com.hazelcast.sql.impl.schema.TableField;
 
 import java.util.List;
-import java.util.Map;
 import java.util.Map.Entry;
 import java.util.stream.Collectors;
 
 public class MappingDefinition {
 
     private final Table table;
```
viliam-durina, line 30: I think we can remove this field. `name()` will return `mapping.name()`, `schema()` will return `SCHEMA_NAME_PUBLIC` and we should not need the `tableFields()` after we remove the hidden columns from the `columns` view.

viliam-durina (changes requested):

viliam-durina (approved):

PR #2974, opened 2021-03-11
Prevent exception stack trace printing when user cancels the job

### LINKED ISSUE #2940: SQL 4.4 exception when closing query
On Jet 4.4, if I query an infinite source and run

```
sqlResult.close();
```

I get lots of these in the server logs

```
Caused by: com.hazelcast.sql.impl.QueryException: Query was cancelled by the user
	at com.hazelcast.sql.impl.QueryException.error(QueryException.java:55)
	at com.hazelcast.sql.impl.QueryException.cancelledByUser(QueryException.java:91)
	at com.hazelcast.jet.sql.impl.JetSqlResultImpl.close(JetSqlResultImpl.java:78)
	at com.hazelcast.sql.impl.AbstractSqlResult.close(AbstractSqlResult.java:41)
	at com.hazelcast.sql.impl.state.QueryClientStateRegistry.close(QueryClientStateRegistry.java:178)
	at com.hazelcast.sql.impl.client.SqlCloseMessageTask.call(SqlCloseMessageTask.java:40)
	at com.hazelcast.client.impl.protocol.task.AbstractCallableMessageTask.processMessage(AbstractCallableMessageTask.java:35)
	at com.hazelcast.client.impl.protocol.task.AbstractMessageTask.initializeAndProcessMessage(AbstractMessageTask.java:153)
	at com.hazelcast.client.impl.protocol.task.AbstractMessageTask.run(AbstractMessageTask.java:116)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
	at com.hazelcast.internal.util.executor.HazelcastManagedThread.executeRun(HazelcastManagedThread.java:76)
	at com.hazelcast.internal.util.executor.HazelcastManagedThread.run(HazelcastManagedThread.java:102)
```

It should be possible to cleanly end such a query. Cleanly end might simply mean suppressing these exceptions.

Fly-Style:
Many exceptions with a warning level were printed when the job was canceled by the user. 
Job cancellation by user is abnormal, but legal action, and user should see corresponding message, not exception stack traces.
This PR introduces such workaround for catching `QueryException` with `CANCELLED_BY_USER` error code.

Closes #2940.

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
@@ -42,6 +42,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CountDownLatch;
@@ -369,8 +370,14 @@ public class TaskletExecutionService {
                 }
                 progressTracker.mergeWith(result);
             } catch (Throwable e) {
-                logger.warning("Exception in " + t.tasklet, e);
-                t.executionTracker.exception(new JetException("Exception in " + t.tasklet + ": " + e, e));
+                if (e instanceof CancellationException) {
+                    logger.info("Job was cancelled by the user.");
+                    CancellationException ex = (CancellationException) e;
+                    t.executionTracker.exception(ex);
+                } else {
+                    logger.warning("Exception in " + t.tasklet, e);
+                    t.executionTracker.exception(new JetException("Exception in " + t.tasklet + ": " + e, e));
+                }
             } finally {
                 userMetricsContextContainer.setContext(null);
             }
--- a/hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/processors/RootResultConsumerSink.java
+++ b/hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/processors/RootResultConsumerSink.java
@@ -27,10 +27,14 @@ import com.hazelcast.jet.core.Watermark;
 import com.hazelcast.jet.sql.impl.JetQueryResultProducer;
 import com.hazelcast.jet.sql.impl.JetSqlCoreBackendImpl;
 import com.hazelcast.sql.impl.JetSqlCoreBackend;
+import com.hazelcast.sql.impl.QueryException;
 
 import javax.annotation.Nonnull;
 
+import java.util.concurrent.CancellationException;
+
 import static com.hazelcast.jet.core.ProcessorMetaSupplier.forceTotalParallelismOne;
+import static com.hazelcast.sql.impl.SqlErrorCode.CANCELLED_BY_USER;
 
 public final class RootResultConsumerSink implements Processor {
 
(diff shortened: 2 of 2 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
```diff
@@ -114,7 +118,7 @@ public TaskletExecutionService(NodeEngineImpl nodeEngine, int threadCount, Hazel
         // register metrics
         MetricsRegistry registry = nodeEngine.getMetricsRegistry();
         MetricDescriptor descriptor = registry.newMetricDescriptor()
-                                                     .withTag(MetricTags.MODULE, "jet");
```
mtopolnik, line 117: This seems to be an indentation-only diff. I'd remove it.

hazelcast-jet-sql/src/test/java/com/hazelcast/jet/sql/SqlAggregateTest.java
```diff
@@ -979,6 +980,17 @@ public void test_groupingSet() {
                 .hasMessageContaining("Function 'GROUPING SETS' does not exist");
     }
 
+    // Test detected behaviour described in https://github.com/hazelcast/hazelcast-jet/issues/2940 .
+    // Fixed in #2974.
+    @Ignore
```
mtopolnik, line 985: Is the point of this test to be run manually? Since it's ignored and doesn't do any assertions.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
```diff
@@ -369,8 +373,15 @@ private void runTasklet(TaskletTracker t) {
                 }
                 progressTracker.mergeWith(result);
             } catch (Throwable e) {
-                logger.warning("Exception in " + t.tasklet, e);
-                t.executionTracker.exception(new JetException("Exception in " + t.tasklet + ": " + e, e));
+                // No need to print collected exceptions when user cancels the job.
+                // We can use instanceof here, because it is 'cold' code.
+                if (e instanceof QueryException && ((QueryException) e).getCode() == CANCELLED_BY_USER) {
+                    logger.warning("Job was cancelled by the user.");
```
mtopolnik, line 379: I didn't test it, but will it print this warning once for each tasklet in the job?

hazelcast-jet-sql/src/test/java/com/hazelcast/jet/sql/SqlAggregateTest.java
Fly-Style: Yes, because it reproduces log messages in terminal.

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/execution/TaskletExecutionService.java
Fly-Style: Correct

hazelcast-jet-sql/src/test/java/com/hazelcast/jet/sql/SqlAggregateTest.java
viliam-durina: I'd remove this test. It will never fail, nobody will ever run it. Also it doesn't belong to this class. For example, there's `JetSqlResultImplTest.when_closed_then_iteratorFails` which is very similar to this one.

viliam-durina (changes requested):
The code converts `QueryException` to `JobTerminateRequestedException`. But this exception is special, it's used internally to handle a requested cancellation through Jet API (using `Job.cancel()`). In the test the query is cancelled through SQL API. I didn't debug exactly why but the job isn't cancelled and gets stuck. Just the logging should be different in the branches, the rest should be same.

Second, I tried the test on `master` and saw the exception printed only once on `WARN` level. This is not what neil reported, he reported many instances. I'm not bothered by one instance, I'd rather do nothing. The error handling code should be as simple as possible - it's used when you're in trouble and you should not run into exceptions when handling exceptions. We should try to reproduce the case when the exception is logged many times for one cancelled query, if there's such a case at all. If not, I'd do nothing.

viliam-durina (approved):

PR #2976, opened 2021-03-11
Add LIMIT clause support

alex-dukhno:
add `LIMIT` clause to Jet SQL

Checklist:
- [ ] Labels and Milestone set


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/exception/EventLimitExceededException.java
```diff
@@ -0,0 +1,16 @@
+package com.hazelcast.jet.impl.exception;
+
+import com.hazelcast.jet.JetException;
+import com.hazelcast.jet.impl.TerminationMode;
+
+import javax.annotation.Nonnull;
+
+public class EventLimitExceededException extends JetException {
+    public EventLimitExceededException() {
+        this(TerminationMode.CANCEL_GRACEFUL);
```
viliam-durina, line 10: `CANCEL_GRACEFUL` is cancellation with a snapshot, an enterprise feature. 
And also what's the reason we're using this enum here? Just to get the text?

hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/opt/physical/CreateDagVisitor.java
```diff
@@ -225,6 +227,23 @@ public Vertex onRoot(JetRootRel rootRel) {
...
+        Object val = fetch.eval(EmptyRow.INSTANCE, ExpressionUtil.NOT_IMPLEMENTED_ARGUMENTS_CONTEXT);
+        assert val instanceof Number;
+
+        long fetchValue = ((Number) val).longValue();
+        assert fetchValue >= 0;
+        AtomicLong limit = new AtomicLong(fetchValue);
+
+        Vertex vertex = dag.newUniqueVertex("Limit", mapUsingServiceP(
+                ServiceFactories.nonSharedService(ctx -> ExpressionUtil.limitFn(limit)),
+                (Function<Object[], Object[]> limitFn, Object[] row) -> limitFn.apply(row)));
```
viliam-durina, line 241: This works only on one-member clusters.

hazelcast-jet-sql/src/test/java/com/hazelcast/jet/sql/SqlLimitTest.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Test
+    public void limit_map() throws Exception {
+        IMap<Object, Object> map = instance().getMap("map");
+        map.put(1, 1);
+        map.put(2, 2);
+        map.put(3, 3);
+
+        SqlResult execute = instance().getSql().execute("select * from map limit 1");
```
viliam-durina, line 73: The test works because it executes on IMDG engine :)

alex-dukhno: I'll remove it before merge. I used it to debug the difference between IMDG and Jet query plan

hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/exception/EventLimitExceededException.java
gierlachg: I also don't like the idea of controlling the flow with an exception.

hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/JetQueryResultProducer.java
```diff
@@ -141,6 +155,9 @@ private boolean isDone() {
                     // finish the rows first
                     return rows.isEmpty();
                 }
+                if (exception instanceof EventLimitExceededException) {
+                    return true;
```
viliam-durina, line 159: Looks like we also should return `rows.isEmpty()`, same as 3 lines above.

hazelcast-jet-sql/src/test/java/com/hazelcast/jet/sql/SqlTestSupport.java
```diff
@@ -120,6 +120,17 @@ public static void assertRowsAnyOrder(String sql, Collection<Row> expectedRows)
         assertThat(actualRows).containsExactlyInAnyOrderElementsOf(expectedRows);
     }
 
+    public static void assertContainsOnlyOneOfRows(String sql, Row... expectedRows) {
+        assertContainsOnlyNOfRows(sql, 1, expectedRows);
+    }
+
+    public static void assertContainsOnlyNOfRows(String sql, int n, Row... expectedRows) {
```
viliam-durina, line 127:
```suggestion
    /**
     * Asserts that the result of {@code sql} contains a subset of {@code expectedRows}, but
     * only a subset of them with size of {@code subsetSize}.
     */
    public static void assertContainsSubsetOfRows(String sql, int subsetSize, Row... expectedRows) {
```
I'd also consider moving this method to `SqlLimitTest`, not sure how useful it's elsewhere.

hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/opt/physical/CreateDagVisitor.java
```diff
@@ -215,13 +217,29 @@ public Vertex onNestedLoopJoin(JoinNestedLoopPhysicalRel rel) {
     }
 
     public Vertex onRoot(JetRootRel rootRel) {
-        Vertex vertex = dag.newUniqueVertex("ClientSink",
-                rootResultConsumerSink(rootRel.getInitiatorAddress()));
+        Vertex vertex;
+        RelNode input = rootRel.getInput();
+        if (input instanceof SortPhysicalRel) {
+            SortPhysicalRel limitRel = (SortPhysicalRel) input;
```
viliam-durina, line 223: `limitRel` -> `sortRel`

hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/opt/physical/CreateDagVisitor.java
```diff
@@ -215,13 +217,29 @@ public Vertex onNestedLoopJoin(JoinNestedLoopPhysicalRel rel) {
...
+            Expression<?> fetch = limitRel.fetch();
```
viliam-durina, line 224: We use only the `fetch` field of the `limitRel`. I think we should at least assert that offset is `0` and that there's no collation. It currently isn't possible because the unsupported-feature-verifier checks there's no ORDER BY, but who know where else Calcite uses the `Sort` node, it's better to assert it.

hazelcast-jet-sql/src/test/java/com/hazelcast/jet/sql/SqlLimitTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+import com.hazelcast.jet.sql.impl.connector.test.TestBatchSqlConnector;
+import com.hazelcast.sql.SqlService;
+import com.hazelcast.sql.impl.type.QueryDataType;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import static java.util.Arrays.asList;
+
+public class SqlLimitTest extends SqlTestSupport {
```
viliam-durina, line 27: We can also add a test for a streaming source, it should work.

viliam-durina (changes requested):

hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/JetQueryResultProducer.java
```diff
@@ -154,4 +168,12 @@ private boolean isDone() {
             super("Done normally", null, false, false);
         }
     }
+
+    private static class ResultLimitReachedException extends Exception {
+        ResultLimitReachedException() {
+            // Use writableStackTrace = false, the exception is not created at a place where it's thrown,
+            // it's better if it has no stack trace then.
+            super("Done by reaching events number in SQL LIMIT clause", null, false, false);
```
viliam-durina, line 176:
```suggestion
            super("Done by reaching the item number in SQL LIMIT clause", null, false, false);
```

hazelcast-jet-sql/src/test/java/com/hazelcast/jet/sql/SqlLimitTest.java
```diff
@@ -0,0 +1,158 @@
...
+    @Test
+    public void floatNumber_asLimitValue() {
+        String tableName = createTable(
+                new String[]{"Alice", "1"},
+                new String[]{"Bob", "2"},
+                new String[]{"Joey", "3"}
+        );
+
+        assertRowsAnyOrder(
+                "SELECT name FROM " + tableName + " LIMIT 5.2",
```
viliam-durina, line 91: Weird, we get parsing error for `-10`, but `5.2` works fine? In that case I'd like to see a test where the number actually limits, I guess `LIMIT 2.99` will return 2 rows out of 3.

hazelcast-jet-sql/src/test/java/com/hazelcast/jet/sql/SqlTestSupport.java
```diff
@@ -120,6 +120,21 @@ public static void assertRowsAnyOrder(String sql, Collection<Row> expectedRows)
         assertThat(actualRows).containsExactlyInAnyOrderElementsOf(expectedRows);
     }
 
+    public static void assertContainsOnlyOneOfRows(String sql, Row... expectedRows) {
```
viliam-durina, line 123: This variant is too niche for the utility, I'd move it to the test class. Also the `assertContainsSubsetOfRows`. I don't think it will be used in any other test than for the LIMIT clause.

hazelcast-jet-sql/src/test/java/com/hazelcast/jet/sql/SqlLimitTest.java
alex-dukhno: holy crap, this is Postgres example:
```
postgres=# select * from numbers ;
 col_si | col_i | col_bi
--------+-------+--------
    100 |   100 |    100
    100 |   100 |    100
      1 |     2 |      3
      1 |     2 |      3
      1 |     2 |      3
      1 |     2 |      3
      1 |     2 |      3
(7 rows)

postgres=# select * from numbers limit 2.99;
 col_si | col_i | col_bi
--------+-------+--------
    100 |   100 |    100
    100 |   100 |    100
      1 |     2 |      3
(3 rows)

postgres=# select * from numbers limit 2.5;
 col_si | col_i | col_bi
--------+-------+--------
    100 |   100 |    100
    100 |   100 |    100
      1 |     2 |      3
(3 rows)

postgres=# select * from numbers limit 2.45;
 col_si | col_i | col_bi
--------+-------+--------
    100 |   100 |    100
    100 |   100 |    100
(2 rows)
```

PR #2991, opened 2021-03-17
Update IMap Portable SQL documentation

gierlachg:


site/docs/sql/imap-connector.md
```diff
@@ -63,8 +63,13 @@ For this format, you need to specify additional options:
...
 
-Jet resolves the column list by looking at the `ClassDefinition` found
-using the given factory ID, class ID, and version.
+If you omit a column list from the `CREATE MAPPING` command, Jet will
+resolve column names and types by looking at the `ClassDefinition`
+found using the given factory ID, class ID, and version.
+
+Given the triplet of factory ID, class ID and version does not
+correspond to an existing `ClassDefinition`, the column list is
+mandatory.
```
viliam-durina, line 72: We should explicitly say that we'll create the `ClassDefinition`:
```suggestion
If the `ClassDefinition` with the given IDs is not known to the cluster,
the column list is mandatory and Jet will create the `ClassDefinition`
based on the column list.
```

viliam-durina (approved):

PR #2992, opened 2021-03-17
Bump Parquet version to 1.11.1

gierlachg:


frant-hartm:
We should run the PR builder once the avro update is merged so it runs with the updated version.

gierlachg:
It's just an example version. The _core_ already uses 1.11.1.

examples/hadoop/pom.xml
```diff
@@ -21,7 +21,7 @@
 
     <artifactId>hazelcast-jet-examples-hadoop</artifactId>
     <properties>
-        <parquet.version>1.10.1</parquet.version>
+        <parquet.version>1.11.1</parquet.version>
```
frant-hartm, line 24: We are already at this version, and it is defined in the root pom, so this can be safely deleted?

frant-hartm (changes requested):

frant-hartm (approved):

Issue #2834, opened 2021-01-20
Avoid of printStackTrace


olukas:
Jet code includes some places where `printStackTrace()` is used. We should rather log it through Logger and in some case we should also work handle exception.

Some examples:
- https://github.com/hazelcast/hazelcast-jet/blob/afbeccd/hazelcast-jet-core/src/main/java/com/hazelcast/jet/function/Observer.java#L101
- https://github.com/hazelcast/hazelcast-jet/blob/afbeccd/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/JetBootstrap.java#L147
- https://github.com/hazelcast/hazelcast-jet/blob/afbeccd/hazelcast-jet-core/src/main/java/com/hazelcast/jet/pipeline/test/ParallelStreamP.java#L99

mtopolnik:
The occurrence in `JetBootstrap` should stay, I think, because after a failed shutdown you can't count on a working Hazelcast logger instance.

gierlachg:
`ParallelStreamP` was addressed in #2994.
Not sure if we can do anything with `Observer` without breaking changes - the other thing is, one can always override the default implementation.

PR #2994, opened 2021-03-18
Remove printStackTrace() from ParallelStreamP

gierlachg:
Addressed `ParallelStreamP` part of #2834


ufukyilmaz (approved):
Thanks for fixing this

PR #2997, opened 2021-03-21
Fix nullability annotation in EventTimeMapper

jerrinot:
1. It's called from https://github.com/hazelcast/hazelcast-jet/blob/2797c7ad3cecffa673e2d6f7ae276777885ad155/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/processor/InsertWatermarksP.java#L63 which accepts Null items

2. It delegates to https://github.com/hazelcast/hazelcast-jet/blob/2797c7ad3cecffa673e2d6f7ae276777885ad155/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/EventTimeMapper.java#L211 where the item is already annotated with @Nullable

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/EventTimeMapper.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/EventTimeMapper.java
@@ -182,14 +182,15 @@ public class EventTimeMapper<T> {
      *                 event, event.getPartition(), nativeEventTime));
      * }</pre>
      *
-     * @param event           the event
+     * @param event           the event to flat-map.
+     *                        If {@code null}, it's equivalent to the behavior of {@link #flatMapIdle()}
      * @param partitionIndex  the source partition index the event came from
      * @param nativeEventTime native event time in case no {@code timestampFn} was supplied or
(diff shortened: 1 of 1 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/core/EventTimeMapper.java
```diff
@@ -189,7 +189,7 @@ public EventTimeMapper(EventTimePolicy<? super T> eventTimePolicy) {
      * @return a traverser over the given event and the watermark (if it was due)
      */
     @Nonnull
-    public Traverser<Object> flatMapEvent(@Nonnull T event, int partitionIndex, long nativeEventTime) {
+    public Traverser<Object> flatMapEvent(@Nullable T event, int partitionIndex, long nativeEventTime) {
```
viliam-durina, line 192: It also needs javadoc. E.g.: 
> `@param event the event. If {@code null}, it's equivalent to the behavior of {@link #flatMapIdle()}.

jerrinot: it would create a circular definition. as the other `flatMap()` is pointing out back to this method.

viliam-durina (approved):

Issue #2853, opened 2021-01-27
com.hazelcast.jet.s3.S3SinkTest.when_writesToExistingFile_then_overwritesFile


olukas:
_master_  (commit 302eb2d229f14b9a6ea306922e8a2b05836b06d1)

Failed on IBM JDK 8: http://jenkins.hazelcast.com/job/jet-oss-master-nightly-ibm-jdk8/30/testReport/junit/com.hazelcast.jet.s3/S3SinkTest/when_writesToExistingFile_then_overwritesFile/

Stacktrace:
```
java.lang.AssertionError: expected:<200> but was:<0>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:633)
	at com.hazelcast.jet.s3.S3TestBase.lambda$testSink$3(S3TestBase.java:97)
	at com.hazelcast.jet.s3.S3TestBase$$Lambda$1994/0000000051736700.run(Unknown Source)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1248)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1350)
	at com.hazelcast.jet.s3.S3TestBase.testSink(S3TestBase.java:89)
	at com.hazelcast.jet.s3.S3TestBase.testSink(S3TestBase.java:72)
	at com.hazelcast.jet.s3.S3SinkTest.when_writesToExistingFile_then_overwritesFile(S3SinkTest.java:89)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:90)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:55)
	at java.lang.reflect.Method.invoke(Method.java:508)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
[...]

PR #3003, opened 2021-03-23
Create directory for each test run in the bucket for isolation

gurbuzali:
Fixes https://github.com/hazelcast/hazelcast-jet/issues/2853

Checklist:


gierlachg (approved):
The downside of random prefix is, if for some reason `deleteObjects()` is not executed we might end up with some leftovers (which AWS will be charging us for).

gurbuzali:
we store very little amount of data for the tests and it is really cheap ($0.023 per GB, per month)

PR #3006, opened 2021-03-24
Enable the usage of ExpressionEvalContext while evaluating VALUES

gierlachg:
Refactored `VALUES`, so the evaluation is performed in a `Processor` and uses `ExpressionEvalContext` - it's an enabler for future work on dynamic parameters.


gierlachg:
Will create PR against platform.

PR #3008, opened 2021-03-25
Enable the usage of ExpressionEvalContext while evaluating table function arguments

gierlachg:
Refactored table functions, so the evaluation of arguments is performed during `DAG` creation -  it's an enabler for future work on dynamic parameters.


hazelcast-jet-sql/src/main/java/com/hazelcast/jet/sql/impl/ExpressionUtil.java
```diff
@@ -39,6 +39,7 @@
      * Useful when evaluating expressions in planning phase where these are not
      * available.
      */
+    @Deprecated
```
viliam-durina, line 42: We should comment why: because dynamic params don't work then.

viliam-durina (changes requested):
There's no test for the new feature

gierlachg:
Will create PR against platform.

PR #3009, opened 2021-03-25
Limit the number of parallel partition reads for map/cache.

### LINKED ISSUE #2926: Job that reads from Hazelcast IMDG IMap source can cause JVM OOM when member heap is small
Given the following pipeline code:

```
pipeline
    .readFrom(Sources.remoteMap(mapName, clientConfig))
    .map(entry -> entry.getKey())
    .writeTo(Sinks.logger());
```

where the remote map is configured for NATIVE format, and the JVM heap is small (i.e. 1 GB), the IMDG members can experience OOM errors when preparing to send the data for the Jet client.

Attached is a reproducer project that demonstrates this behavior (thanks @neilstevenson). You may need to increase the `valueSize` to `4000` to trigger the OOM when running in 4.x and later. Also, attaching a heap dump that was generated from running this project with `-XX:+HeapDumpOnOutOfMemoryError` specified on the IMDG members.

reproducer: [zen7099.zip](https://github.com/hazelcast/hazelcast-jet/files/6037765/zen7099.zip)
heap dump: [hz_4_dump.tar.gz](https://github.com/hazelcast/hazelcast-jet/files/6037775/hz_4_dump.tar.gz)

In the dump you can see that `NioOutboundPipeline.writeQueue` is full.

In a related Slack thread, @viliam-durina made two suggestions for remediating this issue that I wanted to capture here:

1. Currently we limit data transfer based on item count rather than bytes which can be a problem in cases like this with large values. The suggestion was to limit based on bytes, but this would require changes on both Jet and IMDG sides.
2. Reduce the number of parallel requests sent, and reduce the number of items fetched, which would only require changes on the Jet side.

gurbuzali:
Reading from all of the partitions in parallel can cause OOM on the
server-side. We limit the parallel reads to a constant value (5) and
decrease the max fetch size (2048).

Fixes #2926

Checklist:

### DIFF
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadMapOrCacheP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadMapOrCacheP.java
@@ -99,8 +99,13 @@ import static java.util.stream.Collectors.toList;
  */
 public final class ReadMapOrCacheP<F extends CompletableFuture, B, R> extends AbstractProcessor {
 
-    private static final int MAX_FETCH_SIZE = 2048;
+    /**
+     * See <a href="https://github.com/hazelcast/hazelcast-jet/pull/3009#discussion_r606338266">discussion</a>
+     * for the numbers and how {@link #MAX_FETCH_SIZE} and
+     * {@code #MAX_PARALLEL_READ} affect the throughput.
+     */
     private static final int MAX_PARALLEL_READ = 5;
+    private static final int MAX_FETCH_SIZE = 2048;
 
     private final Reader<F, B, R> reader;
     private final int[] partitionIds;
--- a/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadMapOrCacheP.java
+++ b/hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadMapOrCacheP.java
@@ -99,11 +99,13 @@ import static java.util.stream.Collectors.toList;
  */
 public final class ReadMapOrCacheP<F extends CompletableFuture, B, R> extends AbstractProcessor {
 
-    private static final int MAX_FETCH_SIZE = 16384;
+    private static final int MAX_FETCH_SIZE = 2048;
+    private static final int MAX_PARALLEL_READ = 5;
 
     private final Reader<F, B, R> reader;
     private final int[] partitionIds;
     private final IterationPointer[][] readPointers;
+    private final int maxParallelRead;
 
     private F[] readFutures;
 
@@ -112,6 +114,8 @@ public final class ReadMapOrCacheP<F extends CompletableFuture, B, R> extends Ab
     private int currentBatchPosition;
     private int currentPartitionIndex = -1;
     private int numCompletedPartitions;
+    private int nextPartitionReadIndex;
+    private int partitionReadCount;
 
     private Object pendingItem;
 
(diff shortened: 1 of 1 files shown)


hazelcast-jet-core/src/main/java/com/hazelcast/jet/impl/connector/ReadMapOrCacheP.java
```diff
@@ -99,11 +99,13 @@
  */
 public final class ReadMapOrCacheP<F extends CompletableFuture, B, R> extends AbstractProcessor {
 
-    private static final int MAX_FETCH_SIZE = 16384;
+    private static final int MAX_FETCH_SIZE = 2048;
+    private static final int MAX_PARALLEL_READ = 5;
```
viliam-durina, line 103: Did you do any tests to measure these values don't deteriorate the throughput? I'm not sure about them, maybe 5 is too low, maybe not. Local parallelism is 1 by default, one can increase the number of parallel reads by increasing the local parallelism. But we should do some basic benchmarks and put the results here to a comment to this value.

gurbuzali: I've run some benchmarks on my local, 6 cores 32GB :

Environment:
2 Nodes (separate JVM), 5G each, map populated with 10M items (integer)

Pipeline:
```
        p.readFrom(Sources.map("map"))
                .writeTo(Sinks.noop());
```

First I run 20 jobs for warmup
For each value of the parameter (MAX_PARALLEL_READ) I run 100 jobs and get the average

Results:

Before changes ->  average: 860

Since the partition count is 271, and we have 2 nodes, each processor will have around 135 partitions
setting MAX_PARALLEL_READ above 135 is effectively same so starting the testing from 160

MAX_FETCH_SIZE=16384, MAX_PARALLEL_READ=160-> average: 857
MAX_FETCH_SIZE=16384, MAX_PARALLEL_READ=80  -> average: 859
MAX_FETCH_SIZE=16384, MAX_PARALLEL_READ=40  -> average: 800
MAX_FETCH_SIZE=16384, MAX_PARALLEL_READ=20  -> average: 790
MAX_FETCH_SIZE=16384, MAX_PARALLEL_READ=10   -> average: 771
MAX_FETCH_SIZE=16384, MAX_PARALLEL_READ=5    -> average: 779
MAX_FETCH_SIZE=16384, MAX_PARALLEL_READ=3    -> average: 761
MAX_FETCH_SIZE=16384, MAX_PARALLEL_READ=1     -> average: 1106

MAX_FETCH_SIZE=2048, MAX_PARALLEL_READ=160-> average: 810
MAX_FETCH_SIZE=2048, MAX_PARALLEL_READ=80  -> average: 776
MAX_FETCH_SIZE=2048, MAX_PARALLEL_READ=40  -> average: 786
MAX_FETCH_SIZE=2048, MAX_PARALLEL_READ=20  -> average: 775
MAX_FETCH_SIZE=2048, MAX_PARALLEL_READ=10   -> average: 757
MAX_FETCH_SIZE=2048, MAX_PARALLEL_READ=5    -> average: 720
MAX_FETCH_SIZE=2048, MAX_PARALLEL_READ=3    -> average: 688
MAX_FETCH_SIZE=2048, MAX_PARALLEL_READ=1     -> average: 1191

Although `MAX_PARALLEL_READ=3` has better results, I believe it is related to the core count in the computer.
And a single node run showed very close results for `3 -> 1059`  `5 -> 1061`.


Looks like by introducing this limit to the parallel read, we are actually improving the throughput.

viliam-durina: Add a link to this comment to the constants.

viliam-durina (approved):

PR #3074, opened 2021-12-06
Decommission download machine - update links

olukas:
Update links to do not link current download machine. See https://hazelcast.atlassian.net/browse/HZ-741 for more details.


frant-hartm (approved):
LGTM.

I tried a few random links, works both in the browser and via wget.

